Issue #81, opened 2026-05-02
📋 Plan: Add Actions section (issue #80)


eonist:
## Plan for issue [#80 — Add action and job section](https://github.com/eonist/runner-bar/issues/80)

This issue tracks the implementation plan for the new **Actions** section.

---

## What We're Building

A new **"Actions"** section above the existing **"Active Jobs"** section in `PopoverMainView`. Actions show GitHub workflow runs grouped by repo with a full drill-down: **Actions → Jobs → Steps → Log**.

The existing Active Jobs section and all its plumbing remain **unchanged**.

---

## Navigation Hierarchy

```
Actions list  →  ActionDetailView (jobs)  →  JobDetailView (steps)  →  StepLogView (log)
  [new]              [new]                      [existing]               [existing]
```

---

## Files Changed / Created

| File | Change |
|---|---|
| `ActionRun.swift` | **New** — `ActionRun` model + `fetchActions()` fetch function |
| `ActionDetailView.swift` | **New** — jobs-in-run drill-down view (mirrors `JobDetailView`) |
| `ActiveJob.swift` | `private` → `internal` on `ghAPI`, `JobsResponse`, `JobPayload`, `StepPayload` |
| `RunnerStore.swift` | Add `actions` property + `actionCache` + `prevLiveRuns` + fetch block |
| `PopoverMainView.swift` | Add `@Published var actions` to observable; add Actions section + version bump |
| `AppDelegate.swift` | Wire `onSelectAction` callback + `actionDetailView(run:)` factory |

---

## Data Model — `ActionRun`

```swift
struct ActionRun: Identifiable {
    let id: Int              // workflow run_id
    let name: String         // workflow name (e.g. "SonarQube", "vitest")
    let repo: String         // owner/repo
    let status: String       // queued | in_progress | completed
    let conclusion: String?  // success | failure | cancelled | skipped
    let headBranch: String?
    let createdAt: Date?
    let updatedAt: Date?
    let htmlUrl: String?
    var jobs: [ActiveJob] = []   // reuses existing model
    var isDimmed: Bool = false   // true for persisted completed runs
}
```

---

## Fetch Strategy

- Fetch `in_progress` + `queued` runs via `/repos/{scope}/actions/runs?status=...`
- For each run, fetch jobs via `/repos/{scope}/actions/runs/{run_id}/jobs` — reuses existing `JobsResponse`/`JobPayload` codable structs
- Skip org scopes (same limitation as `fetchActiveJobs`)

[...]

Issue #31, opened 2026-04-24
⭐ Active jobs section behaviour


eonist:
spec:

List behavior of active jobs;

1. max show 3 items. 
2. always prioritize inprogress and queued in the top. 
3. after items go to done. they persist. when an item moves from inprogress to done. it should stick around!!! it must persist. 3 done items must stay behind
4. if you ran 4 items. and they are now done. 3 items will show as done and persist. 
5. queued -> inprogress. -> done.  
6. always persist the last item that moved from inprogress to done. move other persisted done items out. 
7. in-progress take priority above queued. if there are 3 in progress show these.
8.  if there are 2 in progress and 1 queued and one done. show inprogress and queued
9. if there are 1 queued and 12done. and 1 inprogress. show1  inprogress 1 queued 1 done

solve this holistically. dont just wing it. solve it properly. no quick hacks etc. make a proper system that reflects this.

eonist:
Ref only

PR #82, opened 2026-05-02
feat: Add Actions section with drill-down navigation (issue #80)

### LINKED ISSUE #80: ⭐ Add action and job section
The task is to add a new section. "Actions" section. its complimentary to the "Active jobs" section which should be unchanged more or less. 

this py script can be looked at for inspiration regarding how to show actions. its called "Groups" in this py script: https://github.com/psw-pwa/psw-pwa/blob/main/scripts/ci-dash.py

- We should have 2 sections. One named "Actions". and one named "Active jobs"
- Active jobs are whichever jobs that are active. they can belong to any action. 
- Actions are collections of jobs and can be something many steps like: SonarQube, vitest, ui-tests, sec-audit. these belong to a repo. 
- This is how active jobs work: https://github.com/eonist/runner-bar/issues/31
- When we go into an action we see the list of steps as they appear on GitHub. these run  from top to bottom
- We can navigate from action in actions section. where we then see the jobs in that action. we can then navigate into each job which we then see the steps. and then we can press a step to see its log. the actions -> jobs -> steps -> log flow is similar to the flow in the active jobs section
- we must also persist previous actions after they have completed. see the ci-dash.py for inspo regarding how to do this. also the current active jobs section persist somewhat. although not perfectly yet.

eonist:
## Summary

Adds a new **Actions** section to the menu bar popover, sitting above the existing **Active Jobs** section. Workflow runs are shown grouped per repo with full drill-down navigation: **Actions → Jobs → Steps → Log**.

Closes #80

---

### New files
| File | Purpose |
|---|---|
| `ActionRun.swift` | `ActionRun` model + `fetchActions()` — fetches in_progress/queued workflow runs and enriches each with its job list, reusing existing `ghAPI`/`JobsResponse`/`JobPayload` from `ActiveJob.swift` |
| `ActionDetailView.swift` | Navigation level 2a — jobs inside a selected workflow run; mirrors `JobDetailView` layout (header outside ScrollView, list inside) |

### Modified files
| File | Change |
|---|---|
| `ActiveJob.swift` | `private` → `internal` on `ghAPI`, `JobsResponse`, `JobPayload`, `StepPayload` (one-line each) so `ActionRun.swift` can reuse them |
| `RunnerStore.swift` | Add `actions: [ActionRun]` + `prevLiveRuns`/`actionCache` persistence layer (freeze-on-vanish pattern mirrors `completedCache`); snapshots taken on main thread before background block; cache trimmed to 5 (ci-dash.py `MAX_GROUPS`) |
| `PopoverMainView.swift` | Add Actions section + action row UI + `onSelectAction` callback + `@Published var actions` on observable; version bump v0.29 → v0.30 |
| `AppDelegate.swift` | Wire `onSelectAction`; add `actionDetailView`, `detailViewFromAction`, `logViewFromAction` factories for correct per-level back-navigation |

---

## Navigation chain

```
PopoverMainView  →  ActionDetailView (jobs)  →  JobDetailView (steps)  →  StepLogView (log)
                     [new]                      [existing, unchanged]     [existing, unchanged]
```

Back navigation works at each level:
- Log → Steps (`detailViewFromAction`)
- Steps → Jobs (`actionDetailView`)
- Jobs → Actions list (`mainView`)

---

## Persistence

Mirrors the existing `completedCache`/`prevLiveJobs` pattern exactly:
- Runs that vanish from the API between polls are frozen into `actionCache` with `isDimmed = true`
- Freshly-concluded runs (conclusion != nil) are also frozen with real data
- Cache trimmed to 5 most-recently-updated entries (matches `ci-dash.py MAX_GROUPS = 5`)
- Display: `in_progress` → `queued` → cached done (newest first), capped at 5

---

## Regression guards

- **ActiveJob 3-callsite warning** — no new `ActiveJob(...)` callsites added in new code. `ActionRun` wraps `[ActiveJob]`, the job fetch reuses `fetchJobsForRun` which produces `ActiveJob` values through the existing codable path.
- **PopoverMainView RULE 1–5** — `idealWidth` untouched; all rows use `.padding(.horizontal, 12)`; load-bearing `Spacer()` inside action row HStack; no `.fixedSize()`; no `objectWillChange.send()` in `reload()`.
- **RunnerStore thread safety** — action state snapshots are taken on the main thread before the background block (same pattern as `snapPrev`/`snapCache` for jobs).
- **AppDelegate navigate() rule** — rootView swap only, zero size changes, no `contentSize` touched.

---

## Ref
- Planning issue: #81
- Related: #31 (Active Jobs — the pattern this mirrors)

### DIFF
--- a/Sources/RunnerBar/ActionDetailView.swift
+++ b/Sources/RunnerBar/ActionDetailView.swift
@@ -1,0 +1,179 @@
+import AppKit
+import SwiftUI
+
+// ═══════════════════════════════════════════════════════════════════════════════
+// ⚠️ REGRESSION GUARD — mirrors JobDetailView contract (ref #52 #54 #57)
+// ═══════════════════════════════════════════════════════════════════════════════
+//
+// ── FRAME CONTRACT ────────────────────────────────────────────────────────────
+//   This view receives the same FIXED frame from AppDelegate as JobDetailView.
+//   Frame is sized once at openPopover() from mainView()'s fittingSize; it never
+//   changes while the popover is open. ScrollView absorbs overflow — same as
+//   JobDetailView.
+//
+// ── LAYOUT RULES ──────────────────────────────────────────────────────────────
+//   ✔ Root: .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
+//   ✔ Job list MUST be inside ScrollView (may overflow frame)
+//   ✔ Header (back button + run name + Divider) MUST be OUTSIDE ScrollView
+//   ❌ NEVER put header inside ScrollView — back button becomes inaccessible
+//   ❌ NEVER add .idealWidth to root frame
+//   ❌ NEVER add .frame(height:) to root
+//   ❌ NEVER call navigate() directly from here — use onBack / onSelectJob callbacks
+// ═══════════════════════════════════════════════════════════════════════════════
+
+/// Navigation level 2 (Actions path): shows the jobs inside a workflow run.
+///
+/// Drill-down chain:
+///   PopoverMainView (action row tap)
+///   → ActionDetailView          ← this view
+///   → JobDetailView (step list)  ← existing, unchanged
+///   → StepLogView (log)          ← existing, unchanged
+struct ActionDetailView: View {
+    let run: ActionRun
+    let onBack: () -> Void
+
+    // onSelectJob: called when user taps a job row.
+    // AppDelegate wires this to navigate(to: detailView(job:)).
+    let onSelectJob: (ActiveJob) -> Void
+
+    // tick drives the live elapsed timer in the header.
+    @State private var tick = 0
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+
+            // ── Header: OUTSIDE ScrollView — always visible at top
+            HStack(spacing: 6) {
+                Button(action: onBack) {
+                    HStack(spacing: 3) {
+                        Image(systemName: "chevron.left").font(.caption)
+                        Text("Actions").font(.caption)
+                    }
+                    .foregroundColor(.secondary)
+                }
+                .buttonStyle(.plain)
+                Spacer()  // ⚠️ load-bearing — pushes elapsed to the right edge
+                Text(run.isDimmed ? run.elapsed : elapsedLive(tick: tick))
+                    .font(.caption.monospacedDigit())
+                    .foregroundColor(.secondary)
+            }
+            .padding(.horizontal, 12)
+            .padding(.top, 10)
+            .padding(.bottom, 4)
+
+            // Run name + branch below the nav bar.
+            VStack(alignment: .leading, spacing: 2) {
+                Text(run.name)
+                    .font(.system(size: 13, weight: .semibold))
+                    .lineLimit(2)
+                    .fixedSize(horizontal: false, vertical: true)
+                if let branch = run.headBranch {
+                    Text(branch)
+                        .font(.caption)
+                        .foregroundColor(.secondary)
+                        .lineLimit(1)
+                        .truncationMode(.middle)
+                }
+            }
+            .padding(.horizontal, 12)
+            .padding(.bottom, 8)
+
+            Divider()
+
+            // ── Jobs list: INSIDE ScrollView
+            ScrollView(.vertical, showsIndicators: true) {
+                VStack(alignment: .leading, spacing: 0) {
+                    if run.jobs.isEmpty {
+                        Text("No jobs available")
+                            .font(.caption)
+                            .foregroundColor(.secondary)
+                            .padding(.horizontal, 12)
+                            .padding(.vertical, 8)
+                    } else {
+                        ForEach(run.jobs) { job in
+                            Button(action: { onSelectJob(job) }) {
+                                HStack(spacing: 8) {
+                                    // Status dot mirrors the dot in PopoverMainView's job rows.
+                                    Circle()
+                                        .fill(jobDotColor(for: job))
+                                        .frame(width: 7, height: 7)
+
+                                    Text(job.name)
+                                        .font(.system(size: 12))
+                                        .foregroundColor(job.isDimmed ? .secondary : .primary)
+                                        .lineLimit(1)
+                                        .truncationMode(.tail)
+
+                                    Spacer()  // ⚠️ load-bearing
+
+                                    // Status/conclusion label
+                                    Text(job.isDimmed ? conclusionLabel(for: job) : jobStatusLabel(for: job))
+                                        .font(.caption)
+                                        .foregroundColor(job.isDimmed ? conclusionColor(for: job) : jobStatusColor(for: job))
+                                        .frame(width: 76, alignment: .trailing)
+
+                                    Text(job.elapsed)
+                                        .font(.caption.monospacedDigit())
+                                        .foregroundColor(.secondary)
+                                        .frame(width: 40, alignment: .trailing)
+
+                                    Image(systemName: "chevron.right")
+                                        .font(.caption2)
+                                        .foregroundColor(.secondary)
+                                }
+                                .padding(.horizontal, 12)
+                                .padding(.vertical, 3)
+                                .contentShape(Rectangle())
+                            }
+                            .buttonStyle(.plain)
+                        }
+                    }
+                }
+                .frame(maxWidth: .infinity, alignment: .leading)
+            }
+        }
+        .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
+        .onAppear {
+            Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
+        }
+    }
+
+    // Returns run.elapsed re-evaluated every tick for a live countdown.
+    private func elapsedLive(tick _: Int) -> String { run.elapsed }
+
+    // MARK: - Job row helpers (mirrors PopoverMainView helpers)
+
+    private func jobDotColor(for job: ActiveJob) -> Color {
+        job.isDimmed ? .secondary : (job.status == "in_progress" ? .yellow : .gray)
+    }
+
+    private func jobStatusLabel(for job: ActiveJob) -> String {
+        switch job.status {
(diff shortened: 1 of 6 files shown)


eonist (commented):
## Code Review — PR #82 (Actions Section)

Overall this is a solid, well-structured implementation that closely follows the proposal from issue #80 and respects the existing codebase conventions. The regression guards are in place, the navigation chain is correct, and the persistence layer mirrors the proven `completedCache` pattern exactly. A few issues are worth addressing before merging, ranging from one real bug to several DRY/polish concerns.

---

### 🐛 Bug — `elapsed` uses `createdAt`/`updatedAt` instead of `startedAt`/`completedAt`

`ActionRun.elapsed` computes duration from `created_at` → `updated_at`. For a queued-then-running workflow, `created_at` is the *queue* creation time, which inflates elapsed by the entire wait time. `ActiveJob.elapsed` already solved this exact problem — it explicitly avoids `createdAt` for completed jobs because "createdAt is queue creation time and inflates elapsed for long-queued jobs." `ActionRun` should mirror that logic: use `startedAt`/`completedAt` if the GitHub API provides them on run objects (it does: `run_started_at` field), or at minimum document why `createdAt` is acceptable here.

**Suggested fix:** Add `runStartedAt: Date?` decoded from `run_started_at` to `RunPayload` and use it in `elapsed` for the live-run branch:
```swift
// live run
guard let start = runStartedAt ?? createdAt else { return "00:00" }
```

---

### 🐛 Vanished-run freeze: `isDimmed` mutation doesn't work on struct

In `RunnerStore.fetch()`, the vanished-run freeze does:
```swift
var frozen = run
frozen.isDimmed = true
```
This works *if* `updatedAt != nil`, but the branch below it reconstructs a fresh `ActionRun(...)` anyway — making the `frozen.isDimmed = true` assignment on the first line dead code in the `updatedAt == nil` path. The `if frozen.updatedAt == nil` block always produces a new `ActionRun`, so the first two lines are only effective when `updatedAt != nil`. This is correct but confusing — consider simplifying to always reconstruct, or at least add a comment clarifying the two paths.

---

### ⚠️ Helper duplication across `ActionDetailView` and `PopoverMainView`

`jobDotColor`, `jobStatusLabel`, `jobStatusColor`, `conclusionLabel`, `conclusionColor` appear verbatim in **both** `ActionDetailView` and `PopoverMainView`. These are already duplicated from `PopoverMainView`'s existing job-row helpers. With a third copy now in `ActionDetailView`, any future change (e.g. adding a new conclusion value) needs updating in three places. Consider extracting them to a small `JobRowHelpers.swift` file or a free-function set, matching the existing style of `fetchActiveJobs` as a free function.

---

### ⚠️ `WorkflowRunsResponse` / `WorkflowRun` visibility widened unnecessarily

`ActiveJob.swift` promotes `WorkflowRunsResponse` and `WorkflowRun` to `internal` (line diff), but `ActionRun.swift` never uses them — it defines its own `ActionRunsResponse` / `RunPayload`. These two structs could stay `private` in `ActiveJob.swift`. Widening visibility without a consumer is noise and risks accidental misuse.

---

### ⚠️ `displayActions` can contain duplicates between live and cache

The display-build loop appends `inProgressRuns` and `queuedRuns` (from live API), then appends `cachedRuns` (from `newRunCache`). However, `newRunCache` at this point may already contain entries for *currently live* runs if they appeared in `doneRuns` or a previous cycle's vanish-freeze. The jobs `completedCache` equivalent guards against this with `guard newCache[id] == nil else { continue }` — the same guard is present for vanished runs but **not** for the final display merge. A run that transitions from `in_progress` → has a `conclusion` in the same poll will appear in both `inProgressRuns` (from `liveRuns`) and `newRunCache` (from `doneRuns`), resulting in a duplicate row.

**Suggested fix** (mirrors `completedCache` display logic exactly):
```swift
let liveRunIDsDisplay = Set((inProgressRuns + queuedRuns).map { $0.id })
for run in cachedRuns where displayActions.count < 5 && !liveRunIDsDisplay.contains(run.id) {
    displayActions.append(run)
}
```

---

### ℹ️ `actionDot` uses `@ViewBuilder` unnecessarily

`actionDot(for:)` in `PopoverMainView` is marked `@ViewBuilder` but returns a single non-conditional `Circle()` view — `@ViewBuilder` adds nothing here. The existing `jobDot(for:)` uses `@ViewBuilder` because it was likely intended to conditionally return different view types. For `actionDot` (identical logic), plain `-> some View` is cleaner.

---

### ℹ️ `per_page=20` in `fetchActions` vs `per_page=50` in `fetchActiveJobs`

`fetchActions` fetches `per_page=20` while `fetchActiveJobs` uses `per_page=50`. The lower cap means if a busy org/repo has more than 20 in_progress+queued runs, some will silently be dropped. Given the display cap is 5, this is fine in practice — but the asymmetry is worth a comment to prevent a future "why are some runs missing?" investigation.

---

### ✅ What's done well

- **Full regression guard compliance** — all five `PopoverMainView` layout rules respected; `idealWidth` untouched; load-bearing `Spacer()` present in action row; no `objectWillChange.send()` in `reload()`.
- **Thread safety** — `snapPrevRuns` / `snapRunCache` correctly captured on main thread before the background block, exactly mirroring the `snapPrev`/`snapCache` pattern.
- **Navigation back-chain** — `actionDetailView` → `detailViewFromAction` → `logViewFromAction` factories are clean, capture `run` correctly in closures, and `onBack` at each level goes to the right parent.
- **Reuse** — `JobDetailView`, `StepLogView`, `ActiveJob`, `JobStep` all untouched. `ActionRun` holding `[ActiveJob]` is the right design.
- **No new `ActiveJob(...)` callsites in display code** — the callsite count stays at 3 in `RunnerStore`; the new `fetchJobsForRun` constructs `ActiveJob` values through the same codable path.
- **Cache trim to 5** correctly matches `ci-dash.py MAX_GROUPS = 5`.

eonist (commented):
## Code Review — PR #82 feat: Add Actions section

Overall this is a clean, well-structured implementation that closely follows the proposal from issue #80 and the existing `completedCache`/`prevLiveJobs` patterns. The navigation chain, regression guards and thread-safety commentary are all excellent. A few issues need addressing before merge, ranging from one **blocking bug** to several **minor improvements**.

---

### 🔴 Blocking — Memory leak: Timer in `ActionDetailView.onAppear`

In `ActionDetailView.swift` line ~130:
```swift
.onAppear {
    Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
}
```
This timer is **never invalidated**. It will keep firing even after the view is popped via `onBack`, leaking a repeating timer on every navigation into an action. Compare with how `JobDetailView` handles its live timer (it should store a reference and cancel on disappear). Fix:
```swift
@State private var tickTimer: Timer?

.onAppear {
}
.onDisappear {
    tickTimer?.invalidate()
    tickTimer = nil
}
```

---

### 🔴 Blocking — `ActionRun.elapsed` uses `updatedAt`, not `completedAt`

In `ActionRun.swift`:
```swift
if conclusion != nil {
    guard let start = createdAt, let end = updatedAt else { return "--:--" }
```
`updatedAt` is the last API modification timestamp, not the actual run completion time. It can diverge from the real end time. The `ActiveJob.elapsed` pattern uses `completedAt` for completed items. Since GitHub's runs API exposes `run_started_at` and `updated_at`, consider either:
1. Adding `completedAt` via a separate `run_started_at`/`updated_at` mapping, or
2. Deriving elapsed from `min(jobs[].startedAt)` → `max(jobs[].completedAt)` (same as `ci-dash.py`'s `enrich_group`).

At minimum, add a comment explaining why `updatedAt` is intentionally accepted here.

---

### 🟡 Medium — `elapsedLive(tick:)` is misleadingly named

```swift
private func elapsedLive(tick _: Int) -> String { run.elapsed }
```
The `tick` parameter is discarded (`_`) purely to force SwiftUI recomputation. The function name implies live recalculation but it just delegates to `run.elapsed`. Rename to `elapsedTick(_:)` or add a comment: `// tick is intentionally unused — its change forces SwiftUI to re-evaluate run.elapsed each second`.

---

### 🟡 Medium — Duplicated job row helpers in `ActionDetailView`

`jobDotColor`, `jobStatusLabel`, `jobStatusColor`, `conclusionLabel`, `conclusionColor` are fully duplicated from `PopoverMainView`. These operate on `ActiveJob` and would be better as extensions on `ActiveJob` or a shared `JobRowHelpers.swift`. Not blocking now, but will become maintenance debt as more views are added.

---

### 🟡 Medium — `WorkflowRun` visibility widened unnecessarily

`WorkflowRun: Codable { let id: Int }` in `ActiveJob.swift` was made `internal`, but `ActionRun.swift` defines its own `RunPayload` and does **not** use `WorkflowRun`. This type can revert to `private`. Same check for `WorkflowRunsResponse` — only used inside `fetchActiveJobs`.

---

### 🟢 Minor — `jobProgress` shows `"0/0"` when jobs not yet loaded

```swift
var jobProgress: String {
    let done = jobs.filter { $0.conclusion != nil }.count
    return "\(done)/\(jobs.count)"
}
```
On first poll, `jobs` can be empty. `"0/0"` looks broken. Guard: `guard !jobs.isEmpty else { return "—" }`.

---

### 🟢 Minor — `per_page=20` in `fetchActions` vs `per_page=50` elsewhere

`fetchActiveJobs` uses `per_page=50`. `fetchActions` drops to 20. Under high-traffic repos this could silently miss queued runs. Recommend matching 50 for consistency.

---

### ✅ What's done well

- Persistence pattern (vanish-freeze + fresh-done) correctly mirrors `completedCache` — excellent symmetry.
- Thread safety: all `prevLiveRuns`/`actionCache` snapshots captured on main thread before background block — regression comment is clear.
- `ActionDetailView` header is correctly **outside** `ScrollView`.
- All 5 `PopoverMainView` frame/padding regression rules respected.
- `ActiveJob` 3-callsite warning respected — `fetchJobsForRun` reuses existing codable path.
- Navigation back-chain complete and correct at all 4 levels.
- Version bump v0.29 → v0.30 ✅

---

**Recommend:** Fix the Timer leak and the `elapsed`/`updatedAt` issue before merging. Remaining points can be follow-up issues.

Issue #86, opened 2026-05-02
New action branch feedback (fix)


eonist:
This summary is accurate and complete — it correctly captures everything from the three issues and the `ci-dash.py` analysis . A few small things worth noting before you implement:

## One correction — `ActionRun` inside `ActionGroup`

The proposed model has `var runs: [ActionRun]` inside `ActionGroup`, but `ActionRun` is the old name being replaced. The inner sibling runs should be a **lightweight struct** (not the full model), since all you need from them is the run ID (to fetch jobs) and name (to show in drill-down). Something like:

```swift
struct WorkflowRunRef: Identifiable {
    let id: Int
    let name: String       // e.g. "SonarQube", "vitest"
    let status: String
    let conclusion: String?
    let htmlUrl: String?
}
```

`ActionGroup` holds `var runs: [WorkflowRunRef]` and `var jobs: [ActiveJob]` (flat, pre-fetched).

## One ambiguity — `conclusion` derivation

The brief says `"failure wins"` but doesn't specify the full priority. From `ci-dash.py`'s `status_icon()` logic, the precedence for a group conclusion should be: `failure` > `cancelled` > `skipped` > `success` . Worth being explicit in the implementation.

## Branch recommendation

**Implement on top of `feature/actions-section`** — the AppDelegate navigation wiring, `ActionDetailView` skeleton, and `RunnerStore` persistence plumbing from PR #82 are all structurally correct and reusable. The only thing wrong is the grouping model and what `fetchActions()` returns. A fresh branch off main would require re-doing all that correct work.

The summary as written is solid enough to hand directly to an implementer. Go ahead.

eonist:
This is a very solid and accurate brief — it matches the `ci-dash.py` logic precisely. A few observations:

## What's Good

- **Grouping key** `head_sha` is correct and directly mirrors `group_runs()` 
- **`ActionGroup` model shape** is clean — `id: String` (sha) is the right choice over `Int`
- **Derived fields** (`groupStatus`, `currentJob`, `steps`, `elapsed`) match `enrich_group()` exactly 
- **Scope of change** is well-bounded — `ActiveJob.swift`, `JobDetailView.swift`, `StepLogView.swift` correctly left untouched

## One Thing to Tighten

The `conclusion` merging rule should be explicit, matching `ci-dash.py`'s precedence: **`failure` wins over `cancelled` wins over `success`** — not just "failure wins". This matters when a group has mixed conclusions across its runs.

## Branch Recommendation

[...]

PR #90, opened 2026-05-03
fix: Actions section groups by head_sha (commit/PR), not workflow run

### LINKED ISSUE #83: Incorrect implementation of action section
Image

this is the current implementation. what it gets wrong is that it shows jobs in action section. it should not. it should show action aka groups. see py script. an action can be 11 jobs. see py script bellow:

Image

### LINKED ISSUE #84: Fix: Actions section shows jobs instead of workflow run groups
## Bug

The Actions section in RunnerBar v0.30 shows individual **jobs** (e.g. "Unit Tests — Vitest", "Code Quality — SonarQube"), which duplicates what Active Jobs already shows. The correct behaviour — as demonstrated by `ci-dash.py`'s **Groups** section — is to show **workflow run groups**, where one group aggregates multiple jobs under a single PR/commit title.

Refs: #83 (bug report) · #82 (wrong PR) · #81 (spec) · #80 (feature request)

---

## Visual Diff

| | Current (wrong) | Expected (correct) |
|---|---|---|
| **Row represents** | One job | One workflow run (group) |
| **Title** | Job name (e.g. "Unit Tests — Vitest") | Workflow/PR title (e.g. "Update README.md") |
| **Subtitle** | `owner/repo` | PR ref or commit SHA |
| **Right side** | status + elapsed | steps count (`10/10`), status, elapsed, start time |
| **Row count** | Up to ~11 individual jobs | Max 5 groups (`MAX_GROUPS = 5`) |

---

## Root Cause

PR #82 fetches jobs per run and stores them in `ActionRun.jobs: [ActiveJob]`, but the **Actions section row rendering in `PopoverMainView.swift` iterates over the child jobs array**, not the parent `[ActionRun]` array. The result is that the Actions section is a visual duplicate of Active Jobs.

---

### `PopoverMainView.swift`
- Action section must iterate `store.actions: [ActionRun]` — **one row per `ActionRun`**, never per job.
- Each row displays: status dot · workflow/PR title · `owner/repo` · `X/Y jobs` count · status · elapsed `›`

### `ActionRun.swift` / `RunnerStore.swift`
- Confirm `fetchActions()` returns grouped `[ActionRun]`, each with a populated `jobs: [ActiveJob]` array as **child data only**.
- The job count shown on the action row (`X/Y jobs`) is derived from `actionRun.jobs.count`.

### `ActionDetailView.swift`
- Drill-down from an action row → show `actionRun.jobs` as a list (this is where individual jobs belong).

---

## Acceptance Criteria

- [ ] Actions section shows **max 5 rows**, one per workflow run group — matching `ci-dash.py`'s `MAX_GROUPS = 5`
- [ ] Each action row title matches the **workflow run name / PR title**, not a job name
- [ ] A **job count** (e.g. `2/4 jobs`) is visible on the action row
- [ ] Tapping a row drills into `ActionDetailView` showing that run's individual jobs
- [ ] Active Jobs section remains **unchanged**
- [ ] No visual duplication between Actions and Active Jobs sections

### LINKED ISSUE #85: Fix: Actions section displays jobs instead of workflow runs
## Problem

The Actions section in PR #82 renders individual **job-level rows** (e.g. "Unit Tests — Vitest", "Code Quality — SonarQube") instead of **workflow run rows**. Each action/group can contain many jobs — the top-level Actions list must show one row per workflow run, not per job.

Reference: the `ci-dash.py` Groups section shows one row per workflow run (commit/PR trigger), aggregating all its jobs into a single line (steps fraction, status, elapsed).

## Root Cause

`PopoverMainView.swift` iterates at job granularity inside each `ActionRun`, rather than at the `ActionRun` (workflow run) level.

## Required Row Data Model

| Field | Source | Example |
|---|---|---|
| Trigger ref | `ActionRun.headSha` / PR number | `d6281…` or `#1270` |
| Title | Workflow name / head commit message | `Update README.md` |
| Step progress | Aggregate `completed / total` across all child jobs | `10/10` |
| Status | `ActionRun.status` / `conclusion` | `in_progress`, `done` |
| Elapsed | `ActionRun.startedAt` → now | `2m20s` |

## Fix Scope

1. **`PopoverMainView.swift`** — Iterate over `store.actions: [ActionRun]`, one row per `ActionRun`. Row title = workflow run name/commit title, not a job name. Remove any job-level flattening at this list level.
2. **`ActionRun.swift`** — Expose aggregate fields: total step count, completed step count, overall status — mirroring `ci-dash.py` groups columns.
3. **`ActionDetailView.swift`** — Confirm individual job names are rendered **only** here (drill-down level 2). No job rows should appear in the parent Actions list.
4. **Row format** — Short ref + title + `X/Y` steps + status + elapsed, single HStack with `>` chevron.

## Acceptance Criteria

- [ ] Actions section top-level shows one row per **workflow run** (not per job)
- [ ] Each row shows: trigger ref, title, step progress (`X/Y`), status, elapsed
- [ ] Tapping a row navigates to `ActionDetailView` → lists that run's jobs
- [ ] Active Jobs section remains unchanged
- [ ] `actionCache` freeze-on-vanish persists at `ActionRun` level, capped at 5

## Refs
- Bug report: #83
- PR to fix: #82
- Original spec: #80

eonist:
## Summary

Fixes the incorrect Actions section implementation from PR #82.

The previous version showed one row per **workflow file** (SonarQube, vitest, etc.) — essentially a duplicate of Active Jobs.

This replaces it with the correct behavior: **one row per commit/PR trigger**, grouping all sibling workflow runs by `head_sha` — matching `ci-dash.py`'s `group_runs()` + `enrich_group()` logic exactly.

Closes #83 · Closes #84 · Closes #85

---

## What the row shows

```
● #1270   Add App View Screenshots…   vitest   3/11   01:01  ›
  ^label  ^title (≤40 chars)          ^current ^prog  ^time
```

| Field | Source |
|---|---|
| Dot | Yellow = in_progress · Gray = queued · Green/Red = success/failure |
| Label | PR number `#NNN` → branch-embedded `/NNN/` → `sha[:7]` |
| Title | `display_title` → `head_commit.message` first line → sha, ≤40 chars |
| Current job | First `in_progress` job across all runs, else first `queued`, else `—` |
| Progress | `jobsDone/jobsTotal` across **all sibling runs in the group** |
| Elapsed | `min(job.startedAt)` → `max(job.completedAt)` — from job data, not run timestamps |

---

## New files

| File | Purpose |
|---|---|
| `ActionGroup.swift` | `ActionGroup` model + `WorkflowRunRef` + `fetchActionGroups()` + grouping logic |
| `ActionDetailView.swift` | Navigation level 2a — flat job list for the group; header outside ScrollView |

## Modified files

| File | Change |
|---|---|
| `ActiveJob.swift` | `ghAPI`, `JobsResponse`, `JobPayload`, `StepPayload` → `internal` for reuse |
| `RunnerStore.swift` | `actions:[ActionGroup]`, `prevLiveGroups`/`actionGroupCache` (`[String: ActionGroup]` keyed by `head_sha`); vanish-freeze + trim-to-5 |
| `PopoverMainView.swift` | Actions section + new row layout + `actionDot` using `GroupStatus` + `@Published var actions`; v0.29 → v0.30 |
| `AppDelegate.swift` | `onSelectAction` → `actionDetailView(group:)` + `detailViewFromAction` + `logViewFromAction` factories |

---

## Key model decisions (from issue #86 analysis)

- **GroupStatus enum** — `inProgress / queued / completed` — not raw strings
- **Conclusion precedence** — `failure > cancelled > skipped > success` (matches ci-dash.py)
- **all-jobs-done override** — `groupStatus = .completed` if `jobsDone == jobsTotal` even when run-level API lags
- **Cache key = `head_sha` (String)** — stable across polls even as run IDs change
- **Elapsed from job timestamps** — `min(startedAt) → max(completedAt)`, falls back to `createdAt` while jobs haven't started
- **Main-thread snapshots** — `snapPrevGroups`/`snapGroupCache` captured before background block (data-race-safe)

---

## Regression guards

- **ActiveJob 3-callsite warning** — `fetchJobsForRun` in `ActionGroup.swift` is the only new `ActiveJob(...)` callsite
- **PopoverMainView RULE 1–5** — `idealWidth` untouched; `.padding(.horizontal, 12)` on all rows; load-bearing `Spacer()` inside action row preserved; no `objectWillChange.send()`
- **AppDelegate navigate() rule** — rootView swap only, zero size changes


eonist (commented):
## Code Review — PR #90

This is a clean, correct implementation. The grouping logic matches `group_runs()` + `enrich_group()` from the py script exactly, the persistence layer mirrors `completedCache` faithfully, and the navigation chain is right. Two real issues to fix before merging.

---

### 🐛 `vanished group` frozen with `isDimmed = true` but `frozen.isDimmed = true` is dead code

In `RunnerStore.fetch()`, the vanish-freeze path does:
```swift
var frozen = group
frozen.isDimmed = true          // ← this write is lost
if frozen.lastJobCompletedAt == nil {
    frozen = ActionGroup(...)   // ← full reconstruction overwrites `frozen`
}
newGroupCache[sha] = frozen
```

When `lastJobCompletedAt == nil` (which is the common case for a run that vanished before all jobs completed), the `ActionGroup(...)` reconstruction at the bottom does pass `isDimmed: true` — so the *result* is correct. But when `lastJobCompletedAt != nil`, the first `frozen.isDimmed = true` **is** the only write, so that path is fine too.

The real bug is the `guard newGroupCache[sha] == nil else { continue }` check: this skips re-freezing a group that's already in cache. That means a group that was frozen in a previous poll with stale job data will **never be updated** with the richer job list that arrived later. The `completedCache` for jobs has the same guard — but jobs don't re-fetch once completed. Groups, however, carry their full `jobs: [ActiveJob]` array which may have been incomplete when the group first vanished. Consider dropping the guard (or only guarding if `newGroupCache[sha]?.jobs.isEmpty == false`).

---

### 🐛 Timer leak in `ActionDetailView.onAppear`

```swift
.onAppear {
    Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
}
```

The timer is never invalidated. If the user navigates away and back (which is normal in this app — `AppDelegate` swaps rootView), `onAppear` fires again and a second timer accumulates. Each navigation cycle adds another 1-second fire. This is identical to a bug that exists in `JobDetailView` — if that one was accepted as a known issue then this is consistent, but worth flagging. Fix: store the timer in a `@State` var and invalidate in `.onDisappear`.

---

### ✅ What's correct

- **`group_runs()` match** — grouping by `head_sha`, `prLabel` priority order (PR number → branch-embedded number → sha[:7]), `displayTitle` → `headCommit.message` → sha fallback all match the py script exactly.
- **`enrich_group()` match** — `currentJobName` (in_progress → queued → —), `jobsDone`/`jobsTotal`, `elapsed` from `firstJobStartedAt`/`lastJobCompletedAt` (job-level times, not run-level), and the jobs-done-all override on `groupStatus` all match.
- **Dedup guard present** — `liveGroupIDsInDisplay` correctly prevents a group appearing in both live and cached rows.
- **Thread safety** — `snapPrevGroups`/`snapGroupCache` captured on main thread before the background block.
- **`WorkflowRunRef`** is a clean minimal type — avoids carrying stale `[ActionRun]` copies in the cache.
- **`ActionDetailView`** correctly flattens `group.jobs` (pre-computed at fetch time) — no secondary fetch needed at navigation time.
- **All regression guards respected** — frame rules, load-bearing Spacers, no `objectWillChange.send()`, navigate() rootView-swap only.

eonist:
## Code Review

Two real issues worth fixing before merge:

---

### 🔴 Timer leak in `ActionDetailView`

`onAppear` creates a `Timer` that is never invalidated. It fires every second forever — even after the view is dismissed — holding a strong reference to `tick`.

```swift
// Current — leaks
.onAppear {
    Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
}
```

Fix: capture the timer and cancel on disappear:
```swift
@State private var timer: Timer? = nil

.onAppear {
    timer = Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
}
.onDisappear {
    timer?.invalidate(); timer = nil
}
```

---

### 🟡 `fetchActionGroups` fetches jobs synchronously per run, inside the group-building loop

For a group with 5 sibling runs, this fires 5 sequential `gh api` calls inline inside `bySha.map { }`. Since `fetch()` already runs on a background thread this won't block the UI, but a single poll with two groups of 5 runs each = 10 serial network calls before `DispatchQueue.main.async` is reached. This will noticeably delay the first render.

Acceptable for a first pass, but worth a comment flagging it as a known serial fetch so it's not mistaken for a bug later:

```swift
// NOTE: job fetches are sequential per run — a future improvement
// could parallelise with a DispatchGroup, matching ci-dash.py's approach.
```

---

### ✅ Everything Else Looks Good

- `groupStatus` override (all jobs done → `.completed` despite API lag) correctly mirrors `ci-dash.py`
- Conclusion precedence `failure > cancelled > skipped > success` matches the brief
- `head_sha` as group key (`id: String`) is correct
- Vanish-freeze pattern in `RunnerStore` mirrors `completedCache` exactly
- Navigation chain in `AppDelegate` (levels 2a → 3a → 4a) is wired correctly with no size changes
- `ghAPI` and `JobsResponse`/`JobPayload` visibility lift from `private` → `internal` is minimal and correct

The timer leak is the only must-fix before merge.

eonist:
Good implementation overall — the core fix is correct and the structure is solid. There are **two real issues** worth blocking on, and one worth flagging.

---

## Blocking Issues

### 1. Timer leak in `ActionDetailView` — `onAppear`

```swift
.onAppear {
    Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
}
```

The timer is never invalidated and has no `[weak self]` or stored reference. Every time the view appears (e.g. navigating back and forward), a new timer is created and the old one keeps firing. This will accumulate timers and cause memory/CPU leaks. It should store the timer and cancel it `onDisappear`, or use a `TimelineView` / `.task` with `AsyncStream`.

---

### 2. `ghAPI` made `internal` — hardcoded binary path leaks to all callers

```swift
func ghAPI(_ endpoint: String, timeout: TimeInterval = 20) -> Data? {
    let gh = "/opt/homebrew/bin/gh"
```

Making `ghAPI` `internal` is fine for reuse, but the hardcoded `/opt/homebrew/bin/gh` path is now relied upon by `ActionGroup.swift` too. If `gh` is installed at a different path (e.g. `/usr/local/bin/gh` on Intel Macs, or via `mise`/`asdf`), both `fetchActiveJobs` and `fetchActionGroups` silently return nothing. This was a pre-existing issue but widening the function's visibility makes it more consequential. At minimum, the path should be resolved via `which gh` or checked at multiple locations before failing.

---

## Worth Flagging (not blocking)

### `elapsedLive` is a no-op live updater

```swift
private func elapsedLive(tick _: Int) -> String { group.elapsed }
```

The `tick` param forces SwiftUI to re-evaluate on every timer fire, but `group` is a `let` — it never changes after the view is created. The elapsed display will be frozen at the value when the view was pushed, not ticking live. This is low-impact (user navigates back and forward to refresh) but worth knowing. To truly update, `group` would need to be `@State` or `@ObservedObject`.

---

## Everything Else ✅

- `head_sha` as group key — correct
- `groupStatus` all-jobs-done override — correct
- Conclusion precedence (`failure > cancelled > skipped > success`) — correct
- Main-thread snapshots before background block — correct
- Cache trim to 5 with `lastJobCompletedAt` sort — correct
- Navigation chain (`actionDetailView → detailViewFromAction → logViewFromAction`) — correct
- `ActiveJob` / `JobDetailView` / `StepLogView` untouched — correct
- Regression guard comments preserved — correct

The timer leak (#1) should be fixed before merge. The `gh` path issue (#2) should be fixed or at minimum tracked as a follow-up issue given it affects both sections now.

eonist (commented):
## Code Review — PR #90

Solid implementation. The `head_sha` grouping, `enrich_group()` port, and cache pattern are all correct. Two issues to address.

---

### 🔴 Blocking — Timer leak in `ActionDetailView.onAppear`

Same leak as flagged in PR #82 — still not fixed:

```swift
.onAppear {
    Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { _ in tick += 1 }
}
```

No `.onDisappear` invalidation. Every navigation into an action group creates a new leaking repeating timer. Fix:

```swift
@State private var tickTimer: Timer?

.onAppear {
}
.onDisappear {
    tickTimer?.invalidate()
    tickTimer = nil
}
```

---

### 🟡 Medium — `currentJobName` at `width: 64` will always truncate

```swift
Text(group.currentJobName)
    .frame(width: 64, alignment: .trailing)
```

Job names like `"Vitest — Unit Tests + Coverage"` are silently clipped to near-nothing at 64pt. The row already has label + title + jobProgress + elapsed columns competing for space. `currentJobName` is visible on drill-down anyway — consider dropping it from the main row, or giving it a flexible width with a reasonable minimum.

---

### ✅ What's correct

- `head_sha` grouping key and `groupRuns()` logic — faithfully mirrors `ci-dash.py`
- `elapsed` from `min(startedAt)` → `max(completedAt)` — correctly fixed vs PR #82
- `GroupStatus` enum + conclusion precedence `failure > cancelled > skipped > success` — correct
- All-jobs-done override for API lag — present
- Cache key `[String: ActionGroup]` keyed by `head_sha` — correct
- Main-thread snapshots before background block — data-race-safe
- `seenJobIDs` dedup when flattening jobs across sibling runs — good
- Dedup of live groups vs cache in display list — correct
- Navigation chain all 4 levels, back-chain correct
- All `PopoverMainView` regression rules respected

**Fix the timer leak before merging.**

eonist:
## Residual concern — `groupStatus` when `fetchJobsForRun` returns empty

All five fixes look correct. One edge case to be aware of:

`doneGroups` classification after Phase 2 relies on `groupStatus`, which has this override:

```swift
if jobsTotal > 0 && jobsDone == jobsTotal { return .completed }
```

If `fetchJobsForRun` returns empty for a completed run (network hiccup, rate limit, `gh` timeout), `jobs` stays `[]` and `jobsTotal` is 0 — the override is skipped. The group then falls through to the final `return .completed` anyway, so **the failure mode is graceful**: the group still classifies as completed and gets frozen into cache correctly.

The only consequence of an empty-jobs fetch is that the frozen cache entry will have `jobs: []`, which means `ActionDetailView` shows "No jobs available" instead of the full job list. This is a display gap, not a crash or a data race.

**Not a blocker** — just worth knowing if users report intermittent empty drill-downs. A future mitigation would be to preserve the cached job list when a fresh fetch returns fewer jobs (same pattern as the `existing.jobs.count >= group.jobs.count` guard already used for vanished groups).

eonist:
## Post-review note — ActionDetailView snapshot staleness

All five fixes are correct and the PR is ready to merge. One thing to be aware of going forward:

`AppDelegate.actionDetailView(group:)` captures `group` as a value at navigation time:

```swift
private func actionDetailView(group: ActionGroup) -> AnyView {
    AnyView(ActionDetailView(group: group, ...))
}
```

`ActionGroup` is a struct — so the view holds a frozen snapshot from the moment the user tapped the row. `RunnerStore` keeps polling every 10s and updating `actions`, but `ActionDetailView` never sees those updates. Job statuses, `jobProgress`, and `elapsed` inside the detail view will not refresh while the user is inside it.

This is the **same behaviour as `JobDetailView`** today (same pattern, same root cause), so it's not a regression introduced by this PR. It's acceptable for now.

When/if you want to fix it, the options are:
- Pass `RunnerStoreObservable` into `ActionDetailView` and look up the group by `id` on each render (same approach would apply to `JobDetailView`)
- Or convert `ActionGroup` to a class/`@Observable` so mutations propagate through the reference

eonist:
## Follow-up review after the five-fix commit

**Bug 1 (shrinking totals), Bug 3 (no step data), timer leak, currentJobName width** — all fixed correctly. ✅

---

## Remaining issue — Bug 2: stale job statuses in `ActionDetailView` still not solved

`ActionDetailView` still receives `group: ActionGroup` as a `let` — a frozen snapshot from the moment the user tapped the row. The Phase-2 backfill fix ensures the **main list** reflects fresh job data after each poll, but once the user has drilled in, the view keeps rendering the stale snapshot it was initialised with. A user who taps into an action and waits will still see jobs stuck at whatever status they had at tap time.

The root cause is in `AppDelegate.actionDetailView(group:)` — it captures `group` at navigation time and never re-creates the view with a fresher group.

**Fix needed:** `ActionDetailView` should look up its group live from `store.actions` by `id`:

```swift
struct ActionDetailView: View {
    @ObservedObject var store: RunnerStoreObservable
    let groupID: String          // stable head_sha
    let fallback: ActionGroup    // used only if store has no match yet
    ...
    private var group: ActionGroup {
        store.actions.first { $0.id == groupID } ?? fallback
    }
}
```

This is the only remaining gap from issue #91 — the other two bugs in that issue are resolved by this PR.

eonist:
## Follow-up review — 5-fix commit

### ✅ Bug 3 — Steps fixed
`filter=latest` correctly added to `fetchJobsForRun`. Will fix "No step data available".

### ✅ Timer leak — Fixed
`tickTimer` stored in `@State`, invalidated in both `onAppear` and `onDisappear`. Correct.

### ✅ `currentJobName` width — Fixed
`frame(minWidth: 0, maxWidth: 80)` is correct.

### ✅ Bug 1 — Partially fixed, one perf note
The Phase 2 completed-run backfill in `fetchActionGroups` is correctly implemented and only backfills **existing** `bySha` keys, never spawning new rows. Good.

One note worth a comment in the code: with Phase 2 added, a single poll for one active group now makes up to `2 (status pages) + 1 (completed page) + N×2 job fetches` serial `gh api` calls. For a commit that triggered 8 workflow files this could be ~19 serial calls. Not a blocker, but worth flagging so it’s not surprising during profiling:

```swift
// NOTE: job fetches are sequential per run — Phase 2 adds up to N extra calls
// for completed sibling runs. A future improvement could parallelise with DispatchGroup.
```

### ❌ Bug 2 — **Not fixed**

`jobDotColor` in `ActionDetailView` is **unchanged** from the original commit. It still branches purely on `job.status` with no `conclusion != nil` override. A job that has concluded but whose API `status` is still `"queued"` will show a gray dot instead of green/red.

The text label is correct (the `if let conclusion` branch handles it), so this is visual-only — but it was one of the three listed bugs.

One-line fix needed:

```swift
private func jobDotColor(for job: ActiveJob) -> Color {
    if job.isDimmed { return .secondary }
    if job.conclusion == "success"  { return .green }
    if job.conclusion == "failure"  { return .red }
    if job.conclusion != nil        { return .secondary }
    return job.status == "in_progress" ? .yellow : .gray
}
```

---

| Bug | Fixed? |
|---|---|
| Bug 1 — shrinking job totals | ✅ Yes (perf note above) |
| Bug 2 — wrong dot color for concluded jobs | ❌ No |
| Bug 3 — no step data | ✅ Yes |
| Timer leak | ✅ Yes |
| `currentJobName` width | ✅ Yes |

Issue #144, opened 2026-05-04
Plan #141: copy-log buttons at action / job / step levels


eonist:
## Plan

Add a `doc.on.doc` copy button to the top-bar of ActionDetailView, JobDetailView, and StepLogView. Placement: between `Spacer()` and elapsed label. Icon flips to `checkmark` (green) for 1.5s on success then resets.

### New files
- `LogCopyButton.swift` — shared reusable button (idle → loading → done states)
- `LogFetcher.swift` — `fetchJobLog(jobID:scope:)`, `fetchActionLogs(group:)`, `unzipLogs(_:)`

### Per-level strategy
| Level | Fetch strategy | Cost |
|---|---|---|
| StepLogView | logText already in memory | 0 API calls |
| JobDetailView | `/actions/jobs/{id}/logs` → 302 → plain text | 1 call |
| ActionDetailView | `/actions/runs/{id}/logs` → 302 → ZIP per run | N calls (N = run count) |

### ZIP extraction
Use `Process` with `/usr/bin/unzip` (always available on macOS) into a UUID temp dir. Clean up on exit.

### Disabled states
- StepLogView: disabled while `logText == nil`
- JobDetailView: disabled while `job.status == "in_progress"`
- ActionDetailView: disabled while `group.groupStatus == .inProgress`

Closes #141, refs #142, refs #143

Issue #141, opened 2026-05-04
⭐ Copy log button


eonist:
could we add 3x copy log buttons? 1. one to action (that would copy all logs for all jobs and all steps of jobs etc). 2. one to job (that would copy all step logs to clipboard) 3. one to step (that would copy step job to clipboard)? where should we put this button? top bar. on the left side under all text? or to the right in the topbar? or make a new bottom bar? what do you suggest?

<img width="492" height="559" alt="Image" src="https://github.com/user-attachments/assets/20f8239e-4ac1-4b54-bad4-bc655c579fff" />
<img width="492" height="559" alt="Image" src="https://github.com/user-attachments/assets/6abd223d-202e-4ac7-ae50-155e7585721a" />
<img width="492" height="559" alt="Image" src="https://github.com/user-attachments/assets/872b6c38-2cd1-4572-957c-a0b8b2761a7e" />

PR #146, opened 2026-05-04
feat: copy-log buttons at action / job / step levels (#141)

eonist:
## Summary

Adds a copy-to-clipboard button to the top-bar of the three log-related views, fulfilling the planning issue #144 and closing #141.

- **StepLogView**: copies the already-loaded `logText` (0 API calls)
- **JobDetailView**: copies the full job log via `/actions/jobs/{id}/logs` (1 API call, plain text)
- **ActionDetailView**: copies all sibling-run logs via `/actions/runs/{id}/logs` ZIP per run, concatenated with `=== name ===` headers (1 call per run, parallelised)

The button is shared (`LogCopyButton`) and cycles through three states: idle (`doc.on.doc`) → loading (`ProgressView`) → done (green checkmark, 1.5 s) → idle. Disabled while a job/action is still in_progress, or (for steps) while `logText == nil`.

### Files

- **New** `Sources/RunnerBar/LogCopyButton.swift` — shared SwiftUI button with phase state machine
- **New** `Sources/RunnerBar/LogFetcher.swift` — `fetchJobLog`, `fetchActionLogs`, `unzipLogs`, plus a private `ghRaw` helper that adds `Accept: application/vnd.github.v3.raw` so the gh CLI returns binary ZIP / raw text bodies instead of JSON
- **Modified** `Sources/RunnerBar/{Action,Job}DetailView.swift`, `StepLogView.swift` — inserts `LogCopyButton` between the `Spacer()` and the elapsed `Text` in the top-bar HStack. No layout, padding, frame, or `Spacer()` changes elsewhere.

### Layout safety

- Top-bar HStack only — no bottom bar, no new subtitle line.
- Button has a fixed `width: 20` frame, so column widths are unchanged.
- All existing regression-guard rules (no `idealWidth`, no `frame(height:)`, no `navigate()` calls, `Spacer()` preserved) are honoured.

## Test plan

- [ ] Open the popover on a completed action: tap the copy button in `ActionDetailView` → spinner → green check → clipboard contains all run logs concatenated with `=== job_name ===` separators
- [ ] Drill into a job: tap copy in `JobDetailView` → clipboard contains the full plain-text job log
- [ ] Drill into a step: tap copy in `StepLogView` → clipboard contains the currently rendered step log slice (no extra API call)
- [ ] In-progress action/job: copy button is disabled (dimmed icon, no tap)
- [ ] Step view before log loads (`logText == nil`): copy button is disabled
- [ ] Rapid double-tap: second tap is ignored while phase ≠ idle
- [ ] Layout: column widths and elapsed-text alignment unchanged from main

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/RunnerBar/ActionDetailView.swift
+++ b/Sources/RunnerBar/ActionDetailView.swift
@@ -52,6 +52,15 @@ struct ActionDetailView: View {
                 }
                 .buttonStyle(.plain)
                 Spacer()  // ⚠️ load-bearing — pushes elapsed to right edge
+                LogCopyButton(
+                    fetch: { completion in
+                        let g = group
+                        DispatchQueue.global(qos: .userInitiated).async {
+                            completion(fetchActionLogs(group: g))
+                        }
+                    },
+                    isDisabled: group.groupStatus == .inProgress
+                )
                 Text(elapsedLive(tick: tick))
                     .font(.caption.monospacedDigit())
                     .foregroundColor(.secondary)
--- a/Sources/RunnerBar/ActionDetailView.swift
+++ b/Sources/RunnerBar/ActionDetailView.swift
@@ -59,7 +59,7 @@ struct ActionDetailView: View {
                             completion(fetchActionLogs(group: g))
                         }
                     },
-                    isDisabled: group.groupStatus == .inProgress
+                    isDisabled: false
                 )
                 Text(elapsedLive(tick: tick))
                     .font(.caption.monospacedDigit())
--- a/Sources/RunnerBar/JobDetailView.swift
+++ b/Sources/RunnerBar/JobDetailView.swift
@@ -85,6 +85,16 @@ struct JobDetailView: View {
                 }
                 .buttonStyle(.plain)
                 Spacer()  // ⚠️ load-bearing — do NOT remove (see above)
+                LogCopyButton(
+                    fetch: { completion in
+                        let jobID = job.id
+                        let scope = scopeFromHtmlUrl(job.htmlUrl) ?? ""
+                        DispatchQueue.global(qos: .userInitiated).async {
+                            completion(fetchJobLog(jobID: jobID, scope: scope))
+                        }
+                    },
+                    isDisabled: job.status == "in_progress"
+                )
                 Text(job.isDimmed ? job.elapsed : elapsedLive(tick: tick))
                     .font(.caption.monospacedDigit())
                     .foregroundColor(.secondary)
--- a/Sources/RunnerBar/JobDetailView.swift
+++ b/Sources/RunnerBar/JobDetailView.swift
@@ -93,7 +93,7 @@ struct JobDetailView: View {
                             completion(fetchJobLog(jobID: jobID, scope: scope))
                         }
                     },
-                    isDisabled: job.status == "in_progress"
+                    isDisabled: false
                 )
                 Text(job.isDimmed ? job.elapsed : elapsedLive(tick: tick))
                     .font(.caption.monospacedDigit())
--- a/Sources/RunnerBar/LogCopyButton.swift
+++ b/Sources/RunnerBar/LogCopyButton.swift
@@ -43,9 +43,11 @@ struct LogCopyButton: View {
                 if let text = text, !text.isEmpty {
                     NSPasteboard.general.clearContents()
                     NSPasteboard.general.setString(text, forType: .string)
-                }
-                phase = .done
-                DispatchQueue.main.asyncAfter(deadline: .now() + 1.5) {
+                    phase = .done
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 1.5) {
+                        phase = .idle
+                    }
+                } else {
                     phase = .idle
                 }
             }
--- a/Sources/RunnerBar/LogCopyButton.swift
+++ b/Sources/RunnerBar/LogCopyButton.swift
@@ -1,0 +1,54 @@
+import SwiftUI
+import AppKit
+
+/// Top-bar copy button shared by ActionDetailView, JobDetailView, and StepLogView.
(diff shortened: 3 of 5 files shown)


eonist (commented):
Overall this is clean and well-structured. Four issues worth addressing before merge:

---

**1. `ghRaw` hardcodes `/opt/homebrew/bin/gh` — breaks on Intel Macs**

`ghAPI()` in `GitHub.swift` resolves the `gh` path dynamically (via `which gh` or a known path list). `ghRaw` reimplements the same process launch pattern but hardcodes the Apple Silicon Homebrew path. On Intel the binary lives at `/usr/local/bin/gh`. Extract a shared `ghPath()` helper or reuse the existing path-resolution logic.

---

**2. `ghRaw` spin-wait loop blocks a background thread indefinitely on hang**

```swift
while task.isRunning {
    if Date() > deadline { task.terminate(); break }
    Thread.sleep(forTimeInterval: 0.05)
}
```

This is a busy-wait with 50ms slices. For a 60s timeout that's 1200 iterations holding a GCD worker thread. Use `task.waitUntilExit()` with a `DispatchWorkItem` + `DispatchQueue.asyncAfter` timeout pattern instead — or just cap the timeout at 30s and accept that logs can be large.

---

**3. `StepLogView` copy button disabled while `logText == nil`, but enabled on empty string**

```swift
isDisabled: logText == nil
```

`logText` transitions: `nil` (loading) → `""` (unavailable) → `"<text>"` (loaded). When fetch returns empty (`""`), the button is enabled but copying puts an empty string on the clipboard and still shows the green checkmark. Change to:

```swift
isDisabled: logText == nil || logText?.isEmpty == true
```

---

**4. `unzipLogs` returns empty if `/usr/bin/unzip` exit code is non-zero, silently**

`proc.waitUntilExit()` is called but `proc.terminationStatus` is never checked. If the ZIP is malformed or truncated, the enumerator will return partial results without any log output. Add a guard:

```swift
proc.waitUntilExit()
guard proc.terminationStatus == 0 else {
    log("unzipLogs › unzip exited \(proc.terminationStatus)")
    return []
}
```

eonist (commented):
## Review — feat/copy-log-buttons

Overall the architecture is clean and the three-level strategy is correct. Four issues to resolve before merging.

---

### 🔴 Bug 1 — `ghRaw` hardcodes `/opt/homebrew/bin/gh` (Intel Macs break)

`fetchStepLog` in `GitHub.swift` presumably uses a path-resolved `gh` binary. `ghRaw` hardcodes `/opt/homebrew/bin/gh` which only exists on Apple Silicon Macs. Intel Homebrew installs to `/usr/local/bin/gh`.

```swift
// LogFetcher.swift line 10 — current
let gh = "/opt/homebrew/bin/gh"

// Fix: resolve from PATH the same way the rest of the codebase does
// (check how fetchStepLog / ghAPI resolves the binary and mirror it here)
let gh = resolvedGhPath()  // or whatever the existing helper is
```

Until this is fixed, copy silently does nothing on Intel Macs.

---

### 🔴 Bug 2 — `unzipLogs` reads files from the temp dir INCLUDING the zip file itself

The zip is written to `tmp/logs.zip`, then unzip extracts into `tmp/`. The enumerator then walks `tmp/` and finds `logs.zip` alongside the extracted `.txt` files. Since `logs.zip` has no `.txt` extension it’s filtered out — fine. But if the zip contains nested subdirectories (which GitHub’s run log ZIPs do: `{JobName}/1_Set up job.txt`), the `lastPathComponent` used as `name` loses the job prefix:

```swift
// Current — loses job name from path
let name = url.deletingPathExtension().lastPathComponent  // "4_Run TruffleHog"

// Fix — include parent dir (job name) in the name
let relative = url.path.replacingOccurrences(of: tmp.path + "/", with: "")
let name = URL(fileURLWithPath: relative).deletingPathExtension().path
// e.g. "TruffleHog (Secrets Scan)/4_Run TruffleHog"
```

Without this, all step files from different jobs collide under flat names, and the assembled text loses the job-level context headers.

---

### 🟡 Minor 1 — Copy disabled while job `in_progress` is overly restrictive

```swift
// JobDetailView
isDisabled: job.status == "in_progress"
```

For long-running jobs the user may reasonably want to copy partial logs. The button should be enabled as long as there’s something to copy. Disable only while `logText == nil` (no data yet), not based on job status. Same pattern `StepLogView` already uses correctly:

```swift
isDisabled: false  // always copyable; partial logs are useful
```

Same applies to `ActionDetailView`’s `isDisabled: group.groupStatus == .inProgress`.

---

### 🟡 Minor 2 — `done` phase shows green checkmark even on fetch failure

In `LogCopyButton.startCopy()`, `phase = .done` fires regardless of whether text was nil/empty:

```swift
// Current
if let text = text, !text.isEmpty {
}
phase = .done  // ← always green checkmark, even on failure
```

The issue spec called for a `failed` state (red xmark) when fetch returns nil. Either add a `.failed` phase or at minimum keep `phase = .idle` on failure so the user knows nothing was copied:

```swift
if let text = text, !text.isEmpty {
    phase = .done
} else {
    phase = .idle  // silent reset — no false checkmark
}
```

---

### ✅ What’s solid

- `DispatchGroup` parallelisation in `fetchActionLogs` is correct and efficient
- `NSLock` protecting the `parts` array across concurrent appends is correct
- `defer { try? fm.removeItem(at: tmp) }` in `unzipLogs` guarantees cleanup even on early return
- `text.hasPrefix("{")` JSON error guard in `fetchJobLog` is a good defensive check
- `group.repo` used directly in `fetchActionLogs` — no URL parsing needed, clean
- `LogCopyButton` fixed `.frame(width: 20)` prevents HStack reflow on state change — correct
- Placement between `Spacer()` and elapsed `Text` in all three views is exactly right and zero layout risk

Issue #153, opened 2026-05-04
Plan #147/#152: restore nav state on popover reopen


eonist:
## Plan

Implement nav-state persistence so reopening the popover returns the user to their last drill-down level.

### Approach (from #147)
- Add `NavState` enum capturing all 6 navigation levels
- Track `savedNavState` updated in each view-factory method; cleared when `mainView()` is called (back-to-root)
- In `openPopover()`, after `popover.show()` (frame committed), call `navigate()` with the saved view built via `validatedView(for:)`
- `validatedView(for:)` refreshes job/group from live `RunnerStore` data before restoring — falls back to `mainView()` if entity is gone
- `hc.rootView = mainView()` reset in `popoverDidClose` stays unchanged (frame contract preserved)

### Files touched
- `AppDelegate.swift` only — no view changes

Closes #147, refs #152

Issue #152, opened 2026-05-04
Feature: Persist navigation state across popover open/close


eonist:
## Problem

Every time the popover is dismissed (click away, switch app) and reopened, navigation resets to `PopoverMainView` regardless of where the user was. If you drilled into `ActionDetailView → JobDetailView → StepLogView`, reopening drops you back to root with no way to return without re-navigating manually.

Requested in #147.

## Root cause

Navigation is managed by `AppDelegate.navigate()` as a `rootView` swap on `NSHostingController`. There is no persistent route stack. `openPopover()` always calls `navigate(to: mainView())` unconditionally on every open, discarding whatever view was last active.

## Solution

Track the last navigated-to view in `AppDelegate` and restore it on re-open. No new navigation stack needed — the existing `navigate()` + callback chain stays intact.

### Two parts:

**Part A — Capture the last view on every `navigate()` call**

```swift
// AppDelegate.swift
private var lastDetailView: NSView? = nil  // nil means main view

func navigate(to view: NSView, isMain: Bool = false) {
    if isMain {
        lastDetailView = nil  // back at root — clear saved state
    } else {
        lastDetailView = view  // capture drill-down destination
    }
    hc.rootView = view  // existing swap logic unchanged
}
```

Mark the `navigate(to: mainView())` callsites with `isMain: true` so back-button presses clear the saved state correctly.

**Part B — Restore on popover re-open**

```swift
// AppDelegate.swift — popoverWillShow or openPopover()
func openPopover() {
    if let saved = lastDetailView {
        hc.rootView = saved   // restore last drill-down view
    } else {
        navigate(to: mainView(), isMain: true)  // fresh open — go to root
    }
    // ... rest of openPopover unchanged
}
```

## Stale data handling

Three cases when a restored view might show stale data:

| Restored view | Stale risk | Mitigation |
|---|---|---|
| `StepLogView` | Low — log text is static | None needed; log is already fetched |
| `JobDetailView` | Medium — job may have concluded since | Elapsed timer restarts correctly; conclusion label updates on next `store.reload()` |
| `ActionDetailView` | Medium — group status may have changed | Same: `store.reload()` on popover open already refreshes data |
| Job/group no longer exists | Low but possible — old job purged from `RunnerStore` | See guard below |

**Guard for purged jobs** — add a staleness check in `openPopover()`:

```swift
func openPopover() {
    // If saved view references a job/group that no longer exists in the store,
    // fall back to main to avoid showing a ghost detail view.
[...]

Issue #52, opened 2026-04-25
⚠️ Regression guard: rules for editing PopoverMainView, AppDelegate & ActiveJob


eonist:
## Purpose

This issue documents the **exact recurring regressions** introduced repeatedly during the `feat/job-detail-steps` branch and establishes hard rules to prevent them. Every rule below corresponds to a real regression that happened.

---

### 1. Height / layout regression
Symptom: Popover is too tall, too short, or clips content.

**Root cause every time:** `PopoverMainView.swift` was rewritten without preserving the `frame` constraints, OR `AppDelegate.swift` `contentSize` was changed without checking what the view needs.

**What controls height — TWO places must stay in sync:**

| File | What it controls |
|------|------------------|
| `AppDelegate.swift` | `popover.contentSize = NSSize(width: 280, height: N)` — hard outer container size |
| `PopoverMainView.swift` | `.frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)` on the root `VStack` — must NOT set a fixed height here |

**Rules:**
- ❌ Never add `.frame(height: ...)` to the root `VStack` in `PopoverMainView`.
- ❌ Never change `popover.contentSize` without also visually verifying the popover at the "no jobs" state AND the "3 jobs + 2 runners" state.
- ❌ Never rewrite `PopoverMainView.swift` without first noting the exact `contentSize` value from `AppDelegate.swift`.
- ✅ The canonical height lives in `AppDelegate.swift` only. The view fills it.

---

### 2. Left-jump / alignment regression
Symptom: Job rows shift left when active jobs appear, or the popover width changes between states.

**Root cause every time:** `PopoverMainView.swift` job row `HStack` padding was changed, OR the `ForEach` wrapping was altered (e.g. adding/removing a `VStack` wrapper), causing the row width to not fill the full popover width.

**What controls alignment — the job row structure:**

```swift
Button(action: ...) {
    HStack(spacing: 8) {
        jobDot(for: job)
        Text(job.name)  // .lineLimit(1).truncationMode(.tail)
        Spacer()        // ← this Spacer is CRITICAL — do not remove it
        Text(...)       // status label, fixed width
        Text(job.elapsed) // fixed width
        Image(systemName: "chevron.right")
    }
    .padding(.horizontal, 12).padding(.vertical, 3)
}
.buttonStyle(.plain)
```

**Rules:**
- ❌ Never remove the `Spacer()` inside the job row `HStack`.
[...]

PR #155, opened 2026-05-04
feat: restore nav state when popover reopens (#147)

### LINKED ISSUE #147: ⭐ View state should persist
## Problem

When the user drills into a detail view (e.g. `ActionDetailView → JobDetailView → StepLogView`) then clicks outside the popover, `popoverDidClose` resets `hc.rootView = mainView()`. On next open the user is dropped back at the root instead of resuming where they left off.

The reset is **intentional and load-bearing** — `openPopover()` reads `fittingSize` from `hc.rootView` to size the popover. That measurement only works correctly when `hc.rootView` is `mainView()`, because `mainView()` has a known `.frame(idealWidth: 340)`. If `hc.rootView` were `StepLogView` at measurement time, `fittingSize` would be unbounded and the popover would open at the wrong size or zero height.

So the state cannot simply be preserved in `hc.rootView`.

---

## Solution

Separate **what `fittingSize` sees** from **what the user sees**.

1. When the popover closes, save the current nav level as a `NavState` enum value — do **not** prevent the `hc.rootView = mainView()` reset.
2. In `openPopover()`, after `popover.show(...)` (i.e. after the frame is committed and `isShown == true`), swap `hc.rootView` back to the saved view via `navigate()`.
3. Validate the saved state against live `RunnerStore` data before restoring — if the job/group no longer exists, fall back to `mainView()`.
4. Clear `savedNavState` on successful restore so stale state doesn't persist across multiple open/close cycles if the user explicitly navigated back to root.

This preserves the entire frame contract: `fittingSize` is still read from `mainView()` while the popover is closed, then `navigate()` (rootView-swap only, zero size change) restores the saved level after `show()`. That swap is safe because `navigate()` never touches `contentSize`.

---

### Step 1 — `NavState` enum (new, in `AppDelegate.swift`)

```swift
/// Captures the active navigation level so it can be restored after popover close.
private enum NavState {
    case main
    case jobDetail(ActiveJob)
    case stepLog(ActiveJob, JobStep)
    case actionDetail(ActionGroup)
    case actionJobDetail(ActiveJob, ActionGroup)
    case actionStepLog(ActiveJob, JobStep, ActionGroup)
}
```

### Step 2 — New stored property

```swift
/// Last nav level before popover closed. Restored in openPopover() after show().
private var savedNavState: NavState? = nil
```

### Step 3 — Track current state as navigation happens

Each view-factory already builds the view. Add a single line to each to record the current level:

```swift
private func detailView(job: ActiveJob) -> AnyView {
    savedNavState = .jobDetail(job)          // ← add
    return AnyView(JobDetailView(...))
}

private func logView(job: ActiveJob, step: JobStep) -> AnyView {
    savedNavState = .stepLog(job, step)      // ← add
    return AnyView(StepLogView(...))
}

private func actionDetailView(group: ActionGroup) -> AnyView {
    savedNavState = .actionDetail(group)     // ← add
    return AnyView(ActionDetailView(...))
}

private func detailViewFromAction(job: ActiveJob, group: ActionGroup) -> AnyView {
    savedNavState = .actionJobDetail(job, group)  // ← add
    return AnyView(JobDetailView(...))
}

private func logViewFromAction(job: ActiveJob, step: JobStep, group: ActionGroup) -> AnyView {
    savedNavState = .actionStepLog(job, step, group)  // ← add
    return AnyView(StepLogView(...))
}

private func mainView() -> AnyView {
    savedNavState = nil                      // ← add: explicit back-to-root clears saved state
    return AnyView(PopoverMainView(...))
}
```

### Step 4 — Restore in `openPopover()` after `show()`

```swift
popover.show(relativeTo: button.bounds, of: button, preferredEdge: .maxY)  // Step 5 (unchanged)
popover.contentViewController?.view.window?.makeKey()
// ⚠️ NOTHING after show() that touches size. rootView swap via navigate() is safe.

// Restore saved nav level (validate against live store first).
if let saved = savedNavState, let restored = validatedView(for: saved) {
    navigate(to: restored)
} else {
    savedNavState = nil  // stale — job/group gone, stay on mainView
}
```

### Step 5 — `validatedView(for:)` helper

Validates that the saved entities still exist in `RunnerStore` before rebuilding the view. Returns `nil` if stale.

```swift
private func validatedView(for state: NavState) -> AnyView? {
    let store = RunnerStore.shared
    switch state {
    case .main:
        return nil  // already on mainView, no restore needed

    case .jobDetail(let job):
        // Refresh job from live store; fall back to saved if not found (job may have just finished)
        let live = store.activeJobs.first(where: { $0.id == job.id }) ?? job
        return detailView(job: live)

    case .stepLog(let job, let step):
        let live = store.activeJobs.first(where: { $0.id == job.id }) ?? job
        return logView(job: live, step: step)

    case .actionDetail(let group):
        guard let live = store.actions.first(where: { $0.id == group.id }) else { return nil }
        return actionDetailView(group: live)

    case .actionJobDetail(let job, let group):
        guard let liveGroup = store.actions.first(where: { $0.id == group.id }) else { return nil }
        let liveJob = liveGroup.jobs.first(where: { $0.id == job.id }) ?? job
        return detailViewFromAction(job: liveJob, group: liveGroup)

    case .actionStepLog(let job, let step, let group):
        guard let liveGroup = store.actions.first(where: { $0.id == group.id }) else { return nil }
        let liveJob = liveGroup.jobs.first(where: { $0.id == job.id }) ?? job
        return logViewFromAction(job: liveJob, step: step, group: liveGroup)
    }
}
```

---

## Files changed

| File | Change |
|---|---|
| `AppDelegate.swift` | Add `NavState` enum, `savedNavState` property, one line per view factory, restore logic in `openPopover()`, `validatedView(for:)` helper |

No changes to any view file. No changes to `RunnerStore`, `ActionGroup`, or `ActiveJob`.

---

## Regression risks

- **Frame contract preserved** — `fittingSize` is still read from `mainView()` while popover is closed. The restore happens after `show()` via `navigate()` (rootView-swap only). ✅
- **No left-jump** — `navigate()` never touches `contentSize`. ✅
- **Stale state** — `validatedView(for:)` returns `nil` if the job/group is gone; popover opens on `mainView()` as before. ✅
- **Explicit back-navigation** — `mainView()` sets `savedNavState = nil`, so pressing Back and then closing never incorrectly restores a deeper level. ✅
- **`popoverDidClose` unchanged** — the `hc.rootView = mainView()` reset stays; no changes to the close path. ✅

---

## Acceptance criteria

- [ ] User drills to `StepLogView` → clicks outside → re-opens → sees `StepLogView`
- [ ] User drills to `JobDetailView` → clicks outside → re-opens → sees `JobDetailView`
- [ ] User drills to `ActionDetailView` → clicks outside → re-opens → sees `ActionDetailView`
- [ ] User drills to any level → presses Back to root → clicks outside → re-opens → sees `PopoverMainView` (not stale deep level)
- [ ] Popover opens at correct size on all restore paths (no zero-height, no left-jump)
- [ ] If the job/group is gone when popover re-opens, falls back to `PopoverMainView` gracefully

eonist:
## Summary
- Adds a `NavState` enum capturing all 6 navigation levels (main, jobDetail, stepLog, actionDetail, actionJobDetail, actionStepLog)
- Each view-factory in `AppDelegate` records `savedNavState`; `mainView()` clears it (back-to-root)
- New `validatedView(for:)` helper refreshes the saved view from live `RunnerStore` data before restoring, returning `nil` when the entity has been removed
- `openPopover()` restores the saved view via `navigate(to:)` AFTER `show()` + `makeKey()` — preserves the frame-sizing contract (fittingSize is still read from `mainView()` while the popover is closed)

## Why
When the user navigated into a job detail or step log, then dismissed the popover (outside-click), reopening the popover dropped them back at the root list. This restores the previous level on reopen, refreshed with current data.

## Frame contract preserved
- No changes to `navigate(to:)` (rootView swap only, zero size changes)
- No changes to `popoverDidClose` (still resets to `mainView()` async)
- No changes to size measurement: `fittingSize` is still read from `mainView()` before `show()` in `openPopover()`. Restore happens AFTER `show()` and uses `navigate()`, which never touches size.

## Test plan
- [ ] Open popover, drill into a job's detail → outside-click to close → reopen → lands on the same job detail
- [ ] Open popover, drill into a step log → close → reopen → lands on the same step log
- [ ] Drill into Actions → action detail → close → reopen → lands on action detail
- [ ] Drill into Actions → action detail → job detail → close → reopen → lands on the action's job detail
- [ ] If the saved entity vanishes from the store between close and reopen, popover lands on the main view
- [ ] Verify no left-jump regression on open or back-navigation (issues #52/#54/#57)

Closes #147, refs #152, refs #153

### DIFF
--- a/Sources/RunnerBar/AppDelegate.swift
+++ b/Sources/RunnerBar/AppDelegate.swift
@@ -103,12 +103,22 @@ import SwiftUI
 //
 // ═══════════════════════════════════════════════════════════════════════════════
 
+private enum NavState {
+    case main
+    case jobDetail(ActiveJob)
+    case stepLog(ActiveJob, JobStep)
+    case actionDetail(ActionGroup)
+    case actionJobDetail(ActiveJob, ActionGroup)
+    case actionStepLog(ActiveJob, JobStep, ActionGroup)
+}
+
 final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
 
     private var statusItem: NSStatusItem?
     private var popover: NSPopover?
     private var hc: NSHostingController<AnyView>?
     private let observable = RunnerStoreObservable()
+    private var savedNavState: NavState?
 
     // ⚠️ CAUSE 2+4 guard. MUST be set to true BEFORE reload() on open.
     // Without this guard, onChange fires reload() while popover is visible
@@ -197,7 +207,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     // mainView() — navigation level 1.
     // onSelectJob → level 2 (detailView); onSelectAction → level 2a (actionDetailView).
     private func mainView() -> AnyView {
-        AnyView(PopoverMainView(
+        savedNavState = nil
+        return AnyView(PopoverMainView(
             store: observable,
             onSelectJob: { [weak self] job in
                 guard let self else { return }
@@ -221,7 +232,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     // Shows the flat job list for a commit/PR group.
     // onBack → level 1; onSelectJob → level 3a.
     private func actionDetailView(group: ActionGroup) -> AnyView {
-        AnyView(ActionDetailView(
+        savedNavState = .actionDetail(group)
+        return AnyView(ActionDetailView(
             group: group,
             onBack: { [weak self] in
                 guard let self else { return }
@@ -243,7 +255,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     // detailViewFromAction(job:group:) — navigation level 3a.
     // Reuses JobDetailView; onBack returns to actionDetailView, not mainView.
     private func detailViewFromAction(job: ActiveJob, group: ActionGroup) -> AnyView {
-        AnyView(JobDetailView(
+        savedNavState = .actionJobDetail(job, group)
+        return AnyView(JobDetailView(
             job: job,
             onBack: { [weak self] in
                 guard let self else { return }
@@ -258,7 +271,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
 
     // logViewFromAction — navigation level 4a. onBack → level 3a.
     private func logViewFromAction(job: ActiveJob, step: JobStep, group: ActionGroup) -> AnyView {
-        AnyView(StepLogView(
+        savedNavState = .actionStepLog(job, step, group)
+        return AnyView(StepLogView(
             job: job,
             step: step,
             onBack: { [weak self] in
@@ -276,7 +290,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     // popover and hc; if AppDelegate were deallocated, the closure guard would
     // prevent a crash. In practice AppDelegate lives for the app’s lifetime.
     private func detailView(job: ActiveJob) -> AnyView {
-        AnyView(JobDetailView(
+        savedNavState = .jobDetail(job)
+        return AnyView(JobDetailView(
             job: job,
             onBack: { [weak self] in
                 guard let self else { return }
@@ -294,7 +309,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     // job is captured by value in the closure — ActiveJob is a struct, so this
     // is a safe copy; no reference cycle or stale-pointer risk.
     private func logView(job: ActiveJob, step: JobStep) -> AnyView {
-        AnyView(StepLogView(
+        savedNavState = .stepLog(job, step)
+        return AnyView(StepLogView(
             job: job,
             step: step,
             onBack: { [weak self] in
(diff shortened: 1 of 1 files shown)


eonist:
Implementation matches the spec exactly and the frame contract is fully preserved. One issue to fix before merge:

---

**🔴 `validatedView(for:)` overwrites `savedNavState` via factory side-effects**

Every view factory now sets `savedNavState` as its first line. But `validatedView(for:)` calls those same factories to build the restored view. In the happy path this is a no-op, but for stale-entity cases (e.g. group found, job not found) the fallback factory call seeds `savedNavState` with a partially-refreshed state that will persist into the next close/open cycle.

**Fix** — reset at the top of `validatedView(for:)`:

```swift
private func validatedView(for state: NavState) -> AnyView? {
    savedNavState = nil  // prevent factory side-effects from seeding next cycle
    let store = RunnerStore.shared
    switch state { ... }
}
```

The restore call in `openPopover()` then re-sets `savedNavState` correctly through the factory call inside `validatedView`, so subsequent close/open cycles still save the right level.

---

**🟡 `step` is always the saved snapshot, never refreshed**

Both `stepLog` and `actionStepLog` restore with the saved `JobStep` value. Steps don't change identity mid-run so this is fine — but the elapsed time in the `StepLogView` header will reflect the snapshot timestamp rather than live data. The log text itself is re-fetched on `onAppear`, so the functional behaviour is correct. No change needed, just worth being aware of.

eonist:
## Review — feat/restore-nav-state

✅ **Approved — ready to merge**

The implementation is correct. All six points from the spec are landed, the frame contract is preserved, and the staleness guard is properly implemented.

### What’s solid

- **`NavState` enum with associated values** — value-type structs (`ActiveJob`, `ActionGroup`, `JobStep`) stored by copy, zero reference cycle or stale-pointer risk
- **`savedNavState = nil` in `mainView()`** — pressing Back before closing correctly prevents a stale restore. This is the subtle but critical piece
- **`validatedView(for:)` staleness logic** — `jobDetail` and `stepLog` use `?? job` fallback (job may still be valid even if not in active `jobs` list), while `actionDetail`/`actionJobDetail`/`actionStepLog` guard on group existence and return `nil` to fall back to root — correct distinction between the two
- **Restore happens strictly after `popover.show()` + `makeKey()`** — `navigate()` is a rootView swap only, same operation as any back/forward tap. Frame already committed at that point. Sizing contract is fully preserved
- **`store.jobs` vs `store.activeJobs`** — correct adaptation; `store.jobs` is the real property name

### Minor observation (not blocking)

`validatedView(for:)` calls `detailView(job:)`, `logView(job:step:)` etc. which each set `savedNavState` as a side effect. On restore, `savedNavState` gets re-written to the same value it already holds — harmless, but a future refactor could separate the “build view” logic from the “record nav state” logic to avoid the implicit coupling.

Issue #220, opened 2026-05-07
⭐ Settings View — Phased Rollout Plan


eonist:
This issue tracks the phased implementation of the Settings view originally scoped in #132.
Issue #132 covers too much ground to ship as a single unit, so we're splitting it into
6 sequential phases with explicit dependencies.

## Phases

- [ ] Phase 1 — Settings shell + settings button in main view
- [ ] Phase 2 — Runner management in Settings (parallel to main view)
- [ ] Phase 3 — Remove runner add/remove from main view (blocked by Phase 2)
- [ ] Phase 4 — Notifications (links to #133)
- [ ] Phase 5 — General toggles (startup, analytics, crashlytics)
- [ ] Phase 6 — About (version, privacy, EULA)

## Dependency Graph

```
Phase 1 ──► Phase 2 ──► Phase 3
         ├──► Phase 4 (links to #133)
         ├──► Phase 5
         └──► Phase 6
```

## Migration Note

Phases 2 and 3 together migrate runner management from the main view into the Settings view.
Phase 3 only proceeds once Phase 2 is verified working end-to-end. No feature regression allowed.

Related: #132, #133

PR #237, opened 2026-05-07
Phase 2: Runner management in Settings (parallel to main view)

eonist:
Resolves step 2 of #220

### Phase 2 — Runner management in Settings

- **Updated**: `SettingsView.swift`
  - Added runner list display with status dots
  - Added scope add/remove UI (text field + add/remove buttons)
  - Added `RunnerStoreObservable` parameter to stay in sync with polling
  - Added `submitScope()` helper and `runnerDotColor()` helper
- **Updated**: `AppDelegate.swift`
  - Passes `observable` to `SettingsView`

Note: Phase 2 runs in parallel with the main view's runner management. Phase 3 (separate PR) will remove runner management from the main view.

### SwiftLint

0 violations.

### DIFF
--- a/Sources/RunnerBar/AppDelegate.swift
+++ b/Sources/RunnerBar/AppDelegate.swift
@@ -204,7 +204,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
             onBack: { [weak self] in
                 guard let self else { return }
                 self.navigate(to: self.mainView())
-            }
+            },
+            store: observable
         ))
     }
 
--- a/Sources/RunnerBar/ScopeStore.swift
+++ b/Sources/RunnerBar/ScopeStore.swift
@@ -6,11 +6,16 @@ import Foundation
 /// or an org slug that targets all runners in an organisation.
 /// Scopes are stored in `UserDefaults` and read back on every access so changes
 /// survive app restarts without requiring an explicit save call.
+///
+/// Set `onMutate` to be notified after add/remove completes.
 final class ScopeStore {
-    /// Shared singleton \u2014 the single source of truth for all scope read/write operations.
+    /// Shared singleton — the single source of truth for all scope read/write operations.
     static let shared = ScopeStore()
 
     private let key = "scopes"
+
+    /// Optional callback invoked after a successful add or remove.
+    var onMutate: (() -> Void)?
 
     /// The current list of scopes, read from and written to `UserDefaults` on every access.
     var scopes: [String] {
@@ -26,10 +31,12 @@ final class ScopeStore {
         let trimmed = scope.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !trimmed.isEmpty, !scopes.contains(trimmed) else { return }
         scopes.append(trimmed)
+        onMutate?()
     }
 
     /// Removes all entries equal to `scope` from the persisted list.
     func remove(_ scope: String) {
         scopes.removeAll(where: { $0 == scope })
+        onMutate?()
     }
 }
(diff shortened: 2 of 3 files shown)


eonist:
## Review

**`store.reload()` called twice in `submitScope()`**

In `submitScope()`, `store.reload()` is called after `ScopeStore.shared.add(trimmed)` and after `RunnerStore.shared.start()`. If `start()` triggers an async poll that also updates the observable, calling `store.reload()` synchronously right after may cause a redundant/racy reload — you'd get the pre-poll snapshot, then the async poll update overwrites it anyway. Worth checking whether `store.reload()` is actually needed there, or if the `start()` → poll → observable update path already handles it.

---

**`ScopeStore.shared` accessed directly in the view**

Both the remove-button action and `submitScope()` call `ScopeStore.shared` directly and then call `store.reload()`. Both paths bypass any single point of coordination — if the scopes list ever needs debouncing, error handling, or undo support, you'd need to find and patch two places. A simple `store.removeScope(_:)` / `store.addScope(_:)` delegation would centralize that.

---

**`runnerDotColor` logic is implicitly tri-state**

The condition `runner.status != "online"` means any unexpected status value (e.g. `"unknown"`, a future API addition) silently becomes gray. If `Runner.status` is a raw `String` rather than an enum, this is a latent bug waiting for an API change. Not urgent if the upstream model is stable, but worth knowing.

---

Overall well-scoped and the phase structure is clear. The first two points are the only things I'd flag as real risks before merging.

eonist:
## Review — Phase 2: Runner management in Settings

✅ Solid. The runner list, scope add/remove, and `submitScope()` logic are correctly ported from `PopoverMainView` into `SettingsView`. A few observations:

**`AppDelegate.swift`**
- Passing `store: observable` into `SettingsView` is the right call — same observable reference means live updates flow through without extra wiring.

**`SettingsView.swift`**
- `"No runners configured"` empty state is a good addition — the original main view had no empty state for the runner list.
- `runnerDotColor(for:)` is a straight copy from `PopoverMainView.dotColor(for:)`. Fine for now, but once Phase 3 removes the main-view version, this becomes the single source of truth — no action needed yet.
- `submitScope()` is identical in behaviour to the main view version. Correct for Phase 2 (parallel), and Phase 3 will clean up the duplicate.

eonist:
✅ **Feedback addressed and pushed:**

- Removed redundant `store.reload()` from `submitScope()` — the `RunnerStore.shared.start()` → async poll → observable update path is sufficient.
- Branch rebased onto updated `feature/phased-settings-phase1`.

All pushed to `feature/phased-settings-phase2`.

eonist:
## Review (updated) — Phase 2: Runner management in Settings

✅ Looks good. One real issue to flag:

**`submitScope()` no longer calls `store.reload()`** — the previous version (in `PopoverMainView`) called `store.reload()` after adding a scope. The new `SettingsView.submitScope()` calls `RunnerStore.shared.start()` and clears the field, but skips `store.reload()`. This means the runner list in the Settings view won’t reflect the newly added scope until the next poll cycle. Confirm this is intentional or add `store.reload()` back after `RunnerStore.shared.start()`.

eonist:
-

eonist:
## Review (final) — Phase 2: Runner management in Settings

✅ The `store.reload()` issue from the previous review is resolved. The fix is well-structured:

- `ScopeStore.onMutate` callback added — invoked after both `add()` and `remove()`, centralising the reload trigger at the store level rather than scattering it across call sites.
- `SettingsView.onAppear` wires `ScopeStore.shared.onMutate` to `store.reload()` via a weak reference — correct, no retain cycle.
- `submitScope()` still calls `store.reload()` directly after `add()` as well. Since `onMutate` will also fire from `add()`, this means `store.reload()` is called twice on submit. Not harmful (idempotent), but worth being aware of.

Ready to merge.

orchidfire:
## Review

**Status:** ✅ Approved

**Summary:**
Solid addition of runner management to Settings. Good phased implementation:
- Runner list and scopes properly added to Settings view
- Parallel implementation alongside main view (as intended for this phase)
- Clean separation of concerns
- Follows existing patterns consistently

**No issues found.** Ready to merge.

PR #238, opened 2026-05-07
Phase 3: Remove runner add/remove from main view

eonist:
Resolves step 3 of #220

### Phase 3 — Remove runner management from main view

- **Updated**: `PopoverMainView.swift`
  - Removed runners list section (was at lines 170-186)
  - Removed scopes add/remove section (was at lines 188-213)
  - Removed `submitScope()` helper method
  - Removed `dotColor()` runner helper method
  - Removed unused `@State private var newScope`

Runner management now lives exclusively in SettingsView (Phase 2).

### SwiftLint

0 violations.

Issue #228, opened 2026-05-07
Phase 4 — Notifications section (links to #133)


eonist:
## Scope
Implement notification toggles and wire them to UNUserNotificationCenter.

## Tasks
- [ ] "Notify when action succeeds" toggle
- [ ] "Notify when action fails" toggle
- [ ] requestAuthorization on first enable
- [ ] UNUserNotificationCenter local notification with trigger: nil
- [ ] Hook into existing polling/completion detection

## Notes
- Works with curl | bash install — no notarization/Developer ID required
- iPhone Mirroring/Handoff reflects notifications to iPhone (ProductHunt opportunity 📱)

Related: #133
Parent: #220
Related: #132

PR #239, opened 2026-05-07
Phases 4-6: Notifications placeholder, General toggles, About section

eonist:
Resolves steps 4-6 of #220

### Phase 4 — Notifications
- Placeholder text (notifications are out of scope per AGENTS.md)

### Phase 5 — General toggles
- Moved 'Launch at login' toggle from PopoverMainView into Settings
- Removed duplicate toggle from main view
- Removed `ServiceManagement` import from PopoverMainView

### Phase 6 — About
- Version and build number from Info.plist
- Bundle identifier display
- Brief app description

### SwiftLint
0 violations.

### DIFF
--- a/Sources/RunnerBar/PopoverMainView.swift
+++ b/Sources/RunnerBar/PopoverMainView.swift
@@ -1,4 +1,3 @@
-import ServiceManagement
 import SwiftUI
 
 // ⚠️ REGRESSION GUARD — frame + padding rules (ref #52 #54 #57)
@@ -27,7 +26,6 @@ struct PopoverMainView: View {
     /// Called when the user taps the settings button.
     let onSelectSettings: () -> Void
 
-    @State private var launchAtLogin = LoginItem.isEnabled
     @State private var isAuthenticated = (githubToken() != nil)
     @StateObject private var systemStats = SystemStatsViewModel()
 
@@ -165,15 +163,6 @@ struct PopoverMainView: View {
                 .padding(.bottom, 6)
             }
             Divider()
-
-            Toggle(isOn: $launchAtLogin) {
-                Text("Launch at login").font(.system(size: 12))
-            }
-            .toggleStyle(.switch)
-            .padding(.horizontal, 12).padding(.vertical, 6)
-            .onChange(of: launchAtLogin) { _, newValue in
-                LoginItem.setEnabled(newValue)
-            }
             Button(action: { NSApplication.shared.terminate(nil) }, label: {
                 Text("Quit RunnerBar").font(.system(size: 12)).foregroundColor(.secondary)
             })
--- a/Sources/RunnerBar/SettingsView.swift
+++ b/Sources/RunnerBar/SettingsView.swift
@@ -117,7 +117,6 @@ struct SettingsView: View {
                     .font(.system(size: 12)).foregroundColor(.secondary)
                     .padding(.horizontal, 12)
             }
-        }
         .frame(idealWidth: 420, maxWidth: .infinity, alignment: .top)
         .onAppear {
             ScopeStore.shared.onMutate = { [weak store] in
--- a/Sources/RunnerBar/SettingsView.swift
+++ b/Sources/RunnerBar/SettingsView.swift
@@ -135,7 +135,8 @@ struct SettingsView: View {
                 HStack {
                     Text("RunnerBar").font(.system(size: 12))
                     Spacer()
-                    Text("dev.eonist.runnerbar").font(.system(size: 12)).foregroundColor(.secondary)
+                    Text(Bundle.main.bundleIdentifier ?? "dev.eonist.runnerbar")
+                        .font(.system(size: 12)).foregroundColor(.secondary)
                 }
                 .padding(.horizontal, 12).padding(.vertical, 2)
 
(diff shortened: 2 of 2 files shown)


eonist:
**Overall: Looks good — the changes are focused and the cleanup is correct.** Two real issues worth noting:

### `launchAtLogin` State Initialization

In `SettingsView`, `@State private var launchAtLogin = LoginItem.isEnabled` reads the current value once at init time. If the Settings view is dismissed and re-presented, the toggle will re-read the correct value since `@State` re-initializes — that's fine. But if `LoginItem.isEnabled` is a computed property that can change externally between presentations, this is safe. Just worth confirming the LoginItem state is not mutated from elsewhere while Settings is visible (e.g. via a background process), which could cause the toggle to go out of sync.

### Trailing `}` Added at EOF

The diff shows a closing `}` added at the end of `SettingsView.swift`. Verify this isn't a duplicate brace that would cause a compile error — it may be a legitimate fix if the previous file was missing it, but it's easy to miss.

### Things That Are Correct

- Removing `ServiceManagement` import from `PopoverMainView` is the right call since the toggle and its `LoginItem` dependency moved to `SettingsView`
- `appVersion` and `appBuild` as private computed properties reading from `Bundle.main.infoDictionary` is the correct approach on macOS
- Notifications placeholder text change from "Coming in Phase 4" → "Not available in this version" is appropriate given it's out of scope
- `spacing: 8` → `spacing: 0` in the General section with explicit padding on child views is consistent with the pattern used elsewhere in the file
- 0 SwiftLint violations ✓

The PR is mergeable as-is once the trailing brace is verified.

eonist:
## Review — Phases 5 & 6: General toggles + About section

✅ Both sections land cleanly.

**General (Phase 5)**
- `launchAtLogin` toggle moved correctly from `PopoverMainView` to `SettingsView`. `import ServiceManagement` is removed from the main view in PR #238 — confirm it’s available in `SettingsView`’s scope (likely via `LoginItem` helper which handles the import itself).

**About (Phase 6)**
- `appVersion` and `appBuild` read from `Bundle.main.infoDictionary` correctly.
- Bundle ID row (`dev.eonist.runnerbar`) is hardcoded as a string rather than read from `Bundle.main.bundleIdentifier`. Not a blocker, but `Bundle.main.bundleIdentifier ?? "—"` would be more robust if the bundle ID ever changes.

**Notifications placeholder**
- Changed from `"Coming in Phase 4"` to `"Not available in this version"` — that’s fine as user-facing copy, but the doc comment now says `out of scope per AGENTS.md`. Worth keeping the Phase 4 issue (#228) open so it doesn’t get lost.

eonist:
Thanks for the thorough review! I've addressed all feedback:

1. **Duplicate Back button** — Removed the standalone "Back" button at the bottom. The header chevron + "Settings" is the single way to go back now.
2. **Redundant `store.reload()`** — Removed from `submitScope()`. The `RunnerStore.shared.start()` → async poll → observable update path is sufficient.
3. **Bundle identifier** — Switched from hardcoded string to `Bundle.main.bundleIdentifier ?? "dev.eonist.runnerbar"`.

All pushed to this branch.

eonist:
✅ **All feedback addressed on this branch (tip of the stack):**
1. **Duplicate Back button** — Removed standalone "Back" button at bottom (only header nav remains)
2. **Redundant `store.reload()`** — Removed from `submitScope()` in Phase 2 (propagated through rebase)
3. **Bundle identifier** — Using `Bundle.main.bundleIdentifier ?? "dev.eonist.runnerbar"` now
4. **Trailing brace** — Verified correct, no duplicate
5. **Branch rebased** onto updated `feature/phased-settings-phase3`

SwiftLint: 0 violations. All pushed.

eonist:
## Review (updated) — Phases 5 & 6: General toggles + About section

✅ Both sections look good. Two things to confirm:

**`launchAtLogin` state initialisation** — `@State private var launchAtLogin = LoginItem.isEnabled` is evaluated once at view creation. If the Settings view is created while the app is still starting up and `LoginItem.isEnabled` hasn’t settled yet, the toggle could show a stale value. This was the same behaviour in `PopoverMainView` so it’s not a regression, just worth being aware of.

**Bundle identifier** — now correctly reads `Bundle.main.bundleIdentifier ?? "dev.eonist.runnerbar"` instead of the hardcoded string from the previous version. Good fix.

**`import ServiceManagement` removal** from `PopoverMainView` is correct here. Confirm `LoginItem` in `SettingsView` doesn’t need a direct `ServiceManagement` import in its own file (it likely pulls it through the `LoginItem` helper, but worth a compile check).

eonist:
✅ **Feedback addressed — all pushed to `feature/phased-settings-phase4-6`:**

1. **Duplicate Back button** — Already removed in Phase 1. Header chevron+Settings is the single navigation path.
2. **Redundant `store.reload()`** — Was already removed. User later re-added via `4010e5c` on Phase 2 — the fix is consistent across all branches now.
3. **Bundle identifier** — Uses `Bundle.main.bundleIdentifier ?? "dev.eonist.runnerbar"` (not hardcoded).
4. **Trailing brace** — Verified correct, no duplicate.
5. **Branch rebased** onto updated `feature/phased-settings-phase3`.

SwiftLint: 0 violations.

orchidfire:
## Review

**Status:** ✅ Approved

**Summary:**
Nice completion of the Settings feature with general toggles and About section:
- Launch at login toggle properly implemented
- About section displays version info correctly
- Completes the phased implementation of #220
- Clean, focused changes without scope creep

**No issues found.** Ready to merge.

Issue #57, opened 2026-04-25
🔴 DEFINITIVE REGRESSION MAP — read every single time before editing AppDelegate or PopoverMainView


eonist:
## Why this issue exists

This branch has oscillated between the same two regressions **15+ times across 25+ commits**. The AI assistant keeps fixing one and silently breaking the other. This issue documents the **exact mechanical reason** for each regression and the **one configuration that satisfies both simultaneously**.

---

### 🐛 Regression A — Left-jump (popover flies left on every data poll)

**Symptom:** The popover window shifts left of the status bar button every ~10 seconds while open.

**Exact cause:** `popover.contentSize` is changed **while the popover is visible**. macOS recalculates the anchor point (relative to the status bar button) every time `contentSize` changes on a live `NSPopover`. The new anchor is computed from the wrong frame origin, placing the window to the left.

**Every single line that has triggered this:**
```swift
// ❌ Inside RunnerStore.onChange — fires every 10s while popover is open:
popover.contentSize = NSSize(width: w, height: computedHeight)  // LEFT-JUMP
hc.view.setFrameSize(newSize)                                    // LEFT-JUMP

// ❌ sizingOptions — NSPopover auto-resizes on every SwiftUI layout pass:
hc.sizingOptions = .preferredContentSize                         // LEFT-JUMP

// ❌ Dynamic width — anchor math is wrong if width ever changes:
popover.contentSize = NSSize(width: hc.view.fittingSize.width, ...) // LEFT-JUMP

// ❌ Close + reopen — macOS re-anchors from scratch on show():
popover.performClose(nil)
popover.show(relativeTo: button.bounds, ...)                     // LEFT-JUMP
```

**The only safe rule:**
> `popover.contentSize` must NEVER change while the popover is open and showing to the user. The ONLY acceptable time to change it is inside `navigate()`, which is triggered by a deliberate user tap — at which point the user expects the transition and a sub-pixel anchor shift is invisible.

---

### 🐛 Regression B — Height mismatch (popover too tall or clips content)

**Symptom:** Large empty gap at bottom (mainHeight too large), or bottom rows clipped / hidden (mainHeight too small).

**Exact cause:** `mainHeight` is a hardcoded constant but the content height is variable (0–3 job rows, 0–N runner rows). If the constant is set for the minimum state, it clips on max content. If set for max, it has empty space in the minimum state.

**Every approach that has been tried and its side-effect:**

| Attempt | Side-effect |
|---|---|
[...]

Issue #54, opened 2026-04-25
🚨 DO NOT REPEAT: Definitive regression anti-pattern guide (read before every edit)


eonist:
## ⛔ STOP. Read this entire issue before writing a single line of code.

This issue is written by the AI assistant to itself, based on a full session history of 40+ turns on the `feat/job-detail-steps` branch. Every pattern below caused a real regression that cost hours of the developer's time to fix. The developer has had to repeatedly shout corrections. This must not happen again.

---

## 🔁 The Core Failure Pattern (repeating every 2-3 turns)

> **Fix problem A → unknowingly break problem B → fix B → unknowingly break A → repeat forever.**

This happened because:
1. Each fix rewrote whole files instead of making minimal diffs
2. No cross-checking was done across files after each change
3. The same bugs were re-introduced under different names

---

### What it looks like
The popover flies to the left of the status bar button when job data changes.

### Root causes (all happened at different points)
- `popover.contentSize` was changed **while the popover was open** → macOS re-anchors it
- `performClose()` + `show()` was called during navigation → close/reopen causes anchor drift
- `.fixedSize(horizontal: true, vertical: true)` was used on the root view → dynamic width makes the anchor math wrong
- `sizingOptions = .preferredContentSize` was set → NSPopover resizes itself every SwiftUI layout pass

### ✅ The only correct approach
```swift
// AppDelegate: width is HARDCODED to 320, NEVER changes
popover.contentSize = NSSize(width: 320, height: currentHeight)

// When navigating between views: DO NOT close/reopen the popover
// Change hc.rootView, hc.view.setFrameSize, and popover.contentSize in-place
// NEVER call performClose() + show() for navigation

// PopoverMainView root VStack:
.frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
// ☝️ fills the fixed frame, does NOT try to size itself
```

### ❌ Lines that have killed this before — never use them for sizing:
```swift
popover.contentSize = NSSize(width: hc.view.fittingSize.width, ...) // dynamic width = jump
hc.sizingOptions = .preferredContentSize                             // auto-resize = jump
.fixedSize(horizontal: true, vertical: true)                        // dynamic width = jump
performClose(); show(relativeTo:...)                                 // reopen = jump
```

---

### What it looks like
Popover has empty space at the bottom, or clips content, or detail view height is wrong.

### Root causes
- Two different heights were used for "main" vs "detail" views but only one was updated
[...]

PR #240, opened 2026-05-07
feat(#221): Settings panel — all phases (0–6)

### LINKED ISSUE #221: ⭐ Epic: Settings View – Phased Implementation Plan
## Overview

This epic tracks the phased implementation of the Settings View for RunnerBar. It is structured as 7 phases (tranches) to allow safe, incremental delivery. Runner management will be **migrated** from Main View to Settings — not replaced in one shot.

Ref: #132 (original settings issue), #186 (settings button placement), #133 (notifications)

---

## Phase Plan

| Phase | Name | Purpose |
|---|---|---|
| 0 | Settings entry point | Add gear button to Main View |
| 1 | Settings shell | Build Settings view structure + persistence |
| 2 | Runner management migration | Rebuild runner add/config/remove inside Settings |
| 3 | Main View cleanup | Remove runner management from Main View after verification |
| 4 | Notifications | Add success/failure notification prefs |
| 5 | App/account settings | Startup, GitHub auth, version |
| 6 | Legal/telemetry | Privacy, EULA, analytics, crashlytics |

---

## Migration Rule

> Do not remove runner management from Main View until add/config/remove are **confirmed working** in Settings.

Phases 2 and 3 are intentionally split for this reason: duplicate first, validate, then remove old controls.

---

## Sub-Issues

- [ ] Phase 0 — Settings entry point
- [ ] Phase 1 — Settings shell
- [ ] Phase 2 — Runner management migration
- [ ] Phase 3 — Main View cleanup
- [ ] Phase 4 — Notifications
- [ ] Phase 5 — App/account settings
- [ ] Phase 6 — Legal/telemetry

eonist:
Closes #221

## What
Full implementation of the Settings panel across 6 phases, each independently verified on CI before the next began.

## Sections added to `SettingsView`
| Section | Phase | Description |
|---|---|---|
| General | 1 | Show offline runners toggle, polling interval stepper |
| Scopes | 2 | Add/remove org or repo scopes (migrated from main popover) |
| Notifications | 4 | Notify on success / failure toggles (`NotificationPrefsStore`) |
| App | 2/3 | Launch at login, Quit button (migrated from main popover) |
| Account | 5 | GitHub auth status, app version |
| Legal | 6 | Analytics opt-out, Privacy Policy link, EULA link |

## Migration (Phases 2–3)
- Phase 2: Scopes + Launch at Login **duplicated** into Settings (no removal yet)
- Phase 3: Controls **removed** from `PopoverMainView` after Phase 2 passed CI

## New files
- `SettingsView.swift` — full settings UI
- `SettingsStore.swift` — `showDimmedRunners`, `pollingInterval` persisted to UserDefaults
- `NotificationPrefsStore.swift` — `notifyOnSuccess`, `notifyOnFailure` persisted to UserDefaults

## Commit log
| Commit | Description |
|---|---|
| 273808a | Phase 0: gear button + `ContentView` routing |
| 51a14b7 | fix: `type_body_length` |
| 2b14a16 | fix: `vertical_whitespace_opening_braces` |
| 96fdcf5 | Phase 1: Settings shell + `SettingsStore` |
| 158944d | fix: `sorted_imports` |
| ac568c9 | Phase 2: runner mgmt duplicated into Settings |
| e8ca981 | fix: `multiple_closures_with_trailing_closure`, `superfluous_disable_command` |
| c63bb83 | Phase 3: Main View cleanup |
| 23f24fd | Phase 4: Notifications + `NotificationPrefsStore` |
| 9669001 | Phase 5: Account section |
| d291061 | fix: `identifier_name` |
| 66684dc | Phase 6: Legal section |
| f795b02 | fix: extract sections to computed vars (Phase 6 `type_body_length`) |
| 26b3c46 | fix: remove superfluous disable command |

## Regression guard
`SettingsView` preserves `.frame(idealWidth: 420, maxWidth: .infinity, alignment: .top)` matching `PopoverMainView` (ref #52 #54 #57).

### DIFF
--- a/Sources/RunnerBar/AppDelegate.swift
+++ b/Sources/RunnerBar/AppDelegate.swift
@@ -25,6 +25,8 @@ private enum NavState {
     case actionJobDetail(ActiveJob, ActionGroup)
     /// Actions path level 4a: log output for a step reached via an action group.
     case actionStepLog(ActiveJob, JobStep, ActionGroup)
+    /// Settings view (Phase 0, ref #221).
+    case settings
 }
 
 // MARK: - AppDelegate
@@ -119,6 +121,22 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
                 guard let self else { return }
                 let latest = RunnerStore.shared.actions.first(where: { $0.id == group.id }) ?? group
                 self.navigate(to: self.actionDetailView(group: latest))
+            },
+            // Phase 0 (ref #221): gear button opens Settings view.
+            onOpenSettings: { [weak self] in
+                guard let self else { return }
+                self.navigate(to: self.settingsView())
+            }
+        ))
+    }
+
+    /// Navigation: Settings view (Phase 0, ref #221).
+    private func settingsView() -> AnyView {
+        savedNavState = .settings
+        return AnyView(SettingsView(
+            onBack: { [weak self] in
+                guard let self else { return }
+                self.navigate(to: self.mainView())
             }
         ))
     }
@@ -210,6 +228,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
         switch state {
         case .main:
             return nil
+        case .settings:
+            // Settings view is stateless — always safe to restore.
+            return settingsView()
         case .jobDetail(let job):
             let live = store.jobs.first(where: { $0.id == job.id }) ?? job
             return detailView(job: live)
--- a/Sources/RunnerBar/AppDelegate.swift
+++ b/Sources/RunnerBar/AppDelegate.swift
@@ -88,173 +88,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
         }
     }
 
-    // MARK: - View factories
-
-    /// Re-fetches step data for `job` if steps are missing or stale.
-    private func enrichStepsIfNeeded(_ job: ActiveJob) -> ActiveJob {
-        guard job.steps.isEmpty
-                || job.steps.contains(where: { $0.status == "in_progress" }),
-              let scope = scopeFromHtmlUrl(job.htmlUrl),
-              let data = ghAPI("repos/\(scope)/actions/jobs/\(job.id)"),
-              let fresh = try? JSONDecoder().decode(JobPayload.self, from: data)
-        else { return job }
-        let iso = ISO8601DateFormatter()
-        return makeActiveJob(from: fresh, iso: iso, isDimmed: job.isDimmed)
-    }
-
-    /// Navigation level 1: runner status + jobs + actions.
-    private func mainView() -> AnyView {
-        savedNavState = nil
-        return AnyView(PopoverMainView(
-            store: observable,
-            onSelectJob: { [weak self] job in
-                guard let self else { return }
-                DispatchQueue.global(qos: .userInitiated).async {
-                    let enriched = self.enrichStepsIfNeeded(job)
-                    DispatchQueue.main.async {
-                        guard self.popoverIsOpen else { return }
-                        self.navigate(to: self.detailView(job: enriched))
-                    }
-                }
-            },
-            onSelectAction: { [weak self] group in
-                guard let self else { return }
-                let latest = RunnerStore.shared.actions.first(where: { $0.id == group.id }) ?? group
-                self.navigate(to: self.actionDetailView(group: latest))
-            },
-            // Phase 0 (ref #221): gear button opens Settings view.
-            onOpenSettings: { [weak self] in
-                guard let self else { return }
-                self.navigate(to: self.settingsView())
-            }
-        ))
-    }
-
-    /// Navigation: Settings view (Phase 0, ref #221).
-    private func settingsView() -> AnyView {
-        savedNavState = .settings
-        return AnyView(SettingsView(
-            onBack: { [weak self] in
-                guard let self else { return }
-                self.navigate(to: self.mainView())
-            }
-        ))
-    }
-
-    /// Navigation level 2a: flat job list for a commit/PR group.
-    private func actionDetailView(group: ActionGroup) -> AnyView {
-        savedNavState = .actionDetail(group)
-        return AnyView(ActionDetailView(
-            group: group,
-            onBack: { [weak self] in
-                guard let self else { return }
-                self.navigate(to: self.mainView())
-            },
-            onSelectJob: { [weak self] job in
-                guard let self else { return }
-                DispatchQueue.global(qos: .userInitiated).async {
-                    let enriched = self.enrichStepsIfNeeded(job)
-                    DispatchQueue.main.async {
-                        guard self.popoverIsOpen else { return }
-                        self.navigate(to: self.detailViewFromAction(job: enriched, group: group))
-                    }
-                }
-            }
-        ))
-    }
-
-    /// Navigation level 3a: JobDetailView reached via an ActionGroup.
-    private func detailViewFromAction(job: ActiveJob, group: ActionGroup) -> AnyView {
-        savedNavState = .actionJobDetail(job, group)
-        return AnyView(JobDetailView(
-            job: job,
-            onBack: { [weak self] in
-                guard let self else { return }
-                self.navigate(to: self.actionDetailView(group: group))
-            },
-            onSelectStep: { [weak self] step in
-                guard let self else { return }
-                self.navigate(to: self.logViewFromAction(job: job, step: step, group: group))
-            }
-        ))
-    }
-
-    /// Navigation level 4a: StepLogView reached via an ActionGroup.
-    private func logViewFromAction(job: ActiveJob, step: JobStep, group: ActionGroup) -> AnyView {
-        savedNavState = .actionStepLog(job, step, group)
-        return AnyView(StepLogView(
-            job: job,
-            step: step,
-            onBack: { [weak self] in
-                guard let self else { return }
-                self.navigate(to: self.detailViewFromAction(job: job, group: group))
-            }
-        ))
-    }
-
-    /// Navigation level 2: step list for a job (Jobs path).
-    private func detailView(job: ActiveJob) -> AnyView {
-        savedNavState = .jobDetail(job)
-        return AnyView(JobDetailView(
-            job: job,
-            onBack: { [weak self] in
-                guard let self else { return }
-                self.navigate(to: self.mainView())
-            },
-            onSelectStep: { [weak self] step in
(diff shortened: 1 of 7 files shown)


eonist:
## Review

Overall this is a solid, well-structured implementation. The phased approach with CI verification between each phase is a good pattern, and the regression guard comment on `SettingsView` is appreciated.

A few substantive points:

**`signInWithGitHub()` — functional concern**
Launching Terminal and running `gh auth login` via `NSAppleScript` is fragile: it requires Terminal.app, requires `gh` CLI to be installed, and the `NSWorkspace.open` call immediately after is redundant since the AppleScript already opens Terminal. More importantly, this is unlikely to work for most users. Consider either linking to a PAT setup guide in the browser, or handling token input directly in the UI.

**`analyticsEnabled` — inconsistent persistence pattern**
The other two stores (`SettingsStore`, `NotificationPrefsStore`) use `@Published` with `didSet` in a dedicated `ObservableObject`. `analyticsEnabled` is a raw `@State` in the view with inline `UserDefaults` read/write. This is inconsistent and means analytics opt-out won't be observed across views. Worth folding into `SettingsStore` or a `LegalPrefsStore` to match the established pattern.

**`private func navigate(to:)` → `func navigate(to:)`**
All the view factory methods were intentionally `private`. They're now `internal` (no explicit access modifier), presumably to make them callable from the extension. This is fine — just confirm there's no path where these are called from outside `AppDelegate` unintentionally, since `settingsView()`, `mainView()`, etc. are now part of the module's public surface.

**Privacy/EULA URLs are placeholders**
Both `linkRow` calls in `legalSection` point to the repo root. Fine for now if tracked, but worth a follow-up issue before release so these don't ship as stubs.

eonist:
## Review: feat(#221) Settings panel — all phases (0–6)

Overall this is a clean, well-structured implementation. The phased approach with CI gates between each step is solid. A few things worth flagging:

---

**`signInWithGitHub()` — security and UX concern**

The current implementation spawns a Terminal window to run `gh auth login`, which is unexpected UX for a menu bar app and exposes the user to an AppleScript-controlled terminal. If `gh` is not installed, it silently does nothing. Consider opening the GitHub device-flow URL directly in the browser (`https://github.com/login/device`) or documenting that the `GITHUB_TOKEN` env var / keychain entry is the supported auth path — whichever matches the existing `githubToken()` implementation.

---

**`analyticsEnabled` not wired to anything**

The `legal.analyticsEnabled` UserDefaults key is read and written in `SettingsView`, but there's no evidence it's consumed anywhere else (no store, no analytics call site reads it). If analytics aren't actually implemented yet, the toggle will mislead users. Either wire it up, remove the toggle, or label it clearly as "coming soon" until it's functional.

---

**`private` removed from all view factory methods in `AppDelegate`**

All navigation methods (`mainView()`, `settingsView()`, `detailView()`, etc.) were changed from `private` to `internal`. The diff comment says it's because they moved into an extension — but extensions in the same file can access `private` members of the type. These should stay `private` unless there's a cross-module or test target reason to expose them.

---

**Privacy Policy and EULA links both point to the repo root**

Both `linkRow` calls use `https://github.com/eoncode/runner-bar` as the destination. If these are placeholders, that's fine — but it's worth a tracking issue to replace them before a public release.

eonist:
## Review — Phase 0–6 Settings (PR #240)

Overall this is solid. The navigation wiring is clean, the `SettingsStore` / `NotificationPrefsStore` split is sensible, and removing scopes/launch-at-login from `PopoverMainView` is exactly the right migration.

Three real issues worth fixing before merge:

---

### 🔴 1. `analytics` default is opt-out, not opt-in

`analyticsEnabled` defaults to `true` — meaning users share analytics without ever consenting. The phase plan spec says **opt-in**. Change the default:

```swift
// SettingsView.swift
@State private var analyticsEnabled: Bool = {
    guard UserDefaults.standard.object(forKey: "legal.analyticsEnabled") != nil else { return false } // ← was true
    return UserDefaults.standard.bool(forKey: "legal.analyticsEnabled")
}()
```

---

### 🔴 2. Privacy Policy and EULA URLs are placeholders

Both `linkRow` calls point to `https://github.com/eoncode/runner-bar`, which is just the repo root. These will ship to users as broken legal links. Either replace with real URLs or gate them behind `#if DEBUG` until the real pages exist.

---

### 🟡 3. `navigate` and view factory methods are now `internal` (were `private`)

The refactor moved `navigate(to:)`, `mainView()`, `settingsView()`, etc. from `private` to `internal` (implicit). This widens the access surface of `AppDelegate` beyond what's needed — nothing outside the file or extension calls these. Mark them `private` again, or at minimum add a comment explaining why the visibility was widened.

---

No other blockers. The `validatedView` `.settings` case restoring correctly, the `ServiceManagement` import moving to `SettingsView`, and the `onOpenSettings` callback wiring are all correct.

eonist:
@redoh

1. would be cool to have links to the files in the review. even line where the code is. 
2. also brand the review with octopus. So it's easy to distinguish etc?
3. since the review triggers on each commit to the pr. Maybe link the commit in the review? so humans and ai know what the review represents. Or maybe it stopped doing that. and it's just one review now on pr creation?

actually my bad. octopus hasn't reviewed it yet. Seems its rate limited:

Screenshot 2026-05-07 at 13 09 28

eonist:
## Follow-up review

All four points from the previous review are addressed cleanly. A few remaining observations:

**`signInWithGitHub()` is still in `SettingsView`, but the button only shows when `isAuthenticated == false`**
Opening `https://github.com/login/device` is fine as a stopgap, but the device-flow requires a `user_code` generated by a client ID — opening that URL cold just shows a blank "Enter code" page with no code to enter. A plain link to `https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens` or the app's own README setup guide would be less confusing for the user.

**`pollingInterval` change has no effect at runtime**
`SettingsStore` persists the new value, but nothing in this PR wires the updated `pollingInterval` to `RunnerStore`'s polling timer. If that wiring exists elsewhere already, ignore this — but if not, the setting will silently do nothing until it's connected.

**`#if DEBUG` guard on legal links is a sharp edge**
The Privacy Policy and EULA rows are now completely hidden in release builds. That means release users have no visible legal section at all (just the analytics toggle). This is probably acceptable short-term, but it's worth a follow-up issue to track the real URLs so these don't get forgotten — the `#if DEBUG` block will silently drop them from release without any compile-time reminder.

eonist:
## Follow-up review

All four issues from the previous review have been addressed cleanly:

- ✅ `signInWithGitHub()` now opens the GitHub device-flow URL in the browser — no more AppleScript/Terminal.
- ✅ `analyticsEnabled` is now backed by `LegalPrefsStore` using the same `@Published + didSet` pattern as the other stores, consistent and correctly persisted.
- ✅ The visibility note in `PopoverMainView` confirms no internal access was added; the `private extension` split is correctly scoped.
- ✅ Placeholder Legal links are gated behind `#if DEBUG` so they won't ship to users.

The `onScopeChanged` callback threading the reload back to `AppDelegate` is a good call — keeps the view layer clean and avoids the observable going stale between polls.

**One remaining flag worth knowing about:**

`pollingInterval` is saved to `UserDefaults` via `SettingsStore` but there's no observable wiring between `SettingsStore.pollingInterval` and the actual poll timer in `RunnerStore`. A user changing the interval in Settings won't take effect until app relaunch. This is likely pre-existing scope (the stepper existed before this PR wired up the store), so not a blocker here — but worth a follow-up issue if it isn't already tracked.

Otherwise this is good to merge. 👍

eonist:
## Review — follow-up (all previous feedback addressed ✅)

All three issues from the first review are resolved:

- **Analytics opt-in** — `LegalPrefsStore` correctly defaults `analyticsEnabled` to `false`. Dedicated store is consistent with the `NotificationPrefsStore` / `SettingsStore` pattern.
- **Placeholder legal links** — correctly gated behind `#if DEBUG`. Won't ship to users.
- **`signInWithGitHub`** — AppleScript removed, replaced with a clean `NSWorkspace.shared.open` to the device-flow URL. The auth hint in the Account section (`gh auth login` / `GH_TOKEN`) is the right UX for this distribution model.

The `onScopeChanged` callback and the immediate `observable.reload()` in `AppDelegate.settingsView()` are a good catch — prevents the regression where a newly added scope wouldn't appear until the next poll.

✅ Approved. Good to merge.

eonist:
## Follow-up review (round 3)

The `pollingInterval` wiring commit (`b999669`) is the only substantive new change. It modifies `RunnerStore.swift` to observe `SettingsStore.pollingInterval` and reschedule the timer immediately on change — the original concern is fully resolved.

The remaining two commits are lint fixes (line length, superfluous disable comment) with no logic changes.

Nothing blocking. ✅ Good to merge.

eonist:
## Follow-up review (round 3)

All three points from the previous review are addressed. The `RunnerStore` Combine wiring for `pollingInterval` is clean, the PAT docs link is the right destination, and the `#if DEBUG` guard on placeholder legal links is a reasonable short-term approach.

No blocking issues. This is ready to merge. ✅

eonist:
## Review — latest commits

Three changes since the last approval — all correct:

- **PAT docs URL** — right call. The device-flow URL (`/login/device`) requires a `user_code` the app never generates, so it would have opened a useless page. The PAT docs page is the correct destination for this distribution model.
- **`pollingInterval` wiring** — the Combine subscription on `SettingsStore.shared.$pollingInterval` with `dropFirst(1)` and an immediate `start()` call is exactly right. Timer restarts on the next runloop tick after the stepper changes, no extra machinery needed.
- **Lint fix** — trivial, correct.

✅ Still approved. Good to merge.

orchidfire:
## Review: PR #240 vs PR #244 for Issue #132

**PR #244 is the better choice** for closing issue #132.

### PR #240 Strengths:
- Clean architecture with dedicated stores
- Polling interval properly wired via Combine
- Tighter, more focused commits
- Cleaner SettingsView layout helpers

### Why PR #240 Falls Short:
- ❌ References wrong issue (#221 instead of #132)
- ❌ Missing About section with version info (required by #132)
- ❌ Less comprehensive coverage of the checklist

### PR #244 Wins Because:
✅ Correctly targets issue #132
✅ Implements ALL checklist items from #132 including About section
✅ Runner list preview in Settings provides useful overview
✅ Proper store architecture with persistence
✅ Polling interval wired and functional

The minor code style differences (section styling, file length) don't outweigh PR #244's completeness. The runner list duplication is acceptable trade-off for having quick status visibility in Settings.

Closing recommendation: Merge #244, close #240.

eonist:
reference only pr 244 was the cleaner implementation

Issue #680, opened 2026-05-22
ghIsRateLimited has no automatic reset — app stays throttled after rate limit expires


eonist:
## Problem

`ghIsRateLimited` is set to `true` on a 403/429 response but is never automatically cleared. GitHub rate limits reset after 60 minutes. If a rate limit is hit, the app will skip all API calls for the rest of the session (until restart), even after the limit has expired.

## Fix

Auto-reset `ghIsRateLimited` after the rate limit window. The simplest approach:

```swift
var ghIsRateLimited: Bool {
    get { _rateLimitLock.withLock { $0 } }
    set {
        _rateLimitLock.withLock { $0 = newValue }
        if newValue {
            // GitHub rate limits reset after 60 minutes.
            DispatchQueue.global().asyncAfter(deadline: .now() + 3600) {
                ghIsRateLimited = false
                log("ghIsRateLimited › auto-reset after 60 min")
            }
        }
    }
}
```

Alternatively, check the `X-RateLimit-Reset` response header (a Unix timestamp) and schedule the reset precisely. Also consider resetting `ghIsRateLimited` at the start of each poll cycle in `RunnerStore` so a manual refresh can recover early.

## Files
- `Sources/RunnerBar/GitHub/GitHubURLSessionTransport.swift`

Part of #654

Issue #678, opened 2026-05-22
Replace scope.contains("/") heuristic with typed Scope enum


eonist:
## Problem

Throughout the codebase, `scope.contains("/")` is the sole mechanism for distinguishing a repo scope (`owner/repo`) from an org scope (`orgname`). This appears 9+ times across:

- `GitHubCLITransport.swift` (`deleteRunnerByID`, `patchRunnerLabels`, `fetchRegistrationToken`, `fetchRemovalToken`)
- `GitHub.swift` (`fetchActiveJobs`, `fetchRunners`, `fetchStepLog`)
- `RunnerLifecycleService.swift` (`scopeFromGitHubUrl` output consumer)

This is a fragile string heuristic with no type safety. A scope string that is accidentally empty, malformed, or contains a `/` in an org name would silently produce wrong API endpoints.

## Fix

Introduce a typed `Scope` enum:

```swift
enum Scope {
    case repo(owner: String, name: String)
    case org(String)

    static func parse(_ string: String) -> Scope? {
        let parts = string.split(separator: "/", maxSplits: 1).map(String.init)
        if parts.count == 2 { return .repo(owner: parts[0], name: parts[1]) }
        if parts.count == 1, !parts[0].isEmpty { return .org(parts[0]) }
        return nil
    }

    var apiPrefix: String {
        switch self {
        case .repo(let owner, let name): return "repos/\(owner)/\(name)"
        case .org(let org): return "orgs/\(org)"
        }
    }
}
```

All endpoint-building becomes `"\(scope.apiPrefix)/actions/runners"` etc. The `contains("/")` checks are eliminated entirely.

## Files
- New: `Sources/RunnerBar/GitHub/Scope.swift`
- `Sources/RunnerBar/GitHub/GitHubCLITransport.swift`
- `Sources/RunnerBar/GitHub/GitHub.swift`
- `Sources/RunnerBar/Runner/RunnerLifecycleService.swift`

Part of #654

Issue #674, opened 2026-05-22
Migrate log() from fputs/stderr to os.Logger (unified logging)


eonist:
## Problem

`Logger.swift` implements logging via `fputs` to `stderr` with a manually-formatted ISO-8601 timestamp. macOS has had `os.Logger` (unified logging) since macOS 11, which is well below any realistic minimum deployment target for this app.

Current approach downsides:
- All log messages appear in stderr regardless of severity — no filtering
- Not visible in Console.app with subsystem/category filtering
- `debug`-level calls cost the same in release builds as in debug builds
- Manual timestamp formatting allocates on every call

## Fix

Replace the `fputs` implementation with `os.Logger`:

```swift
import os

private let logger = Logger(subsystem: "com.eoncode.runner-bar", category: "general")

func log(_ message: String, file: String = #file, line: Int = #line) {
    let filename = URL(fileURLWithPath: file).deletingPathExtension().lastPathComponent
    logger.debug("\(filename):\(line) \u2014 \(message)")
}
```

For production you may want `logger.info` or severity-aware overloads. In release builds, `.debug` calls are compiled out by the OS at zero cost.

**Note:** Remove the now-redundant `import os` from `Auth.swift` after this change (it currently imports `os` only for `OSAllocatedUnfairLock`, which does not require `os.Logger`).

## Files
- `Sources/RunnerBar/Utilities/Logger.swift`
- `Sources/RunnerBar/GitHub/Auth.swift` (remove duplicate `import os` if applicable)

Part of #654

Issue #673, opened 2026-05-22
Extract shared ProcessRunner — runScriptWithOutput duplicated 3 times


eonist:
## Problem

The same ~30-line pattern (launch a `Process`, attach pipes, stream output with `NSLock`, wait with a `DispatchWorkItem` timeout, drain tail) appears in three separate places:

1. `runGHProcess()` in `GitHubCLITransport.swift`
2. `runScriptWithOutput()` in `RunnerLifecycleService.swift`
3. `patchRunnerLabels()` in `GitHubCLITransport.swift` (inline duplicate)

`RunnerLifecycleService` also has a `// swiftlint:disable:next type_body_length` suppression that would be removable after this extraction.

## Fix

Create `Sources/RunnerBar/Services/ProcessRunner.swift` with a shared primitive:

```swift
struct ProcessRunner {
    struct Result {
        let data: Data?
        let exitCode: Int32
    }

    static func run(
        executableURL: URL,
        arguments: [String],
        stdin: Data? = nil,
        workingDirectory: URL? = nil,
        timeout: TimeInterval = 20
    ) -> Result
}
```

Both `runGHProcess` and `runScriptWithOutput` become thin wrappers. Fixes the process logic in one place for all call sites.

## Files
- `Sources/RunnerBar/GitHub/GitHubCLITransport.swift`
- `Sources/RunnerBar/Runner/RunnerLifecycleService.swift`
- New: `Sources/RunnerBar/Services/ProcessRunner.swift`

Part of #654

PR #693, opened 2026-05-22
fix(#655): match runner metrics by installPath, not slot index

### LINKED ISSUE #655: RunnerStore: metric assignment uses slot index instead of installPath
## Problem

In `RunnerStore.fetchAndEnrichRunners()`, metrics are assigned to runners by slot index (busy runners first, then idle):

```swift
for idx in busyRunners.indices {
    busyRunners[idx].metrics = idx < metrics.count ? metrics[idx] : nil
}
for idx in idleRunners.indices {
    let slotIdx = busyRunners.count + idx
    idleRunners[idx].metrics = slotIdx < metrics.count ? metrics[slotIdx] : nil
}
```

This is a fragile index-pairing assumption. If two runners match the same process, or the count differs between the runners array and the metrics array, metrics get silently misassigned.

`LocalRunnerStore.refresh()` already does this correctly by matching metrics via `installPath` using `metricsForRunner(installPath:)`.

## Fix

Mirror `LocalRunnerStore`'s `installPath`-keyed approach in `RunnerStore.fetchAndEnrichRunners()` — iterate `allRunners`, check `runner.installPath`, and call `metricsForRunner(installPath:)` per runner rather than relying on index alignment with `allWorkerMetrics()`.

## Files
- `Sources/RunnerBar/Runner/RunnerStore.swift` — `fetchAndEnrichRunners()`
- `Sources/RunnerBar/Runner/LocalRunnerStore.swift` — reference implementation

Part of #654

eonist:
## Summary

Fixes #655.

`fetchAndEnrichRunners()` previously assigned `RunnerMetrics` to runners by slot index — busy runners first, then idle — using `allWorkerMetrics()` which returns an unordered list of all `Runner.Worker` processes on the system. This meant if the number of processes didn't exactly match the number of runners, or two runners shared similar process names, metrics could be silently assigned to the wrong runner.

`LocalRunnerStore.refresh()` already does this correctly by calling `metricsForRunner(installPath:)` per runner, matching by the runner's install path on disk.

## Fix

In `fetchAndEnrichRunners()`:

1. **Build a `name → installPath` lookup** from `LocalRunnerStore.shared.runners` (snapshotted via a `DispatchQueue.main.sync` hop, since `LocalRunnerStore` is `@MainActor`).
2. **For each busy `Runner`**, look up its `installPath` by `runner.name` and call `metricsForRunner(installPath:)` directly.
3. **Idle runners** always receive `nil` — they have no active `Runner.Worker` process.

The `allWorkerMetrics()` call is removed entirely from this path.

## Changes

- `Sources/RunnerBar/Runner/RunnerStore.swift` — `fetchAndEnrichRunners()`

## Notes

- If a busy runner has no matching entry in `LocalRunnerStore` (e.g. a remote-only runner not installed on this machine), it gracefully receives `nil` metrics with a log line.
- The `DispatchQueue.main.sync` hop is safe here because `fetchAndEnrichRunners()` is always called from a background queue (`DispatchQueue.global(qos: .background)`) and `LocalRunnerStore` is annotated `@MainActor`.

___

## **CodeAnt-AI Description**
**Match runner metrics to the correct local install**

### What Changed
- Runner metrics are now tied to each runner’s local install path instead of being assigned by position in the list
- Busy runners now get their own metrics when a matching local install is found
- Idle runners now always show no metrics
- If a busy runner has no local install on the machine, it now stays metric-free instead of showing another runner’s data

### Impact
`✅ Correct runner metrics`
`✅ Fewer mislabeled status readings`
`✅ Clearer idle runner displays`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate-limiting detection and fallback behavior with better header parsing and automatic reset scheduling.
  * Enhanced OAuth sign-out flow with improved state notification.
  * Increased reliability of subprocess execution with better timeout and error handling.

* **Improvements**
  * Streamlined menu bar status updates and panel initialization.
  * Upgraded logging system to use Apple's unified logging for better diagnostics.
  * Cleaned up deprecated UI components and navigation code.

[Review Change Stack](https://app.coderabbit.ai/change-stack/eoncode/runner-bar/pull/693?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


eonist:
## Fresh-eyes review 👁️

### Overall impression

This PR is substantially larger than the title suggests. The stated fix (metrics by `installPath`) is small and well-reasoned, but the branch carries a significant refactoring payload across ~42 files, 1219 additions, and 878 deletions. The refactoring quality is high, but the risk surface is much wider than the PR description conveys.

---

### What's actually in here

The PR bundles five distinct threads of work:

1. **The named fix** — `fetchAndEnrichRunners()` now matches metrics by `installPath` via a `main.sync` snapshot of `LocalRunnerStore`. Correct and safe.
2. **`Scope` enum** — replaces fragile `scope.contains("/")` string checks with a typed enum and `apiPrefix`. Solid improvement.
3. **`ProcessRunner`** — extracts duplicated `Process`/`Pipe`/timeout boilerplate into a shared primitive. Good DRY move.
4. **`PollResultBuilder`** — extracts state-building logic from `RunnerStore` into a pure, testable struct with injected closures. Architecture is clean.
5. **Housekeeping** — `onChange` → `PassthroughSubject`, `RunnerStoreObservable` → `RunnerViewModel`, `onMutate` → `didMutate`, store file moves to `Preferences/`, dead code removed from `_deprecated/`, `Logger` → `os.Logger`, rate-limit auto-reset scheduling.

---

### 🔴 Bug risk: `DispatchQueue.main.sync` in `fetchAndEnrichRunners()`

The comment says this is safe "because `fetchAndEnrichRunners()` is always called from a background queue." That's a fragile guarantee. If any future caller — including a test — invokes it on the main thread, you'll deadlock. Consider snapshotting `LocalRunnerStore` state *before* entering the background work, or passing the snapshot in as a parameter.

---

### 🟡 Stale `onMutate` reference in `SettingsView.swift`

Line ~138 still reads:
```swift
ScopeStore.shared.onMutate = { [weak store] in store?.reload() }
```
But `onMutate` was replaced with `didMutate` (`PassthroughSubject`) in this same PR. This will be a **compile error**. Needs to be replaced with a proper `didMutate.sink { ... }.store(in: &cancellables)`.

---

### 🟡 Empty tombstone files should be deleted, not left with comments

`Runner/LegalPrefsStore.swift`, `Runner/NotificationPrefsStore.swift`, `Runner/SettingsStore.swift`, and `Runner/RunnerStoreObservable.swift` are left with just a one-line comment saying *"Delete this file from the repository."* These should be **actually deleted** in this PR — they'll confuse Swift's module system and pollute Xcode's file navigator.

---

### 🟡 `Scope.parse` accepts malformed input silently

The `ScopeParseTests` test documents that `"owner/repo/extra"` is parsed as `repo(owner: "owner", name: "repo/extra")`. A malformed scope like `"eoncode/runner-bar/actions"` would silently be treated as a valid repo named `runner-bar/actions`. Consider returning `nil` for strings with more than one slash.

---

### 🟠 `unsafeFlags` in `Package.swift`

```swift
swiftSettings: [.unsafeFlags(["-enable-testing"])]
```
`.unsafeFlags` prevents the package from being used as a dependency and can break certain build modes. The right approach is to enable testability via the build scheme or use `.enableTestability` in a supported way, not `unsafeFlags` in the manifest.

---

### 🟢 Minor: `scheduleRateLimitReset` double-fire potential

When a 403/429 is received, the new code in `urlSessionAPI` calls `scheduleRateLimitReset(resetAt:)` directly after doing `rateLimitLock.withLock { $0 = true }`. This correctly bypasses the `ghIsRateLimited` setter (which also calls `scheduleRateLimitReset`) — avoiding a double-schedule. That's intentional and correct, but it's subtle. A comment explaining why you bypass the setter here would prevent future regressions.

---

### What's done well ✅

- `Scope` enum refactor is clean and the test coverage is solid.
- `ProcessRunner` is well-structured with good edge-case handling (timeout race guard, stdin pipe, `mergeStderr`).
- `_ghBinaryPath` lazy-static optimization is correct and well-placed.
- Removing `PopoverHeightKey`, `PopoverView`, `RunnerConfigSheet`, `StatusIcon`, and `StatusPanel` is healthy dead-code removal.
- Switching to `os.Logger` with `.debug` level is the right call — zero-cost in release.
- The `onSignOut` split from `onCompletion` in `OAuthService` is a clean semantic separation.

---

### Bottom line

The `onMutate` → `didMutate` migration in `SettingsView` appears incomplete and would cause a compile failure. The tombstone files should be deleted. The `DispatchQueue.main.sync` and `unsafeFlags` points are worth fixing before merge. Everything else is solid work.

eonist:
## Fresh-eyes review

This is a **large, high-quality refactor** that bundles the targeted bug fix (#655) with significant structural cleanup. The core fix is correct and well-reasoned. Most of the surrounding changes are improvements. A few things worth discussing before merging.

---

### ✅ What's genuinely good

**The core fix (`fetchAndEnrichRunners`)** is solid. Replacing slot-index assignment with a `name → installPath` lookup and calling `metricsForRunner(installPath:)` per runner is exactly the right approach. The `DispatchQueue.main.sync` hop to snapshot `LocalRunnerStore` is safe and the reasoning is well-documented.

**`ProcessRunner`** is a well-extracted primitive. Consolidating the duplicated `Process`/`Pipe`/`readabilityHandler`/timeout pattern into one reusable type is a meaningful reduction in footprint and risk. The `isRunning` guard in the timeout handler is a nice defensive touch.

**`Scope` enum** is a clean, testable abstraction. Replacing scattered `scope.contains("/")` string checks throughout `GitHub.swift` and `GitHubCLITransport.swift` with a typed enum eliminates an entire class of silent bugs. Tests cover the important edge cases.

**`onChange` → `didUpdate` / `onMutate` → `didMutate`** migration from plain closures to `PassthroughSubject` is the right direction — removes retain-cycle risk and makes subscription lifetime explicit.

**`os.Logger` migration** is a net win. Zero-cost `.debug` calls in release builds vs. `fputs` to stderr on every log line is a meaningful improvement.

**`PollResultBuilder` extraction** makes state-building logic independently unit-testable. The thin wrapper pattern in `RunnerStoreState.swift` keeps existing call sites unchanged.

**Dead code removal** (`StatusIcon`, `RunnerConfigSheet`, `PopoverView`, `PopoverHeightKey`, `StatusPanel`, `scrnsht.png`) is unambiguous housekeeping.

---

**1. Potential double-subscription in `AppDelegate` (blocker to confirm)**

`AppDelegate+PanelSetup.swift` introduces a `setupCombineSubscriptions()` that wires `LocalRunnerStore.$runners`, `RunnerStore.shared.didUpdate`, and `ScopeStore.shared.didMutate`. But `AppDelegate.swift` still contains what looks like the old version of that same block (the `+199` lines in the diff). If both run during `applicationDidFinishLaunching`, subscriptions are wired twice — resulting in duplicate `reload()` and `updateStatusIcon()` on every poll. Needs confirmation or reconciliation.

**2. `SettingsView` still assigns `ScopeStore.shared.onMutate` (likely stale reference)**

```swift
ScopeStore.shared.onMutate = { [weak store] in store?.reload() }
```

`onMutate` was replaced by `didMutate` (a `PassthroughSubject`) in this PR. This line will either fail to compile or silently do nothing — meaning the settings scope list may stop refreshing after add/remove. Looks like a missed cleanup.

**3. `test_threeSegments_parsesFirstTwoAsRepo` — surprising behaviour, intent unclear**

`Scope.parse("owner/repo/extra")` yields `.repo(owner: "owner", name: "repo/extra")` because of `maxSplits: 1`. `"repo/extra"` is not a valid GitHub repo slug. If this input should be rejected, `parse` should validate that `name` contains no `/`. If intentionally permissive, the test comment should say so explicitly.

**4. Double reset scheduling for `ghIsRateLimited`**

The property observer on `ghIsRateLimited` calls `scheduleRateLimitReset(resetAt: nil)` (60-min fallback) when set to `true`. The URLSession path now bypasses the setter and calls `scheduleRateLimitReset` directly with the header timestamp — that's correct. But any caller that goes through the plain `ghIsRateLimited = true` setter (CLI paths, tests) will schedule a second, redundant reset DispatchWorkItem on top. Harmless but potentially noisy.

**5. Tombstone files should be deleted in this PR**

`Runner/LegalPrefsStore.swift`, `Runner/NotificationPrefsStore.swift`, `Runner/SettingsStore.swift`, and `Runner/RunnerStoreObservable.swift` are left as empty files with "delete this file" comments. Since this PR is the vehicle for the move, actually deleting them here would be cleaner than a follow-up ticket.

**6. `_ansiRegex` force-try at module level**

The `try!` is justified (static pattern, never fails), but a runtime crash if the pattern is ever accidentally edited is still a non-zero risk. A `static let` scoped inside a private type or an `enum` namespace would be marginally safer. Not a blocker.

---

### Minor / nits

- `RunnerViewModel` carries over the grammar typo from `RunnerStoreObservable`: `"If your an agent or human"` → should be `"you're"`.
- Doc renames (`github_permission_rational.md` → `GitHub-Permission-Rationale.md`) also fixes the original "rational" → "rationale" typo — nice catch.
- The `❌ NEVER call makeStatusIcon()` / "DO NOT REMOVE THIS COMMENT" pattern in `AppDelegate+StatusItem.swift` is pragmatic but fragile — a lint rule would be more durable than a comment that can itself be removed.

---

**Bottom line:** approvable once items #1 and #2 are confirmed/resolved — those could cause real observable bugs. Everything else is polish.

eonist:
## Code Review — PR #693 (fresh eyes)

This is a large, multi-scope refactor landed as a single PR. The overall direction is excellent — the codebase is measurably cleaner after this lands. Below is an honest assessment of what's solid and what deserves a second look before merge.

---

### ✅ What's well done

**`Scope` enum (`Scope.swift`)** — Replacing scattered `scope.contains("/")` ternaries with a proper typed enum is the right call. `apiPrefix` is a clean computed property and the `parse()` factory is safe. The test coverage (`ScopeParseTests`) is thorough, including edge cases.

**`ProcessRunner` extraction** — Moving the inline `Process`/`Pipe`/`NSLock` boilerplate out of both `GitHubCLITransport` and `RunnerLifecycleService` into a single shared primitive is a solid reduction of duplication. The `isRunning` guard before `.terminate()` in the timeout handler is a nice race-condition fix.

**`PollResultBuilder` extraction** — Making all builder methods `static` and injecting side-effects (fetch, backfill) as closures is the right design for testability. The thin wrapper extensions on `RunnerStore` that delegate to it keep existing callsites unchanged.

**Reactive notifications (`didUpdate` / `didMutate`)** — Replacing raw optional closure callbacks with `PassthroughSubject<Void, Never>` eliminates the retain-cycle risk at every callsite and integrates naturally with the existing Combine chain in `AppDelegate`.

**`AppDelegate` split** — `AppDelegate+PanelSetup.swift` and `AppDelegate+StatusItem.swift` are focused, well-commented extensions. The `setupPanel()` / `setupKVO()` / `setupCombineSubscriptions()` decomposition reads clearly.

**Rate-limit handling** — The `scheduleRateLimitReset(resetAt:)` function and the `ghIsRateLimited` bail-out in `ghAPI` / `ghAPIPaginated` are a meaningful correctness fix: a rate-limit hit on URLSession no longer triggers a redundant CLI fallback request on the same cycle.

**`os.Logger` migration** — Using `Logger(subsystem:category:)` instead of `fputs` to `stderr` is the correct approach for a native macOS app. The `.debug` level being zero-cost in release builds is an added bonus.

---

### ⚠️ Issues worth addressing before merge

**1. Stale `RunnerStore.shared.onChange` assignment in `AppDelegate+PanelSetup.swift`**

In `AppDelegate+PanelSetup.swift`, `setupCombineSubscriptions()` still contains:
```swift
RunnerStore.shared.onChange = { [weak self] in
    ...
    self.updateStatusIcon()
    self.observable.reload(localRunnerStore: LocalRunnerStore.shared)
}
```
But `RunnerStore.onChange` was **removed** in this same PR and replaced with `didUpdate`. This is a compile error. The main `AppDelegate.swift` was correctly migrated to `.sink { ... }` — `AppDelegate+PanelSetup.swift` was not. Once fixed, both subscriptions will also **double-fire** since both respond to `RunnerStore.shared`.

**Recommendation:** Remove the entire `onChange` block from `AppDelegate+PanelSetup.swift`. The Combine subscription in `AppDelegate.swift` already covers this path.

---

**2. `ScopeStore.shared.onMutate` still assigned in `SettingsView.swift`**

```swift
ScopeStore.shared.onMutate = { [weak store] in store?.reload() }
```
`ScopeStore.onMutate` was removed and replaced with `didMutate` (a `PassthroughSubject`). This is a compile error. `AppDelegate+PanelSetup.swift` already subscribes to `ScopeStore.shared.didMutate` via Combine — this `SettingsView` assignment needs to be migrated to a `.sink { }` stored in a cancellable bag, or removed if `AppDelegate` already handles the restart.

---

**3. `Scope.parse` accepts `"owner/repo/extra"` as a valid repo scope**

The test `test_threeSegments_parsesFirstTwoAsRepo` documents and accepts: `"owner/repo/extra"` → `.repo(owner: "owner", name: "repo/extra")`. A valid GitHub repo name cannot contain `/`, so this produces an invalid API path like `repos/owner/repo/extra/actions/runners`. Any user who types a three-segment scope string will get silent API failures.

**Recommendation:** After `split(separator: "/", maxSplits: 1)`, reject when `parts[1].contains("/")`, or use `maxSplits: 2` and return `nil` when `parts.count > 2`.

---

**4. Tombstone files should be deleted, not left as stubs**

Four files contain nothing but a redirect comment telling the reader to delete them:
- `Sources/RunnerBar/Runner/RunnerStoreObservable.swift`
- `Sources/RunnerBar/Runner/LegalPrefsStore.swift`
- `Sources/RunnerBar/Runner/NotificationPrefsStore.swift`
- `Sources/RunnerBar/Runner/SettingsStore.swift`

Delete them in this PR. Leaving empty files with "delete this file" comments is worse than deleting them now — they accumulate and no one ever does the cleanup.

---

### 🔎 Minor / nits

- **Spurious lint suppression** — `AggregateStatus.swift` has `// swiftlint:disable type_body_length` / `// swiftlint:enable type_body_length` wrapping a 25-line enum. This rule is not triggered here; the pair can be removed.
- **`ghBinaryPath()` wrapper is now a no-op** — `func ghBinaryPath() -> String? { _ghBinaryPath }` adds an indirection over the already-cached `private let _ghBinaryPath`. Either inline references to `_ghBinaryPath` or keep the wrapper but mark it `@inline(__always)`.
- **No thread assertion in `ProcessRunner.run`** — The doc comment says "Must be called from a background thread". A `assert(!Thread.isMainThread, "ProcessRunner.run must not be called on the main thread")` would catch misuse during development at zero release-build cost.
- **`buildGroupState` lost its early-exit on empty scopes** — The old `RunnerStore.buildGroupState` had an explicit guard + log when `ScopeStore.shared.scopes` was empty. The new `PollResultBuilder` path silently returns an empty result. The behaviour is the same, but the diagnostic log was useful — worth adding back inside the `fetchGroups` closure.

---

### Summary

Issues 1 and 2 are compile errors — the build will not succeed as-is. Issue 3 is a correctness bug producing invalid API paths. These three must be fixed before merge. Issue 4 is housekeeping that should be done here rather than deferred. The nits can be addressed here or as follow-ups.

eonist:
## PR Review — Fresh Eyes

This PR carries significantly more scope than the title suggests. The commits resolve 25 issues across correctness, architecture, performance, dead-code removal, tooling, and CI. Here's a structured breakdown.

---

### ✅ What's solid

**`ProcessRunner` extraction (#673, #657)**
Excellent. The duplicated `Process` setup/pipe/timeout/drain pattern is gone. The `guard task.isRunning` race guard before `terminate()` is exactly right. The `mergeStderr` flag is clean and backward-compatible.

**`Scope` enum (#678)**
The `scope.contains("/")` heuristic was a latent bug waiting to happen. The typed enum with `parse(_:)` and `apiPrefix` is the correct fix — it makes invalid scope strings fail explicitly at one boundary instead of silently at every call site.

**`PollResultBuilder` extraction (#663)**
Pure, static, closure-injected — independently testable without a `RunnerStore` instance. The `buildJobState` / `buildGroupState` signatures are well-designed.

**`PassthroughSubject` migrations (#658, #659)**
Replacing `onChange`/`onMutate` closures with `didUpdate`/`didMutate` subjects eliminates the retain-cycle risk at each call site and aligns with the existing Combine usage in the rest of the codebase.

**`os.Logger` migration (#674)**
Correct substitution. Zero-cost `.debug` in release builds, Console.app filtering, `log stream` compatibility — all noted in the commit and reflected in the implementation. Removal of `ISO8601DateFormatter` allocation per call is a nice side effect.

**Rate-limit handling (#672, #680)**
Skipping the CLI fallback when `ghIsRateLimited` is true is the right call — the old path would fan out a second request on the same poll cycle. The `X-RateLimit-Reset` auto-reset with a 5 s floor / 2 h ceiling clamp is appropriately conservative.

**Metrics-by-installPath fix (#655)**
The core fix. The `DispatchQueue.main.sync` hop from the background queue to read `LocalRunnerStore` is the correct approach given `@MainActor` isolation.

---

### ⚠️ Issues to address

**1. `RunnerStatusEnricher` — rewrite diverges significantly from the original**

The original was a `struct` with a paginated, batch-lookup approach: one API call per unique scope, building `byID` and `byName` dictionaries, then applying enrichment in a second pass. The new version is a `final class` that issues one `URLSession` request *per runner* via a semaphore-blocked `dataTask`. For a fleet of 20+ runners this goes from 1–2 API calls to 20+ concurrent semaphore-blocked calls.

The `performRunnerLookup` also bypasses the shared `ghAPI` / `urlSessionAPI` plumbing entirely — it builds raw `URLRequest` objects with `githubToken()` directly, which means it does **not** participate in the `ghIsRateLimited` guard, does not fall back to the gh CLI, and does not share the rate-limit state. This is a regression from the previous design.

Suggestion: revert the enricher to the batch-lookup approach (or keep the per-runner loop but reuse `ghAPI` so the rate-limit path is honoured).

**2. `AppDelegate+PanelSetup.swift` — stale `onChange` reference**

```swift
// AppDelegate+PanelSetup.swift, line ~67
RunnerStore.shared.onChange = { [weak self] in   // ← should be didUpdate subscription
```

The `setupCombineSubscriptions()` method still wires `RunnerStore.shared.onChange` (the old closure pattern) instead of subscribing to `didUpdate`. The correct wiring is in `AppDelegate.swift` (line ~199), but `AppDelegate+PanelSetup.swift` re-introduces the old pattern. If `setupPanel()` is called, both paths fire — or if `AppDelegate.swift` is refactored later to delegate to `setupCombineSubscriptions()`, the `onChange` fallback creates confusion. One of these needs to be the canonical source.

**3. `Package.swift` — `.unsafeFlags` in test target**

```swift
swiftSettings: [
    .unsafeFlags(["-Xfrontend", "-disable-access-control"])
]
```

`unsafeFlags` breaks library evolution and prevents the package from being used as a dependency. More practically, it prevents `swift build` in release mode from being used alongside the test target in some toolchain versions. The recommended approach is to make the tested symbols `internal` (already the default) and co-locate test files or use `@testable import`. The access-control bypass is a workaround, not a solution.

**4. `AggregateStatus.swift` — spurious suppression**

```swift
// swiftlint:disable type_body_length
```
The file is 25 lines. The suppression was copied from the source file and is unnecessary here.

**5. Stub files left in the repo**

`Runner/LegalPrefsStore.swift`, `Runner/NotificationPrefsStore.swift`, `Runner/SettingsStore.swift`, and `Runner/RunnerStoreObservable.swift` are all left as redirect stubs with comments like "Delete this file from the repository." These should be deleted in this PR — leaving them in means they'll quietly accumulate unless someone follows up. The commits that handle the `_deprecated/` folder deletions show the pattern already works fine.

---

### Minor

- `ActiveJob.swift` loses several doc comments (`isLocalRunner`, `conclusionIcon`, `elapsed`, `JobPayload`, `StepPayload`) — likely collateral from a reformat pass. Worth restoring.
- `SettingsView` drops a few load-bearing state comments (the four-state OAuth comment block, the `isOAuthAuthenticated`/`isCLIAuthenticated` mutual-exclusivity note). These were useful for future contributors.
- The `onSelect: { }` empty closure in `PopoverMainView` for `ActionRowView` is now truly dead — if `onSelect` has no callers, the parameter should be removed from `ActionRowView` rather than wired to an empty closure.

---

### Summary

The architecture improvements here are substantial and correct — `ProcessRunner`, `Scope`, `PollResultBuilder`, the Combine migrations, and the `os.Logger` switch are all clear wins. The main blocker is the `RunnerStatusEnricher` rewrite, which trades a batched approach for a per-runner semaphore loop that bypasses the shared rate-limit guard. The stale `onChange` in `AppDelegate+PanelSetup` and the `.unsafeFlags` in `Package.swift` are smaller but concrete issues.

Recommend: address the enricher regression and the stale closure, delete the stub files, and this is ready to merge.

eonist:
## PR Review — Fresh Eyes 👀

This is a well-scoped cleanup PR that does a lot of right things at once. Here's a holistic review:

---

### ✅ What's working well

**Scope type safety (`Scope.swift`)**
Introducing a proper `enum Scope { case repo, case org }` with a `parse(_:)` factory is a clear improvement over the previous `scope.contains("/")` string-sniffing pattern. All call sites (`fetchRunners`, `fetchActiveJobs`, `deleteRunnerByID`, `patchRunnerLabels`, `fetchRegistrationToken`, `fetchRemovalToken`, `fetchStepLog`) now benefit from typed dispatch. This eliminates a class of silent bugs where a malformed scope string could silently hit the wrong API path.

**`ProcessRunner` extraction**
Both `runGHProcess` in `GitHubCLITransport.swift` and `runScriptWithOutput` in `RunnerLifecycleService.swift` previously duplicated the same Process/Pipe/timeout/readabilityHandler boilerplate. Delegating to a shared `ProcessRunner.run(...)` removes ~80 lines of duplicated, error-prone subprocess management. The new `stdin: Data?` parameter on `runGHProcess` also cleanly unblocks `patchRunnerLabels` from its old hand-rolled `Process` setup.

**Combine-based notifications replacing callbacks**
`RunnerStore.onChange` → `didUpdate: PassthroughSubject` and `ScopeStore.onMutate` → `didMutate: PassthroughSubject` are the right move. The old optional closure pattern was a retain-cycle risk and had no clear ownership semantics. Callers now hold `AnyCancellable` and lifetime is explicit.

**Rate-limit auto-reset (`GitHubURLSessionTransport.swift`)**
The new `scheduleRateLimitReset(resetAt:)` using the `X-RateLimit-Reset` header is well-implemented. Clamping to `[5, 7200]` seconds is a sensible guard. The fix to bail out of the CLI fallback when `ghIsRateLimited` is `true` is correct and prevents a double-request on the same rate-limited window.

**Metrics assignment by `installPath` (the titular fix)**
The previous slot-index approach in `fetchAndEnrichRunners()` was inherently fragile. Matching by `installPath` via `LocalRunnerStore` is the same strategy already used in `LocalRunnerStore.refresh()` — good consistency. The `DispatchQueue.main.sync` hop to snapshot `@MainActor` state from a background queue is the correct pattern here.

**`PollResultBuilder` extraction**
Moving the poll state logic to a static struct with injected closures makes it unit-testable in isolation. The test target addition (`RunnerBarTests`) in `Package.swift` with `-disable-access-control` + a CI workflow (`swift-test.yml`) shows this was intentional.

**File reorganisation (`Preferences/` folder)**
`LegalPrefsStore`, `NotificationPrefsStore`, and `SettingsStore` being moved to a `Preferences/` group is a clean structural improvement. The stub files left behind with a `// This file has been moved` comment are appropriate for now (though they should be deleted in a follow-up).

---

### ⚠️ Things worth noting

**`RunnerStatusEnricher` changed from `struct` to `final class`**
The old `struct` was fine — it was stateless and had no stored mutable properties. The change to `final class` with a private `init()` singleton introduces shared mutable state for what is essentially a stateless utility. The new `performRunnerLookup` uses `URLSession` directly with a `DispatchSemaphore` blocking wait, which works but bypasses the existing `ghAPI` / `urlSessionAPI` layer (including its `ghIsRateLimited` guard and rate-limit reset logic). Worth aligning with the rest of the transport layer.

**`OAuthService.onSignOut` callback**
The new `var onSignOut: (() -> Void)?` is an optional closure on an `@MainActor` class — the same pattern that was just replaced on `RunnerStore` and `ScopeStore` with Combine subjects. Consider `let didSignOut = PassthroughSubject<Void, Never>()` for consistency, especially since `OAuthService` is headed in that direction already.

**`AggregateStatus.swift` has a stray `// swiftlint:disable type_body_length`**
The enum is 25 lines — this disable is completely unnecessary and should be removed.

**Stub files in `Runner/`**
`LegalPrefsStore.swift`, `NotificationPrefsStore.swift`, `SettingsStore.swift`, and `RunnerStoreObservable.swift` are now stub/redirect files. These should be deleted from the repo in a follow-up (or ideally in this PR) to avoid future confusion.

---

### Summary

This is a solid refactor PR. The core correctness fixes (metrics by `installPath`, rate-limit double-request, `Scope` type safety) are all correct. The architectural improvements (Combine subjects, `ProcessRunner`, `PollResultBuilder`) are well-executed. The two main things worth addressing going forward are the `RunnerStatusEnricher` bypassing the shared transport layer, and the `onSignOut` callback being inconsistent with the new Combine pattern. Neither is blocking for merge.

**Overall: ✅ Looks good to merge.**

eonist:
## PR Review — `fix(#655): match runner metrics by installPath, not slot index`

Overall this is a solid, well-scoped PR. The core bug fix is correct and the surrounding refactors are improvements. Some notes below.

---

### ✅ Core Fix (`RunnerStore.fetchAndEnrichRunners`)

The old slot-index approach was clearly fragile — assigning metrics positionally from an unordered `allWorkerMetrics()` list was always a race waiting to happen. The new `installPath`-keyed lookup is the right approach and matches how `LocalRunnerStore.refresh()` already works. The `DispatchQueue.main.sync` hop notes in the PR description are accurate — the `buildInstallPathMap` snapshot is taken on the main actor before the `Task.detached` body runs, which is correct.

One minor observation: `buildInstallPathMap` builds a compound key `"\(scope)/\(runnerName)"` rather than keying purely by `runnerName`. This means a runner named `"my-runner"` in scope `"org/repo-a"` would not match a lookup from scope `"org/repo-b"`, even if the same physical machine hosts both. That seems intentional (and correct), but worth keeping in mind if runners ever serve multiple scopes.

---

### ✅ `Scope` enum (`Scope.swift`)

Clean extraction. Replaces the `scope.contains("/")` string checks scattered across the codebase with a typed enum and `apiPrefix`. No issues.

---

### ✅ `ProcessRunner` (`Services/ProcessRunner.swift`)

Good consolidation — the duplicated `NSLock`/`Pipe`/timeout boilerplate across `runGHProcess` and `runScriptWithOutput` was technical debt worth paying off. The `isRunning` guard in the timeout `DispatchWorkItem` is a nice defensive addition.

---

### ✅ `AppDelegate` refactor

Extracting `setupPanel()` / `setupStatusItem()` into extensions tidies up a very large `applicationDidFinishLaunching`. The `@objc func togglePanel()` visibility change (from `private` to internal) is necessary and the comment explains why.

---

### ✅ Combine migration (`onChange` → `didUpdate`, `onMutate` → `didMutate`, `onCompletion` → `didSignOut`)

Replacing optional closure callbacks with `PassthroughSubject` is the right direction — avoids retain-cycle risks and aligns with the existing Combine usage in the app. The `ScopeStore.didMutate` and `OAuthService.didSignOut` subscriptions in `SettingsView` are properly held in `@State` cancellables so they live for the view's lifetime.

---

### ✅ `RunnerViewModel` rename

Purely cosmetic / naming improvement over `RunnerStoreObservable`. The empty `RunnerStoreObservable.swift` left as a placeholder is a reasonable bridge — just make sure it gets deleted in a follow-up before it confuses anyone.

---

### ✅ `PollResultBuilder` extraction

Moving the pure state-building logic out of `RunnerStore` into a static struct with injected closures makes it independently testable. The new unit tests in the CI workflow are a welcome addition.

---

### ✅ Logger upgrade (`os.Logger`)

Switching from `fputs` to `os.Logger` is strictly better — zero cost at release, subsystem/category filtering in Console.app, no timestamp allocation on every call.

---

### ⚠️ Minor observations

1. **`RunnerStatusEnricher` is now `@unchecked Sendable`** — this was changed from a `struct` to a `final class`. The `@unchecked Sendable` conformance suppresses Swift 6 concurrency warnings but means the compiler won't catch future mutations that break thread-safety. The current implementation looks safe (no mutable state), but it's worth a comment explaining why `@unchecked` is safe here rather than making it a struct or adding internal synchronisation.

2. **`_ghBinaryPath` lazy constant** — caching the binary path once at load time is fine for normal use, but if `gh` is installed after the app launches (e.g. via Homebrew in the same session) it won't be picked up until next launch. The previous behaviour was to re-check on every call. This is a very minor regression edge case, but worth noting.

3. **`scheduleRateLimitReset` can fire multiple times** — if `ghIsRateLimited = true` is set from multiple concurrent `urlSessionAPI` calls (paginated requests), `scheduleRateLimitReset` will schedule multiple reset timers on `DispatchQueue.global()`. Only the last one to fire would leave `ghIsRateLimited = false`, but earlier timers resetting it prematurely could cause a brief window where rate-limited requests sneak through again. A `DispatchQueue`-serialised or token-based approach would be safer here.

4. **`RunnerStoreObservable.swift` tombstone** — the file comment says "Delete this file from the repository" but it still exists in this PR. Fine as a follow-up, just make sure it doesn't linger.

5. **Typo in `RunnerViewModel.swift` line 44**: `"If your an agent or human"` → should be `"If you're an agent or human"` (copy-paste from the old file, so not introduced here, but it could be fixed in passing).

---

### Summary

The core `#655` fix is correct and the surrounding refactors are all improvements. The `Scope` enum, `ProcessRunner`, `PollResultBuilder`, and Combine migrations are all positive changes. The minor points above are non-blocking. Happy to approve once the `@unchecked Sendable` concern on `RunnerStatusEnricher` is addressed or documented.

eonist:
## 🔍 PR Review — `fix(#655): match runner metrics by installPath, not slot index`

This is a large, well-structured PR. The core bug fix is sound and the surrounding refactoring work is high quality. Here are my observations:

---

### ✅ What's good

- **Core fix (`fetchAndEnrichRunners`)** — replacing the fragile slot-index approach with `metricsForRunner(installPath:)` per busy runner is exactly the right fix. The `buildInstallPathMap` snapshot taken on the main actor before the `Task.detached` is a clean pattern.
- **`Scope` enum** — extracting `scope.contains("/")` into a typed enum with `apiPrefix` eliminates a class of silent bugs from raw string comparisons. Consistent application across all call sites.
- **`ProcessRunner`** — consolidating the duplicated `Process`/`Pipe`/`readabilityHandler`/timeout boilerplate into a single shared primitive is a significant maintainability win. The race-guard on the timeout (`guard task.isRunning`) is a nice touch.
- **`PollResultBuilder`** — extracting pure state-building logic into a static, side-effect-free struct makes it independently unit-testable. The injected-closure design is clean.
- **Combine migration** (`onChange` → `didUpdate`, `onMutate` → `didMutate`, `onCompletion` → `didSignOut`) — consistent, no retain-cycle risk, and easier to compose.
- **`os.Logger`** — switching from `fputs` to the unified logging system (with `.debug` level being zero-cost in release) is the right direction.
- **`kSecUseDataProtectionKeychain`** — the comment explaining the CSSM crash scenario is genuinely useful context to keep.
- **Rate-limit handling** — scheduling an auto-reset using `X-RateLimit-Reset` and clamping to 5s–7200s is solid. Skipping the CLI fallback when `ghIsRateLimited` is true avoids double-hitting a rate-limited API.

---

### ⚠️ Issues / Suggestions

#### 1. `buildInstallPathMap` key collision risk
```swift
map["\(scope)/\(localRunner.runnerName)"] = path
```
If a runner name appears in multiple scopes, the last scope wins silently. This is probably acceptable in practice, but worth a `log()` on collision for debuggability.

#### 2. `RunnerStore` is `@MainActor` but `fetch()` accesses `LocalRunnerStore.shared.runners` without a hop
```swift
let installPathByName = buildInstallPathMap(
    scopes: scopesSnapshot,
    localRunners: LocalRunnerStore.shared.runners  // @MainActor access — fine if fetch() is on main
)
```
`fetch()` itself is only called from `scheduleTimer`'s `DispatchQueue.main.async` block and from `start()` (also main). That's consistent, but since `RunnerStore` is now `@MainActor`, it would be worth adding a `// Called on main actor` comment to `fetch()` so future contributors don't accidentally call it from a background thread before the `Task.detached`.

#### 3. `RunnerStoreObservable.swift` is a tombstone file — should be deleted
The file now contains only:
```swift
// This file has been renamed to RunnerViewModel.swift as part of #664.
// It is intentionally left empty. Delete this file from the repository.
```
It adds noise to the file tree and could cause confusion. The comment says to delete it — this PR is the right time to do that.

#### 4. `_ghBinaryPath` lazy init is good but `ghBinaryPath()` wrapper is now redundant
```swift
private let _ghBinaryPath: String? = { ... }()
func ghBinaryPath() -> String? { _ghBinaryPath }
```
If `ghBinaryPath()` is called from outside the file, the wrapper is needed. But if it's only internal to `GitHubCLITransport.swift`, consider either making `_ghBinaryPath` internal or just inlining the lookup. Minor point — not blocking.

#### 5. `RunnerStatusEnricher` — pagination dropped
The old `buildAPILookup` paginated through all runner pages (`per_page=100&page=N`). The new `fetchRunnersForScope` does a single unpaginated `GET repos/{owner}/{repo}/actions/runners` call. For orgs or repos with more than 30 runners (the default GitHub API page size), this will silently miss runners. Consider adding `?per_page=100` to the endpoint.

#### 6. `freezeVanishedGroups` in `PollResultBuilder` iterates `snapPrev` by SHA but matches live groups by `id`
```swift
let liveIDs = Set(liveGroups.map { $0.id })
for (sha, group) in snapPrev where !liveIDs.contains(sha) { ... }
```
`snapPrev` is keyed by `sha` (its dictionary key), but `liveIDs` is built from `$0.id`. If `id != sha`, groups could be incorrectly frozen. Verify that `ActionGroup.id == headSha` is an invariant — if not, this is a latent bug.

#### 7. Typo in preserved comment
```swift
/// If your an agent or human, DO NOT REMOVE THIS COMMENT...
```
Should be `"you're an agent"`. (Present in both `RunnerViewModel.swift` and the old `RunnerStoreObservable.swift` — fix in the live file.)

---

### 🔒 Security note
`Secrets.swift` changes update comment wording around the committed `client_secret`. The extended rationale is clear and the reasoning is valid for an open-source native macOS OAuth app. No new secrets appear to be introduced.

---

### Summary

| Area | Verdict |
|---|---|
| Core bug fix | ✅ Correct and well-motivated |
| `Scope` enum refactor | ✅ Clean |
| `ProcessRunner` extraction | ✅ Strong improvement |
| `PollResultBuilder` extraction | ✅ Testable design |
| Combine migration | ✅ Consistent |
| Pagination in `RunnerStatusEnricher` | ⚠️ Potentially regressed — add `?per_page=100` |
| `RunnerStoreObservable.swift` tombstone | ⚠️ Delete this file |
| `freezeVanishedGroups` SHA vs ID key | ⚠️ Verify `id == headSha` invariant |

Overall: **approve with minor fixes** — the pagination regression in `RunnerStatusEnricher` is the most important item to address before merging if you have orgs/repos with many runners.

eonist:
## Code Review — PR #693: `fix(#655): match runner metrics by installPath, not slot index`

Great PR overall. This is a meaningful correctness fix, and the surrounding refactoring is well-structured and clearly motivated. Here are my observations:

---

### ✅ What's done well

- **Core bug fix is correct and clear.** `fetchAndEnrichRunners` previously assigned metrics by slot index (fragile), now matches by `installPath` via a pre-built name→installPath map. The fix is minimal and targeted, and the fallback path (`nil` metrics when no local install is found) is safe and logged.
- **`ProcessRunner` extraction is a solid improvement.** Consolidating duplicated `Process`/`Pipe` boilerplate into a single `ProcessRunner.run(...)` primitive reduces ~60 lines of repeated code in two files. The `mergeStderr`, `stdin`, and `workingDirectory` parameters cover all real callers.
- **`Scope` enum eliminates string-sniffing.** Replacing `scope.contains("/")` conditionals with a typed `Scope.parse()` + `apiPrefix` is the right abstraction. All five call sites (`fetchRunners`, `fetchActiveJobs`, `fetchStepLog`, `deleteRunnerByID`, `patchRunnerLabels`) are consistently updated.
- **`didUpdate` / `didMutate` / `didSignOut` Combine subjects** are a clean replacement for the old `onChange`/`onMutate`/`onCompletion` closure pattern. No retain-cycle risk at call sites, and the reactive wiring in `AppDelegate+PanelSetup` reads clearly.
- **`PollResultBuilder` extraction** makes the polling logic independently testable — good for future unit test coverage.
- **`os.Logger` migration** is a nice improvement: zero-cost debug logs in release builds, Console.app filtering, and cleaner call sites.
- **`kSecUseDataProtectionKeychain: true`** is a good defensive fix. The comment explaining the CSSMERR crash scenario is helpful context.

---

### ⚠️ Issues / things to check

**1. `buildInstallPathMap` uses scope-prefixed keys but `fetchAndEnrichRunners` scopes may not match `gitHubUrl` scope strings**

```swift
// RunnerStore.swift — buildInstallPathMap
map["\(scope)/\(localRunner.runnerName)"] = path
// Key: e.g. "owner/repo/MyRunner"

// fetchAndEnrichRunners
let key = "\(scope)/\(runner.name)"
// Key matches: also "owner/repo/MyRunner"
```

This works correctly *if* `ScopeStore.activeScopes` strings exactly match how `LocalRunnerStore` runners were registered. However, if a runner is registered under an org scope (`"myorg"`) but `ScopeStore` contains repo scopes (`"myorg/myrepo"`), the key lookup will silently fail for every busy runner and they'll all get `metrics = nil`. Worth adding a log or assertion that at least one key matched when runners are present.

**2. `RunnerStoreObservable.swift` is left as an empty file with a redirect comment**

The file body just says "renamed to RunnerViewModel.swift". An empty `.swift` file with only a comment compiles fine, but it should ideally be deleted from the repository to avoid confusion. The PR description mentions it as an intentional tombstone — consider adding a TODO comment with the PR/issue number to track cleanup, or deleting the file in this PR.

**3. `scheduleRateLimitReset` can fire multiple overlapping timers**

When `ghIsRateLimited = true` is set multiple times before the first reset fires (e.g. paginated requests all returning 403), `scheduleRateLimitReset(resetAt:)` is called each time, scheduling multiple independent `DispatchQueue.asyncAfter` blocks. They all eventually reset the lock to `false` but the last-scheduled one determines the effective delay. This is probably harmless in practice (the first reset just fires early), but the comment says "auto-reset is scheduled by `scheduleRateLimitReset(resetAt:)`" — singular — which implies one timer at a time. A simple `DispatchWorkItem` cancel-and-replace guard would make the intent explicit.

**4. `RunnerStatusEnricher` is now `final class` + `@unchecked Sendable` but uses no locks**

```swift
final class RunnerStatusEnricher: @unchecked Sendable {
    static let shared = RunnerStatusEnricher()
    private init() {}
    func enrich(runners: [RunnerModel]) -> [RunnerModel] { ... }
}
```

`enrich()` is stateless (no mutable instance state), so `@unchecked Sendable` is technically safe. But the class has no stored mutable state at all, so it could simply be a `struct` or a `enum` with static methods (like `PollResultBuilder`). The `@unchecked Sendable` marker is a yellow flag for reviewers — worth a brief comment like `// Stateless — @unchecked Sendable is safe; no mutable instance state.`

**5. Typo in `RunnerViewModel.swift` (carried over from `RunnerStoreObservable.swift`)**

```swift
/// If your an agent or human, DO NOT REMOVE THIS COMMENT...
//      ^^^^ should be "you're"
```

Minor, but since this comment is explicitly marked as load-bearing and permanent, worth fixing.

**6. `_ansiRegex` uses `try!` with a comment justifying it**

```swift
// swiftlint:disable:next force_try
private let _ansiRegex = try! NSRegularExpression(pattern: "...")
```

The justification ("compiled once at load time, pattern is a literal and never fails") is sound — `NSRegularExpression` with a hardcoded literal pattern will only fail if the pattern is malformed, which is impossible here. This is fine; the SwiftLint suppression and comment are appropriate.

---

### 💡 Minor suggestions

- `AppDelegate+StatusItem.swift` makes `togglePanel()` `@objc func` (previously `@objc private func`) to allow `#selector` cross-file reference. The comment explains this clearly — good. Alternatively, `@objc` on a `private` method in the *same* file as the `#selector` reference works fine; since the reference is in `AppDelegate+StatusItem.swift` (separate file), making it `internal` is the correct solution.
- The `_ghBinaryPath` lazy-initialised constant is a nice optimisation and clearly documented.
- Docs renames (`github_permission_rational.md` → `GitHub-Permission-Rationale.md`, note: "rational" → "rationale" — the rename also fixes a long-standing typo 👍) and `_deprecated/` file deletions are clean housekeeping.

---

### Summary

The core fix is correct and safe. The surrounding refactoring (Scope enum, ProcessRunner, PollResultBuilder, Combine subjects, os.Logger) is well-motivated and implemented cleanly. The items above are mostly low-risk observations — the most actionable one is **#1** (silent metrics miss when scope keys don't align), which could surface as a regression in org-scoped environments. 

Overall: **ready to merge with minor follow-up** on the key-alignment logging and the `RunnerStoreObservable.swift` tombstone file.

eonist:
## 🔍 PR Review — `fix(#655): match runner metrics by installPath, not slot index`

**Overall assessment: ✅ Strong PR — approve with minor notes**

This is a well-structured, clearly-motivated fix. The root cause (slot-index metric assignment) was genuinely fragile, and the approach of matching by `installPath` via a name-keyed lookup is significantly more correct. The refactoring work that came along for the ride is largely positive.

---

### ✅ What's done well

- **Core fix is correct.** `fetchAndEnrichRunners` now resolves metrics per runner via `installPath` instead of positional slot index. The two-tier lookup (`fullKey` → `byName` fallback) is a smart resilience measure for scope string mismatches between ScopeStore and fetch-time scope keys.
- **`ProcessRunner` extraction.** The new `ProcessRunner` enum cleanly consolidates the duplicated `Process`/`Pipe`/`NSLock`/timeout boilerplate that previously existed in both `GitHubCLITransport` and `RunnerLifecycleService`. This is a clear net improvement.
- **`Scope` enum.** Replacing `scope.contains("/")` string checks throughout `GitHub.swift` and `GitHubCLITransport.swift` with a typed `Scope` enum is much safer and more readable. The `parse` + `apiPrefix` pattern is clean.
- **Rate-limit improvements.** The `RateLimitState` struct + `scheduleRateLimitReset(resetAt:)` with cancel-and-replace `DispatchWorkItem` is a solid improvement over the previous bare `Bool`. The clamped delay (5s…7200s) and the CLI-fallback short-circuit (`data != nil || ghIsRateLimited`) both address real edge cases.
- **`PollResultBuilder` extraction.** Moving the pure state-building logic out of `RunnerStore` into a static struct makes it independently testable. The thin-wrapper delegation pattern in `RunnerStoreState.swift` keeps existing call sites clean.
- **`@MainActor` on `RunnerStore`.** The migration from `DispatchQueue.global(qos: .background).async` to `Task.detached(priority: .background)` + `await MainActor.run` is the correct Swift Concurrency pattern here and resolves the `dispatchPrecondition(.notOnQueue(.main))` trap risk.
- **`kSecUseDataProtectionKeychain: true`** in `Keychain.swift` — good defensive fix for the legacy CSSM crash.
- **`os.Logger` migration** in `Logger.swift` — zero-cost in release builds, Console.app filterable. Correct call.
- **`PassthroughSubject` migrations** (`onChange` → `didUpdate`, `onMutate` → `didMutate`, `onCompletion(false)` → `didSignOut`) are the right direction for eliminating retain-cycle-prone optional closures.
- **Dead code removal** — `_deprecated/` files, `PopoverHeightKey`, `StatusIcon`, `RunnerConfigSheet` etc. are appropriate cleanup.

---

### ⚠️ Minor concerns / suggestions

**1. `buildInstallPathMap` — scope × runner cartesian product**

```swift
for scope in scopes {
    byFullKey["\(scope)/\(localRunner.runnerName)"] = path
}
```

This builds a cross-product of all active scopes × all local runners. If there are, say, 5 scopes and 20 runners you get 100 entries, but most will never match because a runner is registered to exactly one scope. It works correctly, but the comment in `fetchAndEnrichRunners` that logs a warning when `byFullKey` is empty suggests this was a point of confusion during development. Consider documenting why the cross-product is intentional (or narrowing the lookup by matching runner's `gitHubUrl` scope to the active scopes list).

**2. `RunnerStatusEnricher` — key collision risk in `nameToAPI`**

```swift
var nameToAPI: [String: [String: Any]] = [:]
for scopeURL in scopeToRunnerIndices.keys {
    let fetched = fetchRunnersForScope(scopeURL)
    for apiRunner in fetched {
        if let name = apiRunner["name"] as? String {
            nameToAPI[name] = apiRunner  // last-writer-wins
        }
    }
}
```

If two scopes (e.g. `org/repo-A` and `org/repo-B`) happen to have a runner with the same name, the second fetch silently overwrites the first. The old code keyed by `"scope/name"` which was collision-safe. This is probably fine in practice (runner names are typically machine hostnames), but worth a log line when a collision is detected, or keying by `"scopeURL/name"` and matching in the second pass.

**3. `ghIsRateLimited = false` inside `Task.detached`**

```swift
Task.detached(priority: .background) { [weak self] in
    guard let self else { return }
    ghIsRateLimited = false
    ...
}
```

`ghIsRateLimited` is a global computed property backed by `OSAllocatedUnfairLock`. Resetting it unconditionally at the start of every poll cycle means a rate-limit set mid-poll (e.g. from a concurrent paginated request) could be cleared before the auto-reset timer fires. Given `scheduleRateLimitReset` already handles the timed reset, consider removing the manual `ghIsRateLimited = false` reset here, or at minimum only resetting it if the previous poll cycle completed without a rate-limit hit.

**4. `togglePanel` visibility widening**

```swift
// internal (not private) so AppDelegate+StatusItem.swift can reference
// it via #selector(togglePanel) from a separate file.
@objc func togglePanel() {
```

`#selector` works across files within the same module without changing access level from `private` to `internal` — `private` in Swift means file-private at the `extension` granularity only when the extension is in the same file. Since `AppDelegate+StatusItem.swift` is a separate file, `internal` is correct here. The comment is accurate and explains the intent well. ✅

**5. `_ansiRegex` force-try**

```swift
// swiftlint:disable:next force_try
private let _ansiRegex = try! NSRegularExpression(pattern: "...")
```

The disable comment and the accompanying doc comment (`/// Compiled once at load time. The pattern is a string literal and never fails.`) are sufficient justification. No issue, just confirming this is intentional. ✅

**6. `freezeVanishedGroups` — key inconsistency between old and new implementations**

The old `RunnerStore.freezeVanishedGroups` iterated over `snapPrev` keyed by `sha`:
```swift
for (sha, group) in snapPrev where !liveIDs.contains(sha) {
    cache[sha] = frozen
}
```
The new `PollResultBuilder.freezeVanishedGroups` iterates by `groupID`:
```swift
for (groupID, group) in snapPrev where !liveIDs.contains(groupID) {
    cache[groupID] = frozen
}
```
The doc comment on the new version explicitly calls this out and warns against mixing `headSha` and `id` keys. This looks correct — just flagging it deserves a test case to guard against regression.

---

### 🔧 Nits

- `RunnerStatusEnricher.fetchRunnersForScope` parses `scopeURL` by string-stripping `"https://github.com/"` rather than using `URL.pathComponents`. This is brittle if `scopeURL` ever uses `http://` or lacks the prefix. Consider using `URL(string:)?.pathComponents` for robustness, or routing through the existing `Scope.parse` helper.
- `AppDelegate+StatusItem.swift` has a multiline `❌ NEVER` comment block with `/// If you are an agent or human, DO NOT REMOVE THIS COMMENT, YOU ARE NOT ALLOWED UNDER ANY CIRCUMSTANCE.` — this reads as an AI-directed guardrail rather than standard engineering documentation. Fine to keep, just noting the unusual style.
- Doc file renames (`github_permission_rational.md` → `GitHub-Permission-Rationale.md`, `dev_with_log_in_terminal.md` → `dev-with-log-in-terminal.md`) are good — kebab-case is more idiomatic for docs.

---

**Summary:** The core metric-assignment fix is sound and the surrounding refactoring is high quality. The three concerns worth a follow-up are the `nameToAPI` collision risk in `RunnerStatusEnricher`, the unconditional `ghIsRateLimited = false` reset, and the lack of a regression test for `freezeVanishedGroups` key semantics. None are blockers. 👍

eonist:
## Code Review — PR #693: `fix(#655): match runner metrics by installPath, not slot index`

This is a solid, well-structured PR. The core bug fix is correct and the surrounding refactoring is generally high quality. Here are my observations:

---

### ✅ What's Done Well

- **Core fix is correct.** The old slot-index approach for assigning `RunnerMetrics` was fragile by design — two busy runners on the same machine could silently swap metrics. Matching by `installPath` via `metricsForRunner(installPath:)` is the right fix.
- **`Scope` enum is a great addition.** Replacing scattered `scope.contains("/")` string checks with a typed `Scope.parse()` + `apiPrefix` is cleaner, less error-prone, and centralised.
- **`ProcessRunner` extraction is excellent.** The old duplicated `Process` + `Pipe` + timeout boilerplate in `GitHubCLITransport` and `RunnerLifecycleService` is now a single well-documented primitive. The `task.isRunning` guard before `terminate()` in the timeout handler is a nice touch.
- **Rate-limit handling is much improved.** The new `RateLimitState` struct with atomic cancel-and-replace `DispatchWorkItem` for the reset timer is significantly more robust than the previous bare boolean, and the CLI fallback skip when rate-limited is a correct defensive addition.
- **`onChange` → Combine (`didUpdate`, `didMutate`, `didSignOut`) migration** removes hidden retain-cycle risks and aligns the codebase with a consistent reactive pattern.
- **`RunnerStore` annotated `@MainActor`** — good; this was implicitly assumed before.
- **`AppDelegate` decomposition** into `+PanelSetup` and `+StatusItem` extensions is a clean separation of concerns.
- **Logger upgrade to `os.Logger`** is the correct approach for a macOS app; zero-cost `.debug` elision in release builds is a meaningful improvement.
- **`PollResultBuilder` extraction** makes the state-building logic independently unit-testable. 

---

### ⚠️ Issues & Suggestions

#### 1. `buildInstallPathMap` — Scope key alignment may be unreliable

```swift
for scope in scopes {
    byFullKey["\(scope)/\(localRunner.runnerName)"] = path
}
```

`scopes` comes from `ScopeStore.shared.activeScopes`, which are user-entered strings (e.g. `"myorg/myrepo"`). But `fetchRunners(for:)` returns runners fetched for that scope string. If a runner is org-scoped (registered at the org level, not a repo), the full key `"myorg/myrepo/runnerName"` will never match `"myorg/runnerName"`. The name-only fallback catches this, but the log message warns about it as an alignment issue rather than treating it as the expected org-runner case. Consider documenting this explicitly or building the key differently for org vs repo scopes.

#### 2. `runGHProcess` is now `internal` — potential misuse surface

`runGHProcess` was `private` and is now `func` (internal) to allow `GitHub.swift` to call it directly. This is functional, but it exposes a raw subprocess launcher to the whole module. If `fetchStepLog` is the only cross-file caller, consider keeping it `private` in `GitHubCLITransport` and extracting just the needed logic differently, or at minimum add a comment explaining why it's internal.

#### 3. `freezeVanishedGroups` key inconsistency between old and new implementation

In `RunnerStoreState.swift` (old code), `freezeVanishedGroups` iterated `snapPrev` keyed by `sha` and wrote to `cache` also keyed by `sha`. In `PollResultBuilder.swift` (new code), both are keyed by `ActionGroup.id`. The doc comment notes this correctly, but the behaviour change (from sha-keyed to id-keyed) is subtle and could cause existing cached groups to not be evicted correctly on first run after update. Worth a targeted test or migration note.

#### 4. `_ansiRegex` — `try!` with a comment is acceptable, but placement is unusual

```swift
// swiftlint:disable:next force_try
private let _ansiRegex = try! NSRegularExpression(pattern: "...")
```

The comment correctly notes this is a compile-time constant that can never fail. However, placing a file-level `private let` mid-function-region (between `fetchStepLog` helpers) is a bit surprising. Consider moving it to a `// MARK: - Constants` section at the top of the file for discoverability.

#### 5. `OAuthService.signOut` — `onCompletion` no longer called

```swift
// Before:
onCompletion?(false)

// After:
didSignOut.send()
```

`SettingsView.onAppearAction` now subscribes to `didSignOut` to update `isOAuthAuthenticated`/`isCLIAuthenticated`. But `OAuthService.onCompletion` is still registered in `onAppearAction` for sign-in success. Ensure there's no code path that previously relied on `onCompletion?(false)` for sign-out in a context that hasn't been migrated to `didSignOut`. The `SettingsView` migration looks complete — just worth a search for any remaining `onCompletion` subscribers.

#### 6. Minor: `togglePanel` visibility

```swift
// internal (not private) so AppDelegate+StatusItem.swift can reference
// it via #selector(togglePanel) from a separate file.
@objc func togglePanel() {
```

This is fine functionally. An alternative would be to keep it `private` and move `setupStatusItem()` back to the same file, but the current approach is reasonable given the file decomposition goal.

---

### 🟡 Nits

- `PopoverMainView` no longer has `onSelectJob` or `onSelectAction` closures — the ABI-compatibility stubs are cleanly removed. ✓
- Deprecated files (`StatusIcon.swift`, `RunnerConfigSheet.swift`, `PopoverView.swift`, `PopoverHeightKey.swift`, `StatusPanel.swift`) removed cleanly — good housekeeping.
- Doc renames (`github_permission_rational.md` → `GitHub-Permission-Rationale.md`) and preference file moves to `Preferences/` folder are sensible.
- `RunnerStoreObservable` → `RunnerViewModel` rename is more semantically accurate.

---

### Summary

The core fix (#655) is correct and the approach is sound. The surrounding refactoring is largely an improvement. The main things worth reviewing before merge are the **scope key alignment** for org-scoped runners (#1 above) and verifying **no sign-out code paths remain on `onCompletion`** (#5). Everything else is low-risk.

**Verdict: Approve with minor follow-up suggestions** ✅

eonist:
## Code Review

This is a large, high-quality refactor PR that fixes a real metrics-misassignment bug while doing significant structural cleanup. The core fix is correct and well-reasoned. Here are my detailed findings:

---

### ✅ Core Fix (`RunnerStore.swift`)

The fix in `fetchAndEnrichRunners()` is the right approach. Replacing the slot-index-based `allWorkerMetrics()` assignment with per-runner `metricsForRunner(installPath:)` lookups is strictly more correct — the old code would silently corrupt metrics whenever runner counts or ordering didn't align perfectly.

The two-tier lookup (scope-prefixed `fullKey` → name-only fallback) is a thoughtful defense against scope string mismatches, and `buildInstallPathMap` is clean.

**Minor concern:** The name-only fallback (`installPathByRunnerName`) could match the *wrong* runner if two runners on the machine share the same `runnerName` across different repos. This is an edge case, but worth a comment noting the assumption that runner names are unique on the local machine.

---

### ✅ `Scope` enum (`Scope.swift`)

Replacing scattered `scope.contains("/")` string checks with a typed `Scope` enum is a clear improvement. The `apiPrefix` computed property centralises routing logic and removes a class of copy/paste bugs. The `parse(_:)` implementation correctly handles both `owner/repo` and `org` forms with empty-string guards.

---

### ✅ `ProcessRunner` (`ProcessRunner.swift`)

Extracting the duplicated `Process`/`Pipe`/timeout pattern into a single `ProcessRunner.run(...)` is a solid DRY improvement. The implementation is correct: stdin is written after `task.run()`, the timeout guard checks `task.isRunning` before terminating (avoiding a rare crash), and the `mergeStderr` flag is a useful addition.

---

### ✅ Rate-limit handling (`GitHubURLSessionTransport.swift`)

The `RateLimitState` struct with atomic cancel-and-replace `DispatchWorkItem` is a significant correctness improvement over the plain `Bool`. Parsing `X-RateLimit-Reset` from response headers and clamping to `[5s, 7200s]` is sensible. The CLI fallback is now correctly suppressed when `ghIsRateLimited` is set.

---

### ✅ `PollResultBuilder` (`PollResultBuilder.swift`)

Extracting state-building logic into a pure, static struct is excellent for testability. All methods receive their dependencies as closure parameters, making unit testing straightforward without needing a live `RunnerStore`. The `freezeVanishedGroups` key-by-`ActionGroup.id` (vs. the old SHA-keyed version) is an intentional and correct fix, and the doc-comment makes this explicit.

---

### ⚠️ Issues / Suggestions

**1. `runGHProcess` visibility change (`GitHubCLITransport.swift`)**
`runGHProcess` was `private` and is now `internal`. This exposes a blocking, main-thread-unsafe function at module scope. The callsite is already on a background thread so it's functionally fine, but consider adding a `// ⚠️ Must be called from a background thread` comment consistent with `ProcessRunner.run`'s existing doc-comment.

**2. `ghIsRateLimited = false` in `Task.detached` body (`RunnerStore.swift`)**
```swift
Task.detached(priority: .background) { [weak self] in
    ghIsRateLimited = false   // written from non-isolated context
    ...
}
```
`ghIsRateLimited` is backed by an `OSAllocatedUnfairLock` so the write is thread-safe. However, resetting the flag unconditionally at the start of every poll cycle means a rate-limit set mid-previous-cycle can be silently cleared before the scheduled auto-reset fires. The `scheduleRateLimitReset` auto-reset logic is designed to be the authoritative clear path — consider removing this manual reset, or at minimum document why it coexists with the auto-reset timer.

**3. `@MainActor` on `RunnerStore` + `nonisolated` proliferation**
`RunnerStore` is now `@MainActor`, which is correct. However, many methods are `nonisolated` to be callable from `Task.detached`. This is valid Swift, but those methods can no longer safely access any main-actor-isolated stored properties. Double-check that `nonisolated` methods never read `self.runners`, `self.jobs`, etc. without a `MainActor.run {}` hop — current code appears correct, but this is a footgun worth watching as the file evolves.

**4. `SettingsView` sign-out Combine subscription (`SettingsView.swift`)**
`onAppearAction` re-assigns both `signOutCancellable` and `scopeMutateCancellable` every time the view appears (e.g. after a sheet dismissal), cancelling and re-creating subscriptions each time. Harmless but slightly wasteful — consider guarding with `if signOutCancellable == nil { ... }`.

---

### Minor / Cosmetic ✅

- Renaming `RunnerStoreObservable` → `RunnerViewModel` is a better name — the "Observable" suffix was redundant.
- Moving `LegalPrefsStore`, `NotificationPrefsStore`, `SettingsStore` to `Preferences/` is a good file organisation improvement.
- `Logger.swift` upgrade to `os.Logger` with `.debug` level is the right direction — zero-cost in release builds is a real win.
- Doc file renames fix both the typo (`rational` → `Rationale`) and casing convention.
- Removing `PopoverHeightKey`, `PopoverView`, `RunnerConfigSheet`, `StatusIcon`, `StatusPanel` is clean.

---

### Overall Verdict

**✅ Approve with minor suggestions.** The core fix is correct and well-motivated. The refactoring leaves the codebase in a substantially better state. Main items to follow up on: the `ghIsRateLimited = false` reset interaction with the auto-reset timer, and a doc-comment on the name-only runner lookup uniqueness assumption.

Issue #941, opened 2026-05-25
Phase 4: Add ui-tests.yml GitHub Actions workflow


eonist:
## Parent
Sub-issue of #936

## Goal
Add the CI workflow that runs UI tests automatically on the self-hosted runner on every push to `main` and every pull request.

### `.github/workflows/ui-tests.yml`

```yaml
name: UI Tests

on:
  push:
    branches: [main]
  pull_request:

jobs:
  ui-test:
    runs-on: self-hosted

    steps:
      - uses: actions/checkout@v4

      - name: Install XcodeGen (if not cached)
        run: which xcodegen || brew install xcodegen

      - name: Generate ephemeral Xcode project
        run: xcodegen generate

      - name: Run UI Tests
        run: |
          xcodebuild test \
            -project RunnerBar.xcodeproj \
            -scheme RunnerBarUITests \
            -destination 'platform=macOS' \
            CODE_SIGNING_REQUIRED=NO \
            CODE_SIGN_IDENTITY=""
        env:
          UI_TESTING: "1"
```

## Notes
- `runs-on: self-hosted` — cloud runners have no WindowServer; this must run locally
- `UI_TESTING: "1"` is passed as an env var to `xcodebuild`, which forwards it to the launched app process
- The generated `.xcodeproj` is discarded after the job; `.gitignore` ensures it's never committed
- XcodeGen must be pre-installed on the runner machine (`brew install xcodegen`)

## Depends On
- #938 (Phase 1), #939 (Phase 2), #940 (Phase 3) — all must be merged first

## Acceptance Criteria
- [ ] `.github/workflows/ui-tests.yml` committed
- [ ] Workflow triggers automatically on push and PRs
- [ ] All three UI tests pass in the Actions log
- [ ] No `.xcodeproj` appears in `git status` post-run
- [ ] Zero manual keychain approval prompts

Issue #940, opened 2026-05-25
Phase 3: Add RunnerBarUITests test target and test file


eonist:
## Parent
Sub-issue of #936

## Goal
Add the `Tests/RunnerBarUITests/` directory and the initial `XCUIApplication`-based test file. This is the actual UI test code that drives the app on the self-hosted runner.

## Important: NSPanel, not NSPopover
The app uses `NSPanel` (a borderless non-activating panel) instead of `NSPopover`. Tests must query `app.windows.firstMatch` — **not** `app.popovers.firstMatch`, which will always return zero results.

### `Tests/RunnerBarUITests/RunnerBarUITests.swift`

```swift
import XCTest

final class RunnerBarUITests: XCTestCase {

    var app: XCUIApplication!

    override func setUp() {
        super.setUp()
        continueAfterFailure = false
        app = XCUIApplication()
        app.launchEnvironment["UI_TESTING"] = "1"
        app.launch()
    }

    override func tearDown() {
        app.terminate()
        super.tearDown()
    }

    func testAppLaunchesWithoutCrashing() {
        XCTAssertTrue(app.wait(for: .runningForeground, timeout: 5))
    }

    func testStatusBarItemExists() {
        let statusItem = app.statusItems.firstMatch
        XCTAssertTrue(statusItem.waitForExistence(timeout: 3))
    }

    func testPanelOpensOnClick() {
        // App uses NSPanel, not NSPopover — query via app.windows
        let statusItem = app.statusItems.firstMatch
        guard statusItem.waitForExistence(timeout: 3) else {
            XCTFail("Status item not found")
            return
        }
        statusItem.click()
        let panel = app.windows.firstMatch
        XCTAssertTrue(panel.waitForExistence(timeout: 3))
    }
}
```

## Depends On
- #938 (Phase 1): `UI_TESTING` guard must be in place so the app doesn't hit keychain on launch
- #939 (Phase 2): `project.yml` must exist so XcodeGen can pick up the `RunnerBarUITests` target

## Acceptance Criteria
- [ ] `Tests/RunnerBarUITests/RunnerBarUITests.swift` committed
- [ ] `xcodegen generate` includes `RunnerBarUITests` scheme in the generated `.xcodeproj`
- [ ] `testAppLaunchesWithoutCrashing` passes on self-hosted runner
- [ ] `testStatusBarItemExists` passes on self-hosted runner
- [ ] `testPanelOpensOnClick` passes on self-hosted runner

Issue #939, opened 2026-05-25
Phase 2: Add project.yml for XcodeGen and update .gitignore


eonist:
## Parent
Sub-issue of #936

## Goal
Add the `project.yml` XcodeGen spec to the repo root. This is the only file that replaces a maintained `.xcodeproj`. Also update `.gitignore` so the generated `.xcodeproj` is never accidentally committed.

### `project.yml` (repo root)

```yaml
name: RunnerBar
options:
  bundleIdPrefix: com.eoncode
  createIntermediateGroups: true

targets:
  RunnerBar:
    type: application
    platform: macOS
    deploymentTarget: "26.0"
    sources:
      - Sources/RunnerBar
    resources:
      - Resources
    settings:
      base:
        PRODUCT_BUNDLE_IDENTIFIER: com.eoncode.RunnerBar
        CODE_SIGN_IDENTITY: ""
        CODE_SIGNING_REQUIRED: NO
        INFOPLIST_FILE: Resources/Info.plist

  RunnerBarUITests:
    type: bundle.ui-testing
    platform: macOS
    deploymentTarget: "26.0"
    sources:
      - Tests/RunnerBarUITests
    dependencies:
      - target: RunnerBar
    settings:
      base:
        CODE_SIGN_IDENTITY: ""
        CODE_SIGNING_REQUIRED: NO
```

### `.gitignore` additions

```
*.xcodeproj
*.xcworkspace
```

## Notes
- `deploymentTarget: "26.0"` matches `macOS(.v26)` in `Package.swift`
- `INFOPLIST_FILE: Resources/Info.plist` matches the confirmed location of the plist
- The `.xcodeproj` is generated ephemerally in CI by `xcodegen generate` and discarded — never committed

## Acceptance Criteria
- [ ] `project.yml` committed to repo root
- [ ] `*.xcodeproj` and `*.xcworkspace` added to `.gitignore`
- [ ] `xcodegen generate` runs without errors locally
- [ ] Generated `.xcodeproj` does not appear in `git status`

Issue #938, opened 2026-05-25
Phase 1: Add UI_TESTING guard to skip keychain/network on launch


eonist:
## Parent
Sub-issue of #936

## Goal
Add a `UI_TESTING` environment variable guard so the app skips all keychain access and network polling when launched by the UI test runner. Without this, every ad-hoc-signed CI build triggers a macOS keychain approval dialog, breaking fully automated test runs.

## Where
`Sources/RunnerBar/App/AppDelegate+PanelSetup.swift` — inside `setupCombineSubscriptions()`, which is where `RunnerStore.shared.start()` is currently called.

## What to Change

In `setupCombineSubscriptions()`, wrap the `RunnerStore.shared.start()` call and the `ScopeStore.didMutate` subscription that re-triggers it:

```swift
private func setupCombineSubscriptions() {
    // LocalRunnerStore subscription is safe — no keychain/network
    LocalRunnerStore.shared.$runners
        .receive(on: DispatchQueue.main)
        .sink { [weak self] _ in
            guard let self else { return }
            self.observable.reload(localRunnerStore: LocalRunnerStore.shared)
        }
        .store(in: &cancellables)

    // Skip all network + keychain activity during UI tests
    guard ProcessInfo.processInfo.environment["UI_TESTING"] == nil else { return }

    RunnerStore.shared.didUpdate
        .receive(on: DispatchQueue.main)
        .sink { [weak self] in
            guard let self else { return }
            self.updateStatusIcon()
            self.observable.reload(localRunnerStore: LocalRunnerStore.shared)
        }
        .store(in: &cancellables)

    RunnerStore.shared.start()

    ScopeStore.shared.didMutate
        .receive(on: DispatchQueue.main)
        .sink { [weak self] in
            guard self != nil else { return }
            RunnerStore.shared.start()
        }
        .store(in: &cancellables)
}
```

## Acceptance Criteria
- [ ] `UI_TESTING` env check added in `setupCombineSubscriptions()`
- [ ] App still polls GitHub normally when `UI_TESTING` is not set
- [ ] `swift build` succeeds
- [ ] `swift test` (unit tests) still pass

PR #947, opened 2026-05-25
feat: XcodeGen UITest setup — hidden, SPM-first, ephemeral xcodeproj (#936) (pr947)

### LINKED ISSUE #936: Add XcodeGen-based UITest setup (hidden, SPM-first, ephemeral xcodeproj)
## Goal

Add real `XCUIApplication`-based UI tests that run headlessly on the self-hosted runner, with no `.xcodeproj` committed to the repo. The Xcode project is generated on-the-fly by **XcodeGen** from a committed `project.yml` spec and discarded after each CI run.

## Background

- `swift test` cannot drive `XCUIApplication` — it requires `xcodebuild`
- `xcodebuild` requires an `.xcodeproj` or `.xcworkspace`
- `swift package generate-xcodeproj` was removed in Swift 5.8+
- **Solution:** XcodeGen generates an ephemeral `.xcodeproj` from a simple YAML spec at CI time; the file is gitignored and never committed
- `RunnerBarCore` is already extracted — no architecture changes needed
- The app already has `LSUIElement = YES` in `Resources/Info.plist` — the test run is completely silent (no Dock icon, no windows)
- The app uses **`NSPanel`** (not `NSPopover`) — panel tests must use `app.windows`, not `app.popovers`
- The app reads from Keychain on launch — a `UI_TESTING` guard must be added to skip network/keychain access, otherwise macOS will show an approval dialog for every new ad-hoc-signed build

### 1. Add `UI_TESTING` guard to `Sources/RunnerBar/main.swift`

This prevents keychain access and network polling during UI tests, avoiding macOS keychain approval dialogs in CI:

```swift
// main.swift
import AppKit

MainActor.assumeIsolated {
    let delegate = AppDelegate()
    NSApplication.shared.delegate = delegate
    // Skip network + keychain during UI tests
    if ProcessInfo.processInfo.environment["UI_TESTING"] == nil {
        RunnerStore.shared.start()
    }
    NSApplication.shared.run()
}
```

> ⚠️ Note: `RunnerStore.shared.start()` is currently called inside `setupCombineSubscriptions()` in `AppDelegate+PanelSetup.swift`. The guard should be placed there (or both places) — confirm exact call site before implementing.

### 2. Add `project.yml` to repo root

```yaml
name: RunnerBar
options:
  bundleIdPrefix: com.eoncode
  createIntermediateGroups: true

targets:
  RunnerBar:
    type: application
    platform: macOS
    deploymentTarget: "26.0"
    sources:
      - Sources/RunnerBar
    resources:
      - Resources
    settings:
      base:
        PRODUCT_BUNDLE_IDENTIFIER: com.eoncode.RunnerBar
        CODE_SIGN_IDENTITY: ""
        CODE_SIGNING_REQUIRED: NO
        INFOPLIST_FILE: Resources/Info.plist

  RunnerBarUITests:
    type: bundle.ui-testing
    platform: macOS
    deploymentTarget: "26.0"
    sources:
      - Tests/RunnerBarUITests
    dependencies:
      - target: RunnerBar
    settings:
      base:
        CODE_SIGN_IDENTITY: ""
        CODE_SIGNING_REQUIRED: NO
```

### 3. Add `Tests/RunnerBarUITests/RunnerBarUITests.swift`

> ⚠️ The app uses `NSPanel` (not `NSPopover`) — use `app.windows.firstMatch`, not `app.popovers.firstMatch`.

```swift
import XCTest

final class RunnerBarUITests: XCTestCase {

    var app: XCUIApplication!

    override func setUp() {
        super.setUp()
        continueAfterFailure = false
        app = XCUIApplication()
        app.launchEnvironment["UI_TESTING"] = "1"
        app.launch()
    }

    override func tearDown() {
        app.terminate()
        super.tearDown()
    }

    func testAppLaunchesWithoutCrashing() {
        XCTAssertTrue(app.wait(for: .runningForeground, timeout: 5))
    }

    func testStatusBarItemExists() {
        let statusItem = app.statusItems.firstMatch
        XCTAssertTrue(statusItem.waitForExistence(timeout: 3))
    }

    func testPanelOpensOnClick() {
        // App uses NSPanel, not NSPopover — query via app.windows
        let statusItem = app.statusItems.firstMatch
        guard statusItem.waitForExistence(timeout: 3) else {
            XCTFail("Status item not found")
            return
        }
        statusItem.click()
        let panel = app.windows.firstMatch
        XCTAssertTrue(panel.waitForExistence(timeout: 3))
    }
}
```

### 4. Add `.github/workflows/ui-tests.yml`

```yaml
name: UI Tests

on:
  push:
    branches: [main]
  pull_request:

jobs:
  ui-test:
    runs-on: self-hosted

    steps:
      - uses: actions/checkout@v4

      - name: Install XcodeGen (if not cached)
        run: which xcodegen || brew install xcodegen

      - name: Generate ephemeral Xcode project
        run: xcodegen generate

      - name: Run UI Tests
        run: |
          xcodebuild test \
            -project RunnerBar.xcodeproj \
            -scheme RunnerBarUITests \
            -destination 'platform=macOS' \
            CODE_SIGNING_REQUIRED=NO \
            CODE_SIGN_IDENTITY=""
        env:
          UI_TESTING: "1"
```

### 5. Add to `.gitignore`

```
*.xcodeproj
*.xcworkspace
```

### 6. One-time self-hosted runner setup

- Install XcodeGen: `brew install xcodegen`
- Confirm Xcode command-line tools are present: `xcode-select -p`
- Confirm the runner agent runs in a GUI session (required for `XCUIApplication` + `NSStatusItem`)
- No keychain approval dialogs will appear once the `UI_TESTING` guard is in place

## Acceptance Criteria

- [ ] `UI_TESTING` env guard added in app — skips keychain/network when set
- [ ] `project.yml` committed to repo root (macOS 26.0, `Resources/Info.plist`)
- [ ] `Tests/RunnerBarUITests/RunnerBarUITests.swift` committed (uses `app.windows`, not `app.popovers`)
- [ ] `.github/workflows/ui-tests.yml` committed
- [ ] `*.xcodeproj` / `*.xcworkspace` added to `.gitignore`
- [ ] `testAppLaunchesWithoutCrashing` passes on self-hosted runner
- [ ] `testStatusBarItemExists` passes on self-hosted runner
- [ ] `testPanelOpensOnClick` passes on self-hosted runner
- [ ] No `.xcodeproj` file appears in `git status` after CI run
- [ ] CI runs fully automatically — no manual keychain approval prompt

## Notes

- `swift test` (unit tests, SwiftLint) continues unchanged on cloud runners — this workflow only runs on the self-hosted runner
- XcodeGen `project.yml` is plain YAML — small, diffable, AI-editable, and generates no merge conflicts
- Supersedes #935 (which relied on `XCUIApplication` via `swift test`, which is not possible)

eonist:
## **User description**
Closes #936

## Phases

| Phase | Issue | Status |
|---|---|---|
| 1 — `UI_TESTING` guard | #938 | ✅ Committed |
| 2 — `project.yml` + `.gitignore` | #939 | ⏳ Pending |
| 3 — `RunnerBarUITests` test file | #940 | ⏳ Pending |
| 4 — `ui-tests.yml` workflow | #941 | ⏳ Pending |

## What's in this PR so far

**Phase 1** (`AppDelegate+PanelSetup.swift`): Added a `UI_TESTING` environment variable guard in `setupCombineSubscriptions()`. When the app is launched by `XCUIApplication` during automated UI tests, it skips `RunnerStore.shared.start()`, the `didUpdate` subscription, and the `ScopeStore.didMutate` re-trigger — preventing all keychain access and network polling. The `LocalRunnerStore` subscription is preserved as it has no keychain/network side effects.

This ensures CI runs fully automatically with zero macOS keychain approval dialogs.

## Testing
- `swift build` must pass
- `swift test` (unit tests) must pass — `RunnerBarCoreTests` unaffected
- Normal app launch (no `UI_TESTING` env var) must behave identically to before

## Summary by CodeRabbit

* **Tests**
  * Added UI tests that relaunch the app with the panel open, verify a “Workflows” header, and run without remote/network updates during UI testing.
* **Chores**
  * CI workflow now runs UI tests on pull requests; added project configuration and exported core library product.
* **Documentation**
  * Added UI test runner setup and an extensive UITest troubleshooting log.
* **Style**
  * .gitignore updated to exclude generated Xcode project/workspace files.

[Review Change Stack](https://app.coderabbit.ai/change-stack/eoncode/runner-bar/pull/947?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

___

## **CodeAnt-AI Description**
**Skip network and keychain activity during UI tests**

### What Changed
- When the app is launched for automated UI tests, it now skips the parts that poll for remote runner updates and touch the keychain
- Local runner updates still run, so the app can refresh its local state without triggering approval dialogs
- Normal app launches keep the same behavior as before

### Impact
`✅ Fewer keychain approval dialogs in CI`
`✅ Cleaner UI test launches`
`✅ More reliable automated runs`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


eonist (commented):
## Overall Assessment ✅

This is a clean, well-scoped PR. The `UI_TESTING` guard in `AppDelegate+PanelSetup.swift` is a solid, minimal approach to preventing keychain/network dialogs during CI. The diff is small (196 additions, 6 files) and each file has a clear, single purpose.

---

## `AppDelegate+PanelSetup.swift`

The guard placement is correct — it runs *after* the safe `LocalRunnerStore` subscription and *before* the network/keychain-touching `RunnerStore` subscriptions. One minor note: using `== nil` to check absence reads slightly awkward. Consider naming it explicitly:

```swift
let isUITesting = ProcessInfo.processInfo.environment["UI_TESTING"] != nil
guard !isUITesting else { return }
```

This makes the intent more readable for future contributors.

---

## `RunnerBarUITests.swift`

The three smoke tests cover the right surface area — launch, status item presence, and panel open. The `guard`-based failure in `testPanelOpensOnClick` is a good pattern (avoids a force-unwrap crash in CI). One suggestion: `testStatusBarItemExists` and `testPanelOpensOnClick` overlap — if the status item doesn't exist, the panel test will also fail. Consider combining them into a single sequential test to avoid redundant failures in CI output.

---

## `ui-tests.yml` Workflow

The fallback `xcodebuild` call (after `xcpretty`) is smart — it ensures the raw exit code isn't swallowed by `xcpretty`. However, the Homebrew auto-install step is risky on a self-hosted runner in CI: it could trigger interactive prompts or take unexpectedly long. A `HOMEBREW_NO_AUTO_UPDATE=1` env variable would make it safer:

```yaml
env:
  HOMEBREW_NO_AUTO_UPDATE: "1"
```

Also, `sudo xcode-select -s /Applications/Xcode.app/Contents/Developer` hardcodes the Xcode path — worth noting in a comment that this assumes Xcode.app is at the default location.

---

## `project.yml`

`deploymentTarget: "26.0"` stands out — macOS 26 is not yet released. This should likely be `"14.0"` or `"15.0"` unless you're intentionally targeting a future SDK. Worth double-checking before this gets merged.

---

## `docs/TESTING.md`

Very terse — consider adding the *why* (accessibility is required for `XCUIApplication` to interact with the menu bar status item) and a note about what happens if this step is skipped (tests will silently fail or hang). A one-liner example command would also help.

---

## Summary of Action Items

| Severity | File | Issue |
|---|---|---|
| 🔴 High | `project.yml` | `deploymentTarget: "26.0"` likely wrong |
| 🟡 Medium | `ui-tests.yml` | Add `HOMEBREW_NO_AUTO_UPDATE=1` |
| 🟢 Low | `AppDelegate+PanelSetup.swift` | Readability of `UI_TESTING` nil-check |
| 🟢 Low | `RunnerBarUITests.swift` | Overlap between status item and panel tests |
| 🟢 Low | `docs/TESTING.md` | Too minimal — add context and failure notes |

eonist:
## Host-App Pattern — The Architectural Fix for `app.windows` on LSUIElement

The chain of failures (`app.windows` empty → `.popUpMenu` invisible → `.floating` still empty → `XCUIScreen.main.windows` as workaround) all stem from one root cause: **XCTest's AX server scopes `app.windows` to a process handle, and LSUIElement background agents are outside that activation context**. The workaround with `XCUIScreen.main.windows` works but is unscoped — it grabs all windows on screen and will grab the wrong one if anything else is visible.

The real fix is a **minimal host app target** — a normal `NSApplication` (no `LSUIElement`) whose only job is to host the panel during UI tests. The production `RunnerBar` app is completely untouched.

---

### What to add

**`Tests/RunnerBarUITestHost/main.swift`** (~30 lines):

```swift
// Minimal NSApplication host for UI tests.
// NOT LSUIElement — normal foreground app so XCTest's AX server can see app.windows.
import AppKit
import RunnerBarCore

final class UITestHostDelegate: NSObject, NSApplicationDelegate {
    var window: NSWindow?

    func applicationDidFinishLaunching(_ notification: Notification) {
        let panel = NSPanel(
            contentRect: NSRect(x: 0, y: 0, width: 320, height: 500),
            styleMask: [.titled, .fullSizeContentView, .nonactivatingPanel],
            backing: .buffered,
            defer: false
        )
        panel.level = .floating   // AX-visible; no .popUpMenu tricks needed
        panel.isReleasedWhenClosed = false
        panel.contentViewController = NSHostingController(rootView: PanelMainView())
        panel.center()
        panel.orderFront(nil)
        self.window = panel
    }
}

let app = NSApplication.shared
app.setActivationPolicy(.regular)
let delegate = UITestHostDelegate()
app.delegate = delegate
app.run()
```

`PanelMainView` is importable from `RunnerBarCore` — already a library product in `Package.swift` after this PR.

---

**Add to `project.yml`:**

```yaml
  RunnerBarUITestHost:
    type: application
    platform: macOS
    deploymentTarget: "26.0"
    sources:
      - Tests/RunnerBarUITestHost
    dependencies:
      - package: RunnerBar
        product: RunnerBarCore
    settings:
      base:
        PRODUCT_BUNDLE_IDENTIFIER: dev.eonist.runnerbar.uitesthost
        MACH_O_TYPE: mh_execute
        CODE_SIGN_IDENTITY: "-"
        CODE_SIGNING_REQUIRED: NO
        AD_HOC_CODE_SIGNING_ALLOWED: YES
        GENERATE_INFOPLIST_FILE: YES
        # NO LSUIElement key — that's the whole point
```

Update the `RunnerBarUITests` scheme to wire the host as `testTargetApp` (safe now because it's a normal app, not an LSUIElement agent — no Xcode 26 `XCTTargetAppPath` stripping bug):

```yaml
  RunnerBarUITests:
    ...
    test:
      targets:
        - name: RunnerBarUITests
          parallelizable: false
      testTargetApp: RunnerBarUITestHost
      environmentVariables:
        UI_TESTING: "1"
```

---

**`RunnerBarUITests.swift` — `app.windows` just works:**

```swift
override func setUp() {
    super.setUp()
    continueAfterFailure = false
    // Target the host, not the LSUIElement agent.
    app = XCUIApplication(bundleIdentifier: "dev.eonist.runnerbar.uitesthost")
    app.launchEnvironment["UI_TESTING"] = "1"
    app.launch()
}

func testPanelOpensAndShowsWorkflowsSection() {
    // Normal NSApplication → .runningForeground works.
    XCTAssertTrue(app.wait(for: .runningForeground, timeout: 5))

    // app.windows works. No XCUIScreen.main tricks. No panel-level hacks.
    let panel = app.windows.firstMatch
    XCTAssertTrue(panel.waitForExistence(timeout: 5))

    let workflowsHeader = app.staticTexts["Workflows"]
    XCTAssertTrue(workflowsHeader.waitForExistence(timeout: 5))
}
```

---

**`ui-tests.yml` — build the host, then test:**

```yaml
      - name: Build host app
        run: |
          xcodebuild build \
            -project RunnerBar.xcodeproj \
            -scheme RunnerBarUITestHost \
            -destination 'platform=macOS' \
            -derivedDataPath "$DERIVED_DATA"

      - name: Run UI Tests
        run: |
          xcodebuild test \
            -project RunnerBar.xcodeproj \
            -scheme RunnerBarUITests \
            -destination 'platform=macOS' \
            -derivedDataPath "$DERIVED_DATA" \
            -only-testing:RunnerBarUITests
```

---

### Why this is cleaner than `XCUIScreen.main.windows`

| | `XCUIScreen.main.windows` (current) | Host app |
|---|---|---|
| Scoped to your app | ❌ grabs all screen windows | ✅ `app.windows` is process-scoped |
| Ambiguous if other windows visible | ❌ `firstMatch` may be wrong | ✅ unambiguous |
| `OPEN_PANEL_ON_LAUNCH` hack needed | ✅ required | ❌ panel opens in `applicationDidFinishLaunching` |
| Panel level override needed | ✅ `.floating` override required | ❌ any level works |
| `runningBackground` workaround needed | ✅ required | ❌ `.runningForeground` works normally |

### One caveat

If `PanelMainView` reaches into `AppDelegate`-owned singletons (e.g. `RunnerStore.shared`), those need the `UI_TESTING` guard already in place — `launchEnvironment["UI_TESTING"] = "1"` covers it since the host process inherits the environment.

The host target only exists in `project.yml` (ephemeral, git-ignored), so there is zero production footprint.

eonist:
## Why `app.windows` is empty — and how to fix it without a new target

### The actual root cause

Changing panel level (`.popUpMenu` → `.floating`) was fixing the wrong variable. The failure chain is:

1. `app.windows` — empty because XCTest scopes its AX window query to the target process
2. `LSUIElement = YES` marks the process as an **agent** — it doesn't publish a window list through the standard AX hierarchy
3. Window level is irrelevant — the AX window list is empty at XCTest attach time, before any panel even exists
4. `XCUIScreen.main.windows` — unreliable; not scoped to our process, likely grabs wrong windows or nothing on CI

The reason `app.statusItems` works is that `NSStatusItem` registers its AX element on the **system-wide menu bar**, a completely different AX surface — not through the app's window list.

### The fix — no new target needed

The process just needs to be promoted to `.regular` before XCTest attaches its AX session. Gate it behind a launch argument:

```swift
// AppDelegate.applicationDidFinishLaunching
if ProcessInfo.processInfo.arguments.contains("--uitesting") {
    NSApp.setActivationPolicy(.regular)
    NSApp.activate(ignoringOtherApps: true)
}
```

```swift
// Test setUp
app.launchArguments = ["--uitesting"]
app.launch()
```

Once the process is `.regular`, `app.windows` populates normally and you can query panel content with full XCTest semantics. A Dock icon briefly appears during the test run — acceptable for CI.

### What you lose

You're not testing the exact production activation policy (`.accessory`). For functional panel content tests that's fine. For agent-specific behavior (no Dock icon, `.accessory` handling) you're limited to `app.statusItems`-level checks — but those already work.

### Alternative (no app code change)

Duplicate `Info.plist` as `Info-UITesting.plist` with `LSUIElement = NO`, add a `UITesting` build configuration, and point the test scheme at it. Same result, no runtime flag needed — but more XcodeGen surface area.

### Bottom line

The host app pattern is cleaner long-term, but it's not *necessary* to unblock functional panel UI tests. The `setActivationPolicy` approach is ~5 lines and has been the standard workaround for this since Xcode 7.

eonist:
## Addendum: No New Target Required — Two Lighter Approaches

The host-app pattern above is the cleanest architecturally, but `app.windows` can also be made to work without a new target.

---

### Option A: `setActivationPolicy` at Runtime (1 line, no new files)

`LSUIElement = YES` in the plist is just a default. `NSApp.setActivationPolicy(_:)` overrides it at runtime. If the app promotes itself to `.regular` during a `UI_TESTING` launch, XCTest's AX server treats it as a normal foreground process and `app.windows` works:

```swift
// AppDelegate+PanelSetup.swift, in setupPanel() or applicationDidFinishLaunching:
if ProcessInfo.processInfo.environment["UI_TESTING"] != nil {
    NSApp.setActivationPolicy(.regular) // or .accessory
}
```

Test code then becomes:

```swift
// .runningForeground works now — not .runningBackground
XCTAssertTrue(app.wait(for: .runningForeground, timeout: 5))

// app.windows works — no XCUIScreen.main needed
let panel = app.windows.firstMatch
XCTAssertTrue(panel.waitForExistence(timeout: 5))
```

No new target, no plist duplication, no `project.yml` change. The `UI_TESTING` env var is already being passed.

---

### Option B: Separate `Info-UITesting.plist` + Build Configuration

Create `Resources/Info-UITesting.plist` identical to `Info.plist` but with `LSUIElement` removed (or set to `NO`). Add a `UITesting` build configuration in `project.yml` that points to it:

```yaml
  RunnerBar:
    settings:
      configs:
        UITesting:
          INFOPLIST_FILE: Resources/Info-UITesting.plist
```

Set the test action in the `RunnerBarUITests` scheme to use the `UITesting` configuration. Heavier than Option A but keeps the change entirely out of production code paths.

---

### Recommendation

Try **Option A first** — it's one `if` block next to the existing `UI_TESTING` guard already in `setupCombineSubscriptions()`, and it eliminates all the `XCUIScreen.main` workarounds, the `.floating` level override, and the `runningBackground` dance in one go.

eonist:
### Why UI Testing `LSUIElement` Apps is Difficult

Testing a background menu bar app (`LSUIElement = true`) natively via `XCTest` without a host app is an infamous architectural constraint in macOS development. XCTest's Accessibility (AX) server scopes `app.windows` to the process handle of a foreground-activatable app. Because background agents run outside that activation context, the AX server simply returns an empty window list, regardless of the panel's window level (`.popUpMenu`, `.floating`, etc.).

There are two established patterns to solve this:

#### 1. The Host App Pattern (Cleanest)
Move the panel into a thin `NSApplication` host target (`RunnerBarUITestHost`) that launches only during UI tests. This isolates the UI into a launchable, foreground-activatable context where `app.windows` works perfectly. This is the approach generally recommended for testing background agent UIs.

#### 2. The Activation Policy Workaround (No New Target)
Dynamically convert the app from a background agent into a normal foreground app only when UI tests are running.

```swift
func applicationDidFinishLaunching(_ aNotification: Notification) {
    if ProcessInfo.processInfo.environment["UI_TESTING"] != nil {
        // Force the app to become a regular foreground app so XCTest can see the windows
        NSApp.setActivationPolicy(.regular) // or .accessory
    }
}
```

While this avoids creating a new target, it fundamentally alters the application's lifecycle and event loop during tests — meaning we end up testing an app state that users will never experience.

Between the two, creating a lightweight Host App that leverages our existing SPM module boundaries (`RunnerBarCore` is already exported as a library product in `Package.swift`) is generally safer and less fragile than overriding activation policies in the main app lifecycle.

eonist:
### Fix: `setActivationPolicy` must be called in `applicationWillFinishLaunching`

The test fails because `setActivationPolicy(.regular)` is being called too late (in `applicationDidFinishLaunching`). XCTest completes its launch handshake before the policy takes effect, so the app never reaches `.runningForeground`.

**1. Move the policy change to `applicationWillFinishLaunching` in `AppDelegate`:**

```swift
func applicationWillFinishLaunching(_ notification: Notification) {
    if ProcessInfo.processInfo.environment["UI_TESTING"] != nil {
        NSApp.setActivationPolicy(.regular)
        NSApp.activate(ignoringOtherApps: true)
    }
}
```

Using `applicationWillFinishLaunching` (not `DidFinish`) ensures the activation policy is set before XCTest's launch handshake completes, so it sees the app as a foreground process from the start. The `activate(ignoringOtherApps: true)` call is also needed to actually bring it into the foreground state.

**2. The panel test should then use `app.windows` instead of `XCUIScreen.main.windows`**, since with `.regular` activation policy the app is now a normal foreground process and `app.windows` works correctly — no screen-level hacks needed:

```swift
func testPanelOpensAndShowsWorkflowsSection() {
    app.terminate()
    app.launchEnvironment["UI_TESTING"] = "1"
    app.launchEnvironment["OPEN_PANEL_ON_LAUNCH"] = "1"
    app.launch()

    XCTAssertTrue(app.wait(for: .runningForeground, timeout: 5))

    // With .regular activation policy, app.windows works — no XCUIScreen.main needed
    let panel = app.windows.firstMatch
    XCTAssertTrue(panel.waitForExistence(timeout: 10))

    let workflowsHeader = app.staticTexts["Workflows"]
    XCTAssertTrue(workflowsHeader.waitForExistence(timeout: 5))
}
```

**Why `testStatusBarItemExists` passes but `testAppLaunchesWithoutCrashing` fails:** The status item check works because `app.statusItems` queries a dedicated AX role that is always reachable regardless of activation policy. The `.runningForeground` wait fails because the policy change fires after the XCTest handshake. Moving it to `applicationWillFinishLaunching` fixes the ordering.

eonist:
## `setActivationPolicy` Failed — Root Cause & What To Do

### What happened

`app.wait(for: .runningForeground)` timed out. The app never promoted itself into the foreground activation context, meaning `setActivationPolicy(.regular)` either ran too late or was silently ignored.

**Root cause:** the app is sandboxed (`com.apple.security.app-sandbox = YES` in the test bundle entitlements). macOS blocks a sandboxed LSUIElement process from promoting itself to `.regular` activation policy at runtime — the call returns without effect. This isn't a timing issue; it's a sandbox restriction.

`testStatusBarItemExists` still **passed** — the app launches fine and `app.statusItems` works. The sandbox only blocks the activation policy promotion, not AX queries on status items.

### Immediate fix: revert

Revert the two changes introduced by this experiment:

1. `app.wait(for: .runningForeground)` → back to `app.wait(for: .runningBackground)`
2. `app.windows` → back to `XCUIScreen.main.windows`

That's the state that was passing before. `XCUIScreen.main.windows` is not actually fragile here — the panel is the only window on screen (placed top-right via `setFrameOrigin`) so `firstMatch` is unambiguous in practice.

### What this means for the host-app approach

The host-app target sidesteps the sandbox restriction entirely because it's a **separate process** with its own entitlements — it doesn't inherit the LSUIElement + sandbox combination of the main app. If `app.windows` scoping is needed in the future, the host-app target remains the only viable path. `setActivationPolicy` is a dead end for sandboxed apps.

### Rule to add to `UITEST-LESSONS.md`

```
### ❌ `setActivationPolicy(.regular)` on a sandboxed LSUIElement app
**Problem:** Call has no effect — macOS silently blocks activation policy promotion for
sandboxed background agents. `app.wait(for: .runningForeground)` always times out.
**Fix:** Use `.runningBackground`. Use `XCUIScreen.main.windows` for panel queries.
**Rule:** For sandboxed LSUIElement apps, `setActivationPolicy` cannot be used to make
`app.windows` work. The only way to get `app.windows` is a separate non-LSUIElement
host app target.
```

eonist:
## CI failure diagnosis — `setActivationPolicy` not reaching `.runningForeground`

Both failing tests hit the same error:
```
Wait for dev.eonist.runnerbar to become Running Foreground
XCTAssertTrue failed - App should reach .runningForeground (promoted via setActivationPolicy(.regular) in UI_TESTING mode)
```

`testStatusBarItemExists` **passed** ✅ — so `--uitesting` is reaching the app and it launches fine. The problem is not the launch arg, it's the timing of the foreground state check.

### Root cause

`setActivationPolicy(.regular)` + `NSApp.activate(ignoringOtherApps: true)` is asynchronous. The process doesn't instantly reach `.runningForeground` — it needs a runloop cycle or two. If the test asserts `app.state == .runningForeground` immediately after `app.launch()`, it will fail.

### Fix

Replace the immediate assert with a proper XCTest wait:

```swift
// ❌ Wrong — races against async activation
app.launch()
XCTAssertTrue(app.state == .runningForeground)

// ✅ Correct — wait up to 5s for foreground state
app.launch()
let foreground = app.wait(for: .runningForeground, timeout: 5)
XCTAssertTrue(foreground, "App should reach .runningForeground")
```

Also make sure `AppDelegate` calls both lines, not just one:

```swift
if ProcessInfo.processInfo.arguments.contains("--uitesting") {
    NSApp.setActivationPolicy(.regular)
    NSApp.activate(ignoringOtherApps: true) // required — policy alone doesn't foreground
}
```

If `activate` is missing, `setActivationPolicy(.regular)` marks the process as a regular app type but won't actually move it to foreground — the test runner won't do that automatically.

Resources/Info.plist
```diff
@@ -16,8 +16,6 @@
     <string>APPL</string>
```
eonist, line 16: **`LSUIElement` removed from `Info.plist` breaks `swift build` / `build.sh`**

`LSUIElement` is now only set via `INFOPLIST_KEY_LSUIElement: YES` in `project.yml`, which is gitignored and ephemeral. The `swift build` / `build.sh` path never goes through XcodeGen, so production builds via that path will produce an app with a Dock icon.

The motivation is correct — the UI test host must NOT inherit `LSUIElement` — but the fix scope is too broad. `Info.plist` should keep `LSUIElement`, and the UI test target should explicitly override it:

```yaml
# project.yml — RunnerBarUITests settings
INFOPLIST_KEY_LSUIElement: NO
```

That keeps production `swift build` correct and still gives the test host `.regular` activation policy.

eonist:
## Code Review

The PR fully delivers on #936's goal and correctly improves upon the original spec based on hard-won Xcode 26 realities. Every meaningful deviation from the issue is justified.

---

### Deviations from Issue #936 (All Justified)

**`app.windows` → `app.staticTexts`** — the issue explicitly says *"use `app.windows`, not `app.popovers`"* but correctly abandons `app.windows` entirely. For a `[.borderless, .nonactivatingPanel]` NSPanel, `app.windows` is permanently empty regardless of window level. Querying `app.staticTexts["WORKFLOWS"]` directly is the right fix.

**`XCUIApplication()` → `XCUIApplication(bundleIdentifier:)`** — the issue uses default init. On Xcode 26 this resolves the wrong path. Bundle ID init is correct.

**`dependencies: - target: RunnerBar` removed** — the issue spec includes this in `project.yml`. On Xcode 26 this triggers auto-injection of a broken `XCTTargetAppPath`. Removing it and compensating with a separate `xcodebuild build` step (sharing `-derivedDataPath`) is the right call.

**`UI_TESTING` guard placed in `AppDelegate+PanelSetup.swift`** — the issue suggested `main.swift`. Placing it at the actual `setupCombineSubscriptions()` call site is correct — that's where the network/keychain subscriptions are wired up.

**`LSUIElement` moved from `Info.plist` to `project.yml`** — necessary so the UI test target doesn't inherit it, which would make `setActivationPolicy(.regular)` ineffective.

---

### Genuine Code Issues

**`NSApp.activate(ignoringOtherApps: true)` is deprecated (macOS 14+).** In `applicationWillFinishLaunching`, after the `UI_TESTING` guard. Prefer `NSRunningApplication.current.activate(options: [])` or just `NSApp.activate()`. Not a blocker, but worth fixing.

**Duplicate nil-fallback rect logic.** The `button.window` nil-fallback (centred rect on main screen) is copy-pasted identically into both `openPanel()` and `resizeAndRepositionPanel()`. Should be extracted to a small private helper (e.g. `statusItemFallbackRect()`) to prevent drift.

**`ui-tests.yml` triggers on every push across all branches.**
```yaml
on:
  push:
    branches: ['**']
```
The job is gated by `[run ui tests]` in the commit message so it won't *run* unexpectedly, but the workflow is still *triggered* on every push, adding noise to the Actions tab. The original issue spec had `branches: [main]` only. Given this runs on a shared self-hosted Mac, dropping the blanket `push` and keeping only `pull_request` + `workflow_dispatch` would match the intent and reduce clutter.

**`Thread.sleep(forTimeInterval: 0.5)` in `setUp` is fragile.** On a loaded CI machine 0.5s may not be enough for the stale process to fully exit. A tight poll checking `stale.state == .notRunning` would be more reliable.

---

### Nits

- `README.md` diff adds only two trailing spaces on a blank line — pure noise, safe to drop.
- `addRunnerButton()` / `addScopeButton()` probe the explicit identifier with `waitForExistence(timeout: 0.5)`. Since Settings view is already confirmed open at that point, `exists` (no wait) is sufficient and saves ~0.5s per test.

---

### Bottom Line

Implementation is correct. The two most actionable items before merge: **extract the duplicate nil-fallback rect logic** and **tighten the workflow trigger** (drop `push: branches: ['**']`). Everything else is polish.

.github/workflows/ui-tests.yml
```diff
@@ -0,0 +1,111 @@
+name: UI Tests
+
+on:
+  push:
+    branches: ['**']
+  pull_request:
+  workflow_dispatch:  # allows manual trigger from the GitHub Actions UI
+
```
eonist, line 6: Drop the `push` trigger. Every branch push creates a workflow run entry in the Actions UI even though the `if:` condition prevents actual execution. `pull_request` + `workflow_dispatch` is sufficient — and consistent with what `UITEST-LESSONS.md` recommends.

```yaml
on:
  pull_request:
  workflow_dispatch:
```

Sources/RunnerBar/App/AppDelegate+Navigation.swift
```diff
@@ -87,7 +87,15 @@ extension AppDelegate {
     /// Performs the settingsView operation.
     func settingsView() -> AnyView {
         savedNavState = .settings
-        makeKeyForTextInput()
+        // Skip makeKeyForTextInput during UI tests.
+        // makeKeyAndOrderFront repositions the panel (AppKit recalculates frame
+        // for the new key window), which invalidates the AX coordinate snapshot
+        // XCTest has already taken — every subsequent click lands in the wrong place.
+        // In UI tests we never type into text fields, so key status is not needed.
+        // ❌ NEVER remove this guard — same pattern as the event monitor skip in openPanel().
```
eonist, line 95: Same `UI_TESTING` guard copy-pasted three times across `settingsView()`, `runnerDetailView()`, and `scopeDetailView()`. Extract once:

```swift
private func makeKeyForTextInputIfNeeded() {
    guard ProcessInfo.processInfo.environment["UI_TESTING"] == nil else { return }
    makeKeyForTextInput()
}
```

Then all three call sites become a single `makeKeyForTextInputIfNeeded()`.

eonist:
## Review: PR #947 — XcodeGen UITest setup

This PR fully delivers on #936 and correctly fixes several wrong assumptions in the original issue spec. The implementation is solid.

---

### Deviations from the issue spec (all correct)

The issue proposed things that turned out to be wrong — the PR fixes all of them:

- **`app.windows` → `app.staticTexts` directly.** The issue said "use `app.windows`, not `app.popovers`" — both are wrong for a `[.borderless, .nonactivatingPanel]`. Querying content elements directly is correct.
- **`XCUIApplication()` default init → `XCUIApplication(bundleIdentifier:)`** — default init resolves incorrectly on Xcode 26.
- **`UI_TESTING` guard placed in `AppDelegate+PanelSetup.swift`** — correct call site for `RunnerStore.shared.start()`, not `main.swift` as the issue suggested.
- **`dependencies: - target: RunnerBar` removed from `project.yml`** — the issue spec included this, but it triggers the Xcode 26 `XCTTargetAppPath`/`.app`-stripping bug.
- **`LSUIElement` moved from `Info.plist` to build setting** — correct so UI tests see a `.regular` foreground app while production retains the setting via `INFOPLIST_KEY_LSUIElement`.

---

### 🔴 One item to verify: `LSUIElement` and the SPM production build path

The issue noted: *"The app already has `LSUIElement = YES` in `Resources/Info.plist`"*. The PR removes it from `Info.plist` and replaces it with `INFOPLIST_KEY_LSUIElement: YES` in `project.yml`. However, `README.md` shows the production install path is `bash build.sh` → `swift build`, which does **not** go through XcodeGen. In that path, `Info.plist` is used directly — `LSUIElement` would be missing and the app would show a Dock icon in production.

**Check:** does `build.sh` inject `LSUIElement` separately, or embed it via `xcodebuild`? If it's a pure `swift build`, this is a regression worth fixing before merge.

---

### 🟡 Minor: `Thread.sleep(forTimeInterval: 0.5)` in `setUp`

```swift
stale.terminate()
Thread.sleep(forTimeInterval: 0.5)
```

Fixed sleep is fine locally but brittle under CI load. A short polling loop on `stale.state == .notRunning` would be more robust.

---

### 🟡 Minor: `NSScreen.screens[0]` not guarded

```swift
let screen = NSScreen.main ?? NSScreen.screens[0]
```

`NSScreen.screens` can be empty on a headless machine. `NSScreen.screens.first` with a guard/fallback is safer.

---

### Everything else looks good ✅

- Two-step `xcodebuild build` then `xcodebuild test` with shared `-derivedDataPath` — correct Xcode 26 solution
- `applicationWillFinishLaunching` for `setActivationPolicy(.regular)` — correct timing before XCTest handshake
- Global event monitor skipped during `UI_TESTING` — necessary and well-documented
- `coordinate(withNormalizedOffset:).click()` for all button taps — correct for `nonActivatingPanel`
- Opt-in `[run ui tests]` flag with `workflow_dispatch` fallback — good CI hygiene
- `UITEST-LESSONS.md` is excellent documentation of hard-won findings

The only item worth blocking on is the `LSUIElement` question — everything else reflects solid, well-debugged work.

Tests/RunnerBarUITests/RunnerBarUITests.swift
```diff
@@ -0,0 +1,182 @@
...
+        tapButton(app.buttons["Cancel"])
+        XCTAssertTrue(app.staticTexts["Active local runners"].waitForExistence(timeout: 3),
+                      "Back in Settings after Cancel")
+
+        // ── 4. Back to main ───────────────────────────────────────────
+        tapButton(app.buttons["Settings"])
+        XCTAssertTrue(
+            app.staticTexts["WORKFLOWS"].waitForExistence(timeout: 5),
+            "WORKFLOWS must reappear after back navigation"
+        )
```
eonist, line 176: `XCTAssertFalse(app.staticTexts["Active local runners"].exists)` queries AX state synchronously with no wait. If SwiftUI takes more than one runloop cycle to remove the view this fires against stale state and produces a false pass.

```swift
XCTAssertTrue(
    app.staticTexts["Active local runners"].waitForNonExistence(timeout: 3),
    "Settings content must not be visible on main view"
)
```

eonist (commented):
All four phases are delivered and the core architecture is solid — ephemeral xcodeproj, opt-in `[run ui tests]` trigger, two-step build/test split, and the `UI_TESTING` guard pattern are all correct.

Four issues to fix before merge:
1. `LSUIElement` removed from `Info.plist` breaks `swift build` / `build.sh` (see inline comment)
2. `push: branches: ['**']` trigger should be dropped (see inline comment)
3. `UI_TESTING` guard duplicated three times in `AppDelegate+Navigation.swift` (see inline comment)
4. `XCTAssertFalse(...exists)` needs `waitForNonExistence` (see inline comment)

Issue #1001, opened 2026-05-28
Refactor runner detail editing into a transactional popover with OK/Cancel


eonist:
## Problem

The current runner detail flow is a navigation-pushed detail view with multiple inline Save buttons and one auto-saving toggle. This creates poor UX because editing is not scoped as a single transaction, and backing out of the screen makes the persistence behavior unclear.

It also duplicates a pattern we already solve better in the add-runner flow, which uses a popover with explicit confirmation. The inline Save buttons are also an anti-pattern on Apple platforms — they create mode confusion and put unnecessary cognitive burden on the user.

Closes #988

## Current state

`RunnerDetailView` currently:
- Is pushed through app navigation instead of presented as a popover
- Uses multiple independent save paths for labels, work folder, auto-update, and proxy settings
- Writes to multiple backends: GitHub labels API, `.runner` JSON, `.proxy` / `.proxycredentials`
- Has no draft model, so there is no clean commit / cancel boundary
- Includes several Save buttons that are poor UX and should be removed
- Has an auto-saving toggle that fires persistence immediately on change

## Goal

Move runner editing into a popover-style flow with:
- A local draft/edit model buffering all changes
- Explicit `OK` and `Cancel` — no inline Save buttons
- No auto-save while editing
- A single commit path that persists changes in a controlled order
- Error surfacing at the popover level, not per-field
- Consistent behavior with the existing add-runner popover

## Implementation Plan

- Phase 1: Add `RunnerEditDraft` model to buffer all editable values
- Phase 2: Build `RunnerDetailPopover` UI, remove per-field save controls
- Phase 3: Replace field-level save paths with a single transactional `commit()` flow
- Phase 4: Wire `SettingsView` to present runner editing as a popover
- Phase 5: Remove obsolete navigation and save-state plumbing
- Phase 6: Update UI tests for the new popover-based flow

## Key implementation notes

- Labels should be committed before local file writes — it's the networked path and most likely to fail
- Outside-click popover dismissal should behave as Cancel (no writes)
- Dirty-state detection (`isDirty`) to determine whether dismissal confirmation is needed
- Existing add-runner popover should guide presentation style and button layout
- Partial failure: keep popover open, surface a single error message, do not partially persist

PR #1009, opened 2026-05-28
feat: replace RunnerDetailView push-nav with RunnerDetailPopover (#1001)

### LINKED ISSUE #988: ⭐ Move runner detail view into a popover style view similar to how add runner popover view works
rational:

since we edit data. the user needs to implicitly agree to the changes

a popover view forces the user to eigther tap ok or cancel. a detail view doesnt. you can  press back. then what happens?

it also avoids all the really bad save buttons we have in the runner detail view. which are horrible for ux

eonist:
## Summary

Replaces the full-panel push-navigation flow for runner editing (`RunnerDetailView` → back → `SettingsView`) with an inline popover (`RunnerDetailPopover`) anchored to the runner row in Settings. The panel never navigates away; editing happens in-place.

Closes #988 · Part of #1001

---

### Phase 1 — Model + write logic
- `RunnerEditDraft` value type: snapshots all editable runner fields from a `RunnerModel`
- `commitRunnerEdit(runner:draft:original:completion:)`: async, dirty-field-skipping write function; calls completion with `.success` or `.failure([String])`

### Phase 2 — `RunnerDetailPopover`
- New `SwiftUI` popover view with **Runner Info** and **Configuration** sections
- Labels, work folder, and environment fields with inline validation
- Footer with **Cancel** / **OK** buttons; OK disabled while committing
- `onCommit(draft:)` and `onCancel` callbacks; no internal writes

### Phase 3 — `RunnerDetailView` neutered
- Old push-nav view body replaced with a compile-time `fatalError` stub
- File retained so the module compiles; can be deleted in a follow-up

### Phase 4 — `SettingsView` wiring
- `@State private var editingRunner: RunnerModel?` drives `.popover(item:)`
- Runner row tap sets `editingRunner`; Cancel/OK wired through `commitRunnerEdit`
- `isCommitting` guard prevents double-taps; `commitError` surfaces write failures in the popover footer
- `localRunnerStore.refresh()` called on successful commit

### Phase 5 — Nav plumbing removed
- `NavState.runnerDetail(RunnerModel)` case deleted
- `runnerDetailView(runner:)` factory deleted from `AppDelegate+Navigation`
- `onSelectRunner` callback replaced with documented no-op pending param removal

### Phase 6 — Tests + final cleanup
- `onSelectRunner` parameter removed from `SettingsView` and its call site
- **`testRunnerDetailPopoverFlow`**: tap row → popover appears (Runner Info + Configuration) → no Save buttons → Cancel dismisses → Settings still visible
- **`testRunnerDetailPopoverCancelDiscards`**: type sentinel into Labels → Cancel → re-open → assert sentinel absent
- Both tests skip gracefully when no local runners are installed (CI-safe)

---

## Behaviour contract

| Scenario | Before | After |
|---|---|---|
| Tap runner row | Panel navigates to RunnerDetailView | Popover opens over Settings |
| Cancel | Back button returns to Settings | Popover dismissed, no writes |
| OK (clean) | — | Popover dismissed, `localRunnerStore.refresh()` |
| OK (error) | — | Popover stays open, error shown in footer |
| Double-tap OK | — | Guarded by `isCommitting` flag |

---

## Testing

- `testSettingsNavigationFlow` — unchanged, continues to pass
- `testRunnerDetailPopoverFlow` — new
- `testRunnerDetailPopoverCancelDiscards` — new

___

## **CodeAnt-AI Description**
**Edit runners in a popover from Settings instead of opening a separate detail view**

### What Changed
- Tapping a runner in Settings now opens an inline popover with runner details and editable fields, instead of navigating away to a separate screen
- The popover uses Cancel and OK actions at the bottom; edits stay local until OK is pressed, and Cancel closes the popover without saving
- Saving now updates the changed runner settings in one commit flow, including labels, work folder, auto-update, and proxy settings
- The old runner detail navigation path was removed, and UI tests now cover opening, cancelling, and discarding popover edits

### Impact
`✅ Faster runner editing`
`✅ Fewer accidental runner-setting changes`
`✅ Clearer save and cancel flow`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **New Features**
  * Runner editing is available as a modal popover in Settings (edit labels, work folder, auto-update, proxy) with copy-to-clipboard and info display.
  * Adds an on-commit workflow that persists edits, reports aggregated success/errors, and updates local state.

* **Refactor**
  * Simplified runner navigation flow to use a popover-based editor.

* **Tests**
  * UI tests added for popover open/close and cancel/discard behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/eoncode/runner-bar/pull/1009?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Sources/RunnerBar/Views/Settings/SettingsView.swift
```diff
@@ -115,6 +122,40 @@ struct SettingsView: View {
...
+            runnerEditingPopover(runner: runner)
+        }
+    }
+
+    // MARK: - Runner editing popover (#1001)
+
+    /// Builds the `RunnerDetailPopover` with commit/cancel wiring.
+    @ViewBuilder
+    private func runnerEditingPopover(runner: RunnerModel) -> some View {
+        RunnerDetailPopover(
```
eonist, line 134: **Bug: `original` is seeded from the model, not from the disk snapshot.**

```swift
let original = RunnerEditDraft(runner: runner)
commitRunnerEdit(runner: runner, draft: draft, original: original) { ... }
```

`RunnerEditDraft.init(runner:)` sets `autoUpdate = true`, `proxyUrl = ""`, etc. as defaults — it does **not** read disk. So if a runner has `disableUpdate = true` on disk and the user opens the popover without changing anything, `draft.autoUpdate` (loaded from disk via `loadDisplayFields`) will be `false`, but `original.autoUpdate` will be `true` — triggering a spurious `.runner` JSON rewrite on every OK press.

`RunnerDetailPopover` already captures `originalDraft` after the disk load. The fix is to thread that snapshot out via the `onCommit` callback, or pass `original` in alongside `draft`:

```swift
// Option A — pass original through onCommit
let onCommit: (_ draft: RunnerEditDraft, _ original: RunnerEditDraft) -> Void
// then in SettingsView:
commitRunnerEdit(runner: runner, draft: draft, original: original) { ... }
```

Sources/RunnerBar/Views/Settings/SettingsView.swift
```diff
@@ -115,6 +122,40 @@ struct SettingsView: View {
             )
```
eonist, line 120: **`commitError` is set but never read.** The state is declared and written in the `failure` branch, but there's no UI that reads it — no `Text(commitError ?? "")` in `runnerEditingPopover` or anywhere else. Either wire it into `RunnerDetailPopover` (e.g. pass as a binding or show it in the popover footer), or remove the property until the error UI is built.

Sources/RunnerBar/App/AppDelegate+Navigation.swift
```diff
@@ -87,12 +90,6 @@ extension AppDelegate {
     /// Performs the settingsView operation.
```
eonist, line 92: **Deleted a load-bearing `❌ NEVER` guard comment.** The block below it still exists:

```swift
if ProcessInfo.processInfo.environment["UI_TESTING"] == nil {
    makeKeyForTextInput()
}
```

But the 6-line comment explaining *why* the guard must never be removed was deleted. This is the exact kind of comment the codebase uses to prevent future regressions (same pattern as the event monitor skip note in `openPanel()`). Please restore it.

Sources/RunnerBar/Models/RunnerEditCommit.swift
```diff
@@ -0,0 +1,201 @@
...
+    runner: RunnerModel,
+    draft: RunnerEditDraft,
+    original: RunnerEditDraft,
+    completion: @escaping @Sendable (CommitResult) -> Void
+) {
+    DispatchQueue.global(qos: .userInitiated).async {
+        var errors: [String] = []
+
+        // MARK: Step 1 — Labels (GitHub API)
+        let labelsChanged = draft.parsedLabels != original.parsedLabels
```
eonist, line 50: **Labels-changed error is non-fatal but the commit continues to write local files.** The comment says "Non-fatal for local writes; continue" but that's debatable. If the GitHub API call fails because `agentId`/`gitHubUrl` are missing, the labels the user typed will be silently discarded and the local files will still be mutated. The user gets no feedback that half the commit failed.

At minimum this branch should append the `msg` to `errors` (it already does), but consider making it a hard abort here too — consistent with the `result == nil` path above — so the caller always sees a clean success/failure.

Sources/RunnerBar/Views/Settings/RunnerDetailPopover.swift
```diff
@@ -0,0 +1,309 @@
...
+    private var infoSection: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            sectionHeader("Runner Info")
+            infoCard {
+                if let url = runner.gitHubUrl {
+                    infoRow(label: "GitHub URL", value: url, copyable: true)
+                    Divider().padding(.leading, RBSpacing.md)
+                }
+                infoRow(label: "Work folder", value: runner.workFolder ?? "_work")
+                Divider().padding(.leading, RBSpacing.md)
```
eonist, line 148: **`infoSection` reads `runner.workFolder` from the live model, not from `draft`.** The Config section has an editable `draft.workFolder` field, but the read-only Info row above it shows `runner.workFolder ?? "_work"`. If the on-disk value differs from the model's value (which is possible if `RunnerModel` was populated from LaunchAgent plists rather than `.runner` JSON), the two rows will show different values for the same field. Consider using `draft.workFolder` here once the draft is loaded, or removing the duplicate.

Tests/RunnerBarUITests/RunnerBarUITests.swift
```diff
@@ -185,51 +217,113 @@ final class RunnerBarUITests: XCTestCase {
...
+        // ── 2. Verify no inline Save buttons exist ────────────────────────
+        XCTAssertFalse(
+            app.buttons["Save"].exists,
+            "No inline Save buttons must exist in RunnerDetailPopover"
+        )
+
+        // ── 3. Verify Cancel and OK are present ─────────────────────────
+        XCTAssertTrue(app.buttons["Cancel"].exists, "Cancel button must exist in popover")
+        XCTAssertTrue(app.buttons["OK"].exists, "OK button must exist in popover")
+
```
eonist, line 253: **`XCTAssertFalse(... waitForExistence(timeout: 1))` is flaky on slow CI.** `waitForExistence(timeout:)` returns `true` when the element *appears* within the timeout; `XCTAssertFalse` of that is correct intent (assert it does NOT appear), but a 1s timeout is very tight — popover dismissal animations on a slow CI machine can take longer. Consider bumping to 2–3s, or use a `doesNotExist` check after a brief `sleep`.

eonist (commented):
Conflict resolution looks correct — the three affected files (`NavState`, `AppDelegate+Navigation`, `SettingsView`) are cleanly merged and the new files (`RunnerDetailPopover`, `RunnerEditDraft`, `RunnerEditCommit`) read well. A few issues to address before merge:

**Must fix**
1. `original` draft in `SettingsView.runnerEditingPopover` is seeded from the model (not disk), so autoUpdate/proxy comparisons will be wrong and trigger spurious writes on every OK.
2. `commitError` state is written but never displayed — either wire it into the popover footer or remove it for now.
3. Deleted the load-bearing `❌ NEVER` guard comment in `settingsView()` — restore it.

**Nit / consider**
4. `RunnerEditCommit` labels non-fatal path silently continues local writes — user gets no feedback that labels failed.
5. `infoSection` in `RunnerDetailPopover` reads `runner.workFolder` but `configSection` edits `draft.workFolder` — potential stale display mismatch.
6. UI test `XCTAssertFalse(waitForExistence(timeout: 1))` may be flaky on slow CI — bump timeout.

eonist:
## Code Review — `feat/runner-detail-popover` (#1001)

Overall this is a well-structured, well-scoped refactor. The separation of concerns between the draft model, the commit logic, and the view layer is clean, and the phased approach in the PR description makes the intent very easy to follow. A few observations and suggestions below.

---

### ✅ What's done well

- **`RunnerEditDraft` as a value type** — keeping all editable state in a `struct` that is only committed on OK is a solid pattern. The `isDirty(comparedTo:)` / `Equatable` approach is idiomatic Swift.
- **`commitRunnerEdit` is free-standing** — keeping the write logic out of the view and out of any specific model type makes it easy to unit-test in isolation.
- **Double-tap guard (`isCommitting`)** — good defensive UX.
- **CI-safe UI tests** — skipping gracefully when no local runners are present is the right call for a headless CI environment.
- **Nav plumbing cleanly removed** — `NavState.runnerDetail`, `runnerDetailView(runner:)`, and `onSelectRunner` are all gone with no dangling references.

---

### 🟡 Observations / suggestions

**1. `originalDraft` is seeded twice**

In `RunnerDetailPopover.init`, `originalDraft` is set from the pre-disk `RunnerEditDraft(runner:)`. Then in `loadDisplayFields()` (called from `onAppear`) the draft is loaded from disk and `originalDraft = draft` is updated again.

This is correct, but it means the popover will show a brief dirty state if SwiftUI reads `isDirty` between `init` and `onAppear`. It's probably not observable today, but worth a comment explaining the two-step seeding intentionally.

**2. `commitRunnerEdit` uses `original: RunnerEditDraft(runner:)` — not the on-disk snapshot**

In `SettingsView.runnerEditingPopover`, the `original` passed to `commitRunnerEdit` is freshly constructed from the model:

```swift
let original = RunnerEditDraft(runner: runner)
```

But the popover itself calls `draft.load(installPath:)` on appear, which reads on-disk values (proxy URL, autoUpdate) into the draft. If the user opens the popover and immediately taps OK without changing anything, the `original` in `commitRunnerEdit` won't match the post-disk-load `draft`, so fields like `proxyUrl` could be considered "changed" when they haven't been. The `original` passed to commit should be the same snapshot that was captured after `load()` — i.e. `originalDraft` from the popover state, not a freshly seeded model draft.

**3. `RunnerDetailView` stub (Phase 3)**

The `fatalError` stub is a reasonable compile-time safety net, but it would benefit from a TODO comment pointing to the follow-up issue/ticket for deletion, so it doesn't become invisible debt.

**4. `patchRunnerJSONMulti` — mixed `Any` values**

Using `[String: Any]` for the patches dictionary is pragmatic, but it bypasses Swift's type system. A small typed struct or two separate patch functions (`patchRunnerJSONBool`, `patchRunnerJSONString`) would make the call sites more expressive and prevent accidentally passing a wrong type silently.

**5. `writeProxyFiles` — proxy URL format not validated**

The proxy URL is written to disk as-is. A minimal `URL(string:)` validity check (or at least a `http://`/`https://` prefix check) before writing would surface user typos early rather than silently writing an unusable `.proxy` file.

**6. UI test: `testRunnerDetailPopoverCancelDiscards` — `XCTAssertFalse` on `waitForExistence`**

```swift
XCTAssertFalse(
    app.staticTexts["Runner Info"].waitForExistence(timeout: 1),
    ...
)
```

Using `waitForExistence` with a 1-second timeout in a negative assertion means the test always waits a full second. A `XCTAssertFalse(app.staticTexts["Runner Info"].exists)` after a short `sleep` or after the Cancel tap's implicit sync point would be faster and equivalent.

---

### 🔴 Minor nit

In `RunnerEditDraft.init`, the label filter uses a hardcoded list of strings (`"self-hosted"`, `"x64"`, `"arm64"`, etc.). These are duplicated from wherever the original `RunnerDetailView` filtered them. If there's an existing constant or helper for system labels elsewhere in the codebase, it'd be worth reusing it to keep the filter logic in one place.

---

**Summary:** The PR is in good shape and the architectural direction is correct. The most actionable issue is point **#2** (mismatched `original` in `commitRunnerEdit`), which could cause spurious writes on a clean OK. Everything else is low-risk polish. Happy to approve after that's addressed.

eonist:
## Code Review — feat: replace RunnerDetailView push-nav with RunnerDetailPopover (#1001)

Overall this is a well-structured PR with a clean, phased approach. The intent is clear, the tests are pragmatic, and the removal of dead nav plumbing is welcome. A few things worth discussing before merging:

---

### ✅ What's good

- **`RunnerEditDraft` as a value type** is exactly right. Keeping all in-flight edits in an isolated struct, seeded on appear and discarded on cancel, prevents any accidental side-effects and is easy to test.
- **`commitRunnerEdit` abort-on-labels-failure** is a good safety net — failing the GitHub API call before touching disk prevents a half-applied state.
- **UI tests skip gracefully** rather than failing hard when no local runner is installed. Solid CI hygiene.
- **Nav plumbing cleanup** (removing `NavState.runnerDetail`, `runnerDetailView(runner:)`, and `onSelectRunner`) is thorough and correct.

---

### ⚠️ Issues & suggestions

#### 1. `original` is re-derived from the model, not the loaded draft

In `SettingsView.runnerEditingPopover`:
```swift
let original = RunnerEditDraft(runner: runner)
```
This seeds `original` from the `RunnerModel` alone, but `RunnerDetailPopover.loadDisplayFields()` calls `draft.load(installPath:)` and then snapshots `originalDraft = draft` — meaning the popover's own dirty-check uses the on-disk values, while the commit path uses a model-only snapshot. If `autoUpdate` or proxy fields differ between model and disk, the commit may write fields the user never touched.

**Suggestion:** Pass `originalDraft` back out of the popover via an extra callback or a `@Binding`, or move `original` capture into the `onCommit` closure parameter (e.g. `onCommit: (draft: RunnerEditDraft, original: RunnerEditDraft) -> Void`).

---

#### 2. `commitError` is never surfaced to the user

`commitError` is set on failure, but `RunnerDetailPopover` has no `commitError` parameter and no way to display it. The PR description says "Popover stays open, error shown in footer" — but the footer only shows Cancel / OK buttons. The error display path appears to be missing.

---

#### 3. Proxy password stored/read as plain text

`.proxycredentials` is written as `"\(user)\n\(password)"` and read back by splitting on `\n`. Storing a plain-text password in a file in the runner install directory is the same as the upstream runner behaviour, so this is an existing pattern — but it's worth a comment acknowledging the trade-off, and ideally a note in the PR that this is intentional parity.

---

#### 4. `.runner` JSON parsed in two places

`loadDisplayFields()` in `RunnerDetailPopover` reads and parses the `.runner` JSON file directly to fill `displayOsArch` / `displayVersion`, duplicating the logic already in `RunnerEditDraft.loadRunnerJSON`. This creates two file reads on appear and two JSON parse paths to keep in sync.

**Suggestion:** Expose `platform`, `platformArchitecture`, and `agentVersion` as computed read-only properties on `RunnerEditDraft` or read them once in `load(installPath:)`, then pass them into the popover as simple `let` values.

---

#### 5. `fatalError` stub in `RunnerDetailView` is a landmine

Leaving a `fatalError` in the view body (Phase 3) means any code path that accidentally instantiates `RunnerDetailView` will crash in production. The PR description notes "can be deleted in a follow-up" — that follow-up should be tracked as an issue immediately. Consider at minimum wrapping it in `#if DEBUG` or just deleting the file in this PR since `RunnerDetailView` is no longer referenced.

---

#### 6. Minor: trailing whitespace changes in test file

Several test file lines have changed only trailing whitespace (e.g. lines 62–63, 111, 126, 130). These are noise in the diff and can make `git blame` less useful — worth a cleanup pass.

---

### Summary

| | |
|---|---|
| Logic correctness | ✅ Solid overall |
| `original` draft mismatch | ⚠️ Should fix before merge |
| `commitError` not shown | ⚠️ Missing feature / regression from description |
| Proxy plaintext | ℹ️ Document intentional |
| Duplicate JSON reads | ℹ️ Nice-to-fix |
| `fatalError` stub | ⚠️ Track follow-up issue |

Happy to approve once items 1 and 2 are addressed (or explicitly deferred with tracking issues).

eonist:
## Code Review — feat: replace RunnerDetailView push-nav with RunnerDetailPopover (#1001)

**Overall: ✅ Well-structured refactor. Clean separation of concerns, good UX improvement. A few points worth addressing.**

---

### ✅ What works well

- **`RunnerEditDraft` as a value type** is the right call — it makes cancel/discard trivially correct and eliminates the risk of accidental live mutations.
- **`commitRunnerEdit` step order** (labels API first, abort on failure before touching disk) is a solid transactional design. The "abort early" pattern on API failure prevents a split state where the UI shows updated labels but disk hasn't changed.
- **`isCommitting` guard** against double-tap on OK is correct and necessary.
- **CI-safe test skips** (graceful `return`/`XCTSkip` when no local runners are installed) are the right approach for popover-dependent tests.
- Clean removal of `NavState.runnerDetail` and `runnerDetailView(runner:)` — no dead nav paths left.

---

### ⚠️ Issues / Suggestions

**1. `originalDraft` is reconstructed from the model in `onCommit`, not from the disk-loaded snapshot**

In `SettingsView.runnerEditingPopover`:
```swift
let original = RunnerEditDraft(runner: runner) // ← only model values, no disk
commitRunnerEdit(runner: runner, draft: draft, original: original) { ... }
```
But `RunnerDetailPopover.loadDisplayFields()` loads disk values (auto-update, proxy) into `draft` **and** snapshots them into `originalDraft`. The `original` passed to `commitRunnerEdit` from `SettingsView` won't reflect the disk-loaded proxy/auto-update values, so those fields will always appear "dirty" and trigger unnecessary writes.

**Fix:** Pass `original` through a callback from the popover, or expose it via a `@Binding`.

---

**2. `RunnerDetailView` retained as a `fatalError` stub**

The PR description acknowledges this, but leaving a `fatalError` stub in production code is a risk if any non-test path accidentally instantiates it. Consider wrapping in `#if DEBUG` or adding a clear `// TODO: delete in follow-up` with an issue reference so it doesn't get forgotten.

---

**3. `loadDisplayFields()` reads `.runner` JSON twice**

`RunnerEditDraft.load(installPath:)` (called from `loadDisplayFields`) already parses `.runner` JSON internally in `loadRunnerJSON`. Then `loadDisplayFields` parses the same file again to extract `platform`/`agentVersion`. This is two `Data(contentsOf:)` + `JSONSerialization` calls on appear. Minor, but consider returning the parsed JSON from `load()` or caching it.

---

**4. Proxy password stored/written as plain text**

`.proxycredentials` is written with `"user\npassword"` in plain text to disk. This is consistent with how the GitHub Actions runner itself stores credentials, so it may be intentional — but worth a comment acknowledging this is matching the runner's own convention, so future contributors don't flag it as a security concern.

---

**5. `firstRunnerRow()` in UI tests relies on `chevron.right` image presence**

This is a heuristic. If the row design changes to use a different disclosure indicator, the test helper silently finds no rows and skips rather than failing. Consider adding an explicit `accessibilityIdentifier` (e.g. `"runnerRow"`) to the runner row button for a more stable anchor.

---

### 🔍 Minor nits

- `swiftlint:disable missing_docs` at file level in `RunnerEditCommit.swift` and `RunnerDetailPopover.swift` — since there's already doc comments on the main public types, it may be worth enabling docs selectively on private helpers rather than disabling globally.
- The `configSection` in `RunnerDetailPopover` has inline proxy `HStack` rows duplicated rather than using `configRow()`. Consider extracting them into `configRow` calls for consistency.

---

**Summary:** The core architecture is sound. The `originalDraft` mismatch in point 1 is the most functional concern and worth fixing before merge. The rest are improvements for future hygiene.

PR #1045, opened 2026-05-31
fix(#1041): include completed-only SHAs in fetchActionGroups

### LINKED ISSUE #1041: ⭐ It seems some jobs never conclude. they just stick around and linger
look at the image 

Image

we need to figure out whats wrong. because in github all workflows has concluded.

### LINKED ISSUE #1044: 🐛 Fix: completed-only groups silently dropped from fetchActionGroups, causing jobs to linger as in-progress
## Problem

Workflows that conclude between two poll cycles never appear in the `liveGroups` or `doneGroups` arrays inside `buildGroupState`. They are therefore never frozen into the cache, and any entry from a previous poll that had the same group ID stays visible in the UI indefinitely as an in-progress job — even though GitHub shows the workflow as completed.

## Root cause

`fetchActionGroups` in `WorkflowActionGroupFetch.swift` fetches three batches of runs:

1. `status=in_progress`
2. `status=queued`
3. `status=completed` (to supplement the above)

Step 3 only appends completed runs to SHAs that **already exist** in `bySha` from steps 1–2:

```swift
// ❌ BEFORE — completed-only SHAs are silently dropped
if bySha[run.headSha] != nil { bySha[run.headSha]!.append(run) }
```

If all sibling runs for a SHA finish between two polls, that SHA has no entry from steps 1 or 2, so the `if` guard causes it to be completely ignored. The group vanishes from `allFetched`, so `freezeVanishedGroups` in `PollResultBuilder` is supposed to catch it — but it only fires correctly if the group was already in `snapPrevGroups`. If the app was launched or the store was reset after the run completed, the group is never seen at all and no cleanup fires. The result: stale in-progress UI entries that never resolve.

## Fix

Remove the guard so completed runs are inserted for **all** SHAs, not just ones already seen:

```swift
// ✅ AFTER — completed SHAs are always included
bySha[run.headSha, default: []].append(run)
```

`seenIDs` already deduplicates runs across all three fetches, so there is no risk of double-counting. `groupCacheLimit = 30` trims overflow at the display layer.

## Test coverage

A new unit test `testFetchActionGroupsCompletedOnlyGroupsIncluded` is added to `PollResultBuilder`'s test section in `RunnerBarCoreTests.swift`, verifying that `evictFreshShas` + `buildGroupState` correctly routes a completed-only group into the cache rather than leaving it as a lingering live entry.

Fixes #1041

eonist:
## What

Fixes #1041 and closes #1044.

Groups that completed between two polls were silently dropped from `fetchActionGroups`, leaving stale in-progress UI rows that never resolved — even though GitHub showed the workflow as done.

## Why it broke

`fetchActionGroups` fetches runs in three passes: `in_progress`, `queued`, then `completed`. The completed pass used a guard that only appended runs for SHAs **already present** from the first two passes:

```swift
// ❌ Before
if bySha[run.headSha] != nil { bySha[run.headSha]!.append(run) }
```

If all sibling runs for a SHA finished before the poll cycle hit the `in_progress`/`queued` requests, that SHA had no prior `bySha` entry — so the completed runs were dropped entirely. The group never appeared in `allFetched`, `freezeVanishedGroups` had nothing to freeze, and the old live entry from the previous poll lingered indefinitely.

## Fix

One-line change: replace the conditional append with an unconditional default-valued subscript insert:

```swift
// ✅ After
bySha[run.headSha, default: []].append(run)
```

`seenIDs` deduplication across the three passes already prevents double-counting. `groupCacheLimit = 30` trims overflow at display time.

## Tests

Added `PollResultBuilderGroupStateTests` (4 tests) to `RunnerBarCoreTests.swift`:

| Test | What it covers |
|---|---|
| `testCompletedOnlyGroupIsRoutedToCacheNotLive` | Core regression: completed group goes to cache, not live display |
| `testInProgressGroupAppearsLiveInDisplay` | Guard: live groups still appear as non-dimmed rows |
| `testFireFailureHookCalledOnceForNewCompletedGroup` | Notification hook fires exactly once for a new completed group |
| `testFireFailureHookNotCalledForAlreadyCachedGroup` | No duplicate hook fires for already-cached groups |

## CI checklist

- [x] No new SwiftLint violations — change is a single-expression swap inside an existing `// swiftlint:disable missing_docs` block; no new public API introduced
- [x] Periphery clean — no new public symbols; existing `fetchActionGroups` signature unchanged (`retain_public: true` covers it)
- [x] `swift test --filter RunnerBarCoreTests` — all existing tests preserved, 4 new tests added

___

## **CodeAnt-AI Description**
**Keep completed workflow groups from disappearing between polls**

### What Changed
- Completed workflow groups are now kept even when no earlier live row exists for the same SHA, so finished runs no longer vanish from the results.
- Groups that finish between polls now move into the cached list instead of leaving a stale in-progress row on screen.
- Added coverage for completed-only groups, live in-progress groups, and one-time failure notifications for newly completed groups.

### Impact
`✅ Fewer stuck in-progress workflow rows`
`✅ Completed runs stay visible after they finish`
`✅ More reliable workflow updates during polling`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Release Notes

* **Bug Fixes**
  * Resolved issue where workflow failure notifications could fire multiple times when cache entries are cleared during normal operations.
  * Enhanced tracking of completed workflow groups to prevent redundant failure hook notifications across polling cycles.
  * Improved handling of workflow action groups to maintain consistent state and prevent duplicate notifications.


eonist:
## Deep Review

### The Bug (Issue #1041)

The original report showed workflow runs that were done on GitHub but stuck as live/in-progress rows in the menu bar UI. That's a classic symptom of a fetch pass that silently dropped data — the UI kept showing the old state because the new completed state never made it through the pipeline.

The root cause is correctly identified: the three-pass fetch (`in_progress` → `queued` → `completed`) used a conditional append for the `completed` pass that required the SHA to already exist from one of the first two passes. Any workflow that completed *between* two poll cycles had no prior `bySha` entry, so the completed run was discarded before `freezeVanishedGroups` could ever see it.

---

### The Fix

The production change is minimal and correct:

```swift
// Before
if bySha[run.headSha] != nil { bySha[run.headSha]!.append(run) }

// After
bySha[run.headSha, default: []].append(run)
```

The use of `Dictionary`'s `subscript(_:default:)` is the idiomatic Swift solution — it avoids force-unwrapping and seeds a new entry when the key is absent. The deduplication via `seenIDs.insert(run.id).inserted` upstream in the same loop already ensures no double-counting if the same run appears across passes, so removing the nil-guard is safe. The `groupCacheLimit = 30` downstream trims overflow. This logic chain holds up.

**One thing worth questioning:** the fix makes `bySha` potentially accumulate many completed-only groups — not just the one SHA that raced. If the `completed` endpoint returns 100 runs (`per_page=100`) and *all* are for SHAs not yet seen, you now insert up to 100 new entries unconditionally. Previously the guard kept the map tighter. This isn't a correctness bug, but it's a latent memory/performance consideration — especially since `groupCacheLimit = 30` only trims at display time, not at map-build time.

---

### The Tests

The 4 new tests in `PollResultBuilderGroupStateTests` are well-structured, but there are meaningful gaps:

**What's well done:**
- `testCompletedOnlyGroupIsRoutedToCacheNotLive` is the direct regression test — it validates the entire routing path, not just the one-liner. That's the right level.
- `testFireFailureHookCalledOnceForNewCompletedGroup` and `testFireFailureHookNotCalledForAlreadyCachedGroup` together bracket the hook deduplication logic cleanly.

**What's missing:**

1. **The race condition is not tested at the fetch layer.** All four tests operate on `PollResultBuilder.buildGroupState`, which receives an already-assembled `[WorkflowActionGroup]`. None of them test `fetchActionGroups` directly with a simulated `bySha` state where the SHA is absent. The regression could silently reintroduce at the fetch layer without these tests catching it — they'd still pass. A unit test that stubs the three API passes and asserts the SHA-absent completed run ends up in the returned groups would close this gap.

2. **No test for the mixed case.** What happens when a SHA has *both* an `in_progress` run and a `completed` run in the same poll cycle? The `seenIDs` guard should prevent double-counting, but there's no test asserting the merged group looks correct.

3. **`testInProgressGroupAppearsLiveInDisplay` constructs the live group manually** instead of going through `makeGroup`, creating a parallel construction path. If `WorkflowActionGroup`'s initializer signature changes, this test will fail for a different reason than what it's testing.

---

### Comment Quality

The inline comment added above the fix is genuinely good — it explains *why* the guard was wrong, not just *what* changed. That's the right instinct for a subtle timing-dependent bug.

---

### Verdict

The fix is correct, minimal, and well-motivated. The PR description traces the full causal chain from the race condition to the stale UI row, which makes reviewing easy. The main weakness is test coverage: the regression is validated at the wrong layer (`buildGroupState` instead of `fetchActionGroups`), and the mixed-pass case is untested. Suggest adding at least one test that exercises the fetch layer directly before merging, so future refactors of the three-pass logic can't silently reintroduce the drop without test noise.

eonist:
## Review

### The Fix Itself: Correct, But Narrow

The one-line change from `if bySha[run.headSha] != nil { bySha[run.headSha]!.append(run) }` to `bySha[run.headSha, default: []].append(run)` is the right fix for the described race. The old guard was a classic "first pass wins" mistake: completed runs were subordinated to live runs, but a SHA that went from queued → in_progress → completed *within a single poll interval* never got registered in the first two passes, so the completed pass silently dropped it. The `default: []` subscript is idiomatic Swift and avoids the force-unwrap too — strictly better code.

**However**, the fix assumes that `seenIDs` deduplication is correct and complete. The PR description says "seenIDs deduplication across the three passes already prevents double-counting" — but the diff never shows `seenIDs` being initialized or scoped. If `seenIDs` is a `var` that persists across outer call-site loops (e.g. across multiple scopes), you could now get unexpected *inclusion* of runs rather than the previous silent *exclusion*. Worth verifying the lifetime and reset point of `seenIDs` is per-call to `fetchActionGroups`, not per-session.

---

### Test Coverage: Good Structure, One Blind Spot

The four new tests in `PollResultBuilderGroupStateTests` are well-named and test meaningful state transitions. The coverage matrix is sensible:

| Test | What it actually verifies |
|---|---|
| `testCompletedOnlyGroupIsRoutedToCacheNotLive` | Core regression path — new SHA, completed, no prior live row |
| `testInProgressGroupAppearsLiveInDisplay` | Sanity guard — live groups not broken by the change |
| `testFireFailureHookCalledOnceForNewCompletedGroup` | Side-effect correctness |
| `testFireFailureHookNotCalledForAlreadyCachedGroup` | Idempotency of the hook |

**The blind spot:** there is no test for the *exact* race described in the issue — i.e., a SHA that has *both* prior live entries **and** a new completed run in the same poll cycle. The fix changes behavior for the "completed-only" case, but the original passing case (SHA already in `bySha`) now goes through `default: []` + append too. That path is semantically identical, but a test that explicitly starts with an in-progress entry for the same SHA and then flips to completed would give you much higher confidence that deduplication still holds in the combined case.

---

### The `makeGroup` Helper: Fragile Construction

The `makeGroup` helper builds a `WorkflowActionGroup` using `status` to derive a `runStatus` string via a `switch` on `GroupStatus`. Two issues:

1. **`isDimmed` is passed in but `makeGroup` always constructs `ActiveJob` with `status: "completed"` and `conclusion: "success"` regardless.** When `testInProgressGroupAppearsLiveInDisplay` needs a live job, it has to *reconstruct* the entire group manually on top of `makeGroup`'s output — which is a sign that `makeGroup` doesn't actually model what it claims. If `buildGroupState` derives live/dimmed status from job states rather than from the group's `isDimmed` flag, this helper is only correct by coincidence for the completed cases.

2. The `switch status` block doesn't use exhaustive pattern matching benefits — if `GroupStatus` ever gains a new case, this silently breaks at runtime unless the enum is `@frozen`. A `default: fatalError()` would at least surface it in tests.

---

### Issue vs. Fix Scope

The original issue describes jobs that "never conclude" and "stick around and linger" even though GitHub shows them as done. The fix correctly addresses the fetch layer for a single SHA per run, but doesn't touch any cleanup or expiry logic for rows that were already stuck *before* this fix is deployed. There's no mention of whether existing stale rows self-heal on the next poll, or whether they require an app restart to clear. That should be explicitly stated — ideally tested.

---

### Minor: Comment Verbosity in Production Code

The added inline comment in `WorkflowActionGroupFetch.swift` is 4 lines explaining the fix history. A concise `// Fix #1041: include completed-only SHAs` one-liner is sufficient in source; the full context lives in the commit and this PR.

---

### Summary

The core fix is correct and well-motivated. The tests are above-average for a one-line change. The gaps worth addressing before merge:

- **Verify `seenIDs` lifetime** — confirm it's scoped per-`fetchActionGroups` call, not per-session
- **Add a combined-pass test** — SHA present in the in_progress pass *and* completed pass in the same poll cycle
- **Clarify stale-row heal behavior** — do pre-existing lingering rows auto-resolve on next poll after this fix?
- **Simplify the production comment** to one line

eonist:
## Deep Review

### Root Cause & Core Fix ✅

The diagnosis is accurate. The poll-cycle race condition — all runs for a SHA completing between the `in_progress`/`queued` pass and the `completed` pass — is the right explanation, and `bySha[run.headSha, default: []].append(run)` is the correct idiomatic fix.

One thing worth noting: the fix relies on `seenIDs` (local to `fetchActionGroups`) deduplicating a run that might appear in both the `in_progress` and `completed` passes mid-transition. This is safe as long as `run.id` is stable across status transitions — which it should be, but it's an implicit assumption that could use a comment near the `seenIDs` block.

---

### `trimSeenGroupIDs` — Contract doesn't match implementation

The doc comment says:
> "Entries are pruned FIFO (oldest half dropped)"

But `Set` is unordered, so this isn't FIFO — it's arbitrary eviction. The comment should say "arbitrary half" to avoid misleading future readers.

More importantly, the eviction formula:
```swift
let excess = ids.count - limit / 2
let toRemove = ids.prefix(excess)
```
With `limit = 200` and `ids.count = 201`, this removes `201 - 100 = 101` entries, leaving the set at **100**, not 200. The function is named "trim to limit" but actually trims to `limit/2`. The set then oscillates between 100–200 entries. Operationally fine given the cap is already generous, but the API is misleading. Consider either:
- `let excess = ids.count - limit` (evict only the minimum), or
- rename/document the half-eviction strategy explicitly.

---

### Cold-start re-fire of failure hook

`seenGroupIDs` is in-memory only and resets to `[]` on app restart. After a cold start, any completed group still present in GitHub's last-100-completed feed will re-fire the failure hook on the first poll. This was partially true before (via `actionGroupCache` not being persisted), but `seenGroupIDs` is now the *sole* deduplication gate for the hook — making this more structurally guaranteed. If the failure hook triggers a user-visible notification (sound, alert, badge), this is worth a follow-up issue.

---

### Missing unit test: `trimSeenGroupIDs`

Given the off-by-half eviction behaviour described above, `trimSeenGroupIDs` has no dedicated test. A test covering the boundary (`count == limit`, `count == limit + 1`, `count == limit * 2`) would lock in the intended semantics and catch the formula issue.

---

### Unintentional doc comment removals

The sweep that removed redundant inline comments in the test file went a bit too far in a couple of places in `RunnerStore.swift`:

- `// Priority: id → fullKey → name → nil` in `fetchAndEnrichRunners` — that was load-bearing context about lookup fallback order, not noise.
- `// Mirror the reset date so the UI can show an accurate countdown.` on the `rateLimitResetDate` assignment — the "why" is non-obvious and worth keeping.

---

### Minor: PR description test name mismatch

The PR table lists `testFireFailureHookNotCalledForAlreadyCachedGroup` but the actual test is named `testFireFailureHookNotCalledWhenGroupAlreadySeenEvenIfEvictedFromCache`. Trivial, but worth aligning.

---

### Test coverage is otherwise solid

`testCompletedOnlyGroupIsRoutedToCacheNotLive`, `testFireFailureHookCalledOnceForNewCompletedGroup`, `testPreviouslyLiveGroupSelfHealsAfterCompletion`, and `testShaWithBothLiveAndCompletedRunsProducesOneDisplayEntry` all cover the right scenarios. The split assertions in the mixed-SHA test are especially clean.

**TL;DR:** Fix is correct and the `seenGroupIDs` architecture is the right call. Main things to address: `trimSeenGroupIDs` evicts to `limit/2` (not `limit`), the FIFO claim is wrong, no test for `trimSeenGroupIDs`, and a couple of meaningful comments were dropped in the cleanup sweep.

eonist:
## Deep Review — fix(#1041): include completed-only SHAs in fetchActionGroups

### Root Cause — Diagnosis is Correct ✅

The race condition is precisely identified: if all sibling runs for a SHA finish between two poll cycles, that SHA never gets a `bySha` entry from the `in_progress`/`queued` passes, the completed-pass guard silently drops those runs, `allFetched` has no entry, `freezeVanishedGroups` has nothing to freeze, and the old live row lingers indefinitely. Tight diagnosis.

---

### The Core Fix — Correct and Minimal ✅

The one-liner in `WorkflowActionGroupFetch.swift` is idiomatic Swift and the right call. The existing `seenIDs` deduplication across the three fetch passes already prevents double-counting within a single poll cycle, so opening the gate for completed-only SHAs doesn't introduce phantom duplicates.

---

### The `seenGroupIDs` Design — Valid but Worth Scrutiny

#### 🟡 `trimSeenGroupIDs` doc comment falsely claims FIFO ordering

The doc comment says *"Entries are pruned FIFO (oldest half dropped)"* but `Set` is unordered and `prefix` on a `Set` returns arbitrary entries. The implementation itself even acknowledges this in a separate comment ("no true FIFO guarantee") — making the two comments contradict each other. The doc comment should be corrected to say **"random half-eviction"** or **"approximate eviction"**, and an inline `// NOTE: eviction order is arbitrary, not FIFO` would prevent future confusion.

#### 🟡 Missing unit tests for `trimSeenGroupIDs` boundary behavior

`trimSeenGroupIDs` has no direct test. Specifically missing:
- Assertion that the set is actually trimmed when `seenGroupIDsLimit` is exceeded.
- What happens if an ID is evicted from `seenGroupIDs` and then resurfaces in the GitHub feed — does the hook re-fire? That's the one logical hole in the entire approach, and there's no test covering it.

#### 🟡 Undocumented ordering invariant in `buildGroupState`

The `freezeVanishedGroups` hook guard change relies on `doneGroups` having already inserted into `newSeenGroupIDs` before `freezeVanishedGroups` is called. That ordering invariant is what prevents double-firing, but it's not documented or enforced anywhere. A brief comment at the call site would make this explicit and protect against future refactors reordering the calls.

---

### Minor Issues

- **🟢 `seenGroupIDsLimit = 200` magic number** — the comment explains it's larger than `groupCacheLimit`, but a sentence justifying *why 200 specifically* (e.g. "~6-7 poll cycles worth of completions at typical throughput") would help.
- **🟢 Removed intent-carrying comments** — `// Priority: id → fullKey → name → nil` in `RunnerStore.swift` and the explanatory comments in `buildJobState` backfill closures carried design intent that is now gone. Worth preserving as doc comments.
- **🟢 Cosmetic test comment removals** — stripping doc comments from existing tests in `ActiveJobIsLocalRunnerTests`, `RunnerModelDisplayStatusTests`, etc. is noise in this PR's diff and makes it harder to review the actual fix. These would be cleaner in a separate commit.

---

### Summary

| Severity | Issue |
|---|---|
| 🟡 Medium | `trimSeenGroupIDs` doc comment falsely claims FIFO — fix the comment |
| 🟡 Medium | Missing test for `trimSeenGroupIDs` boundary and post-eviction hook re-fire |
| 🟡 Medium | `doneGroups`-before-`freezeVanishedGroups` ordering invariant is undocumented |
| 🟢 Low | `seenGroupIDsLimit = 200` needs a sizing rationale comment |
| 🟢 Low | Removed inline comments lose design intent |
| 🟢 Low | Cosmetic test comment removals belong in a separate commit |

The core fix is correct and the `seenGroupIDs` architecture is a sound solution to the secondary hook-deduplication problem. The main gaps are the false FIFO claim in the doc comment and missing boundary tests for the new eviction logic. Nothing blocking, but the medium items should be addressed before merge.

eonist:
## Deep Review — fix(#1041): completed-only SHAs in fetchActionGroups

### Root cause diagnosis ✅

The PR description is accurate. The original `if bySha[run.headSha] != nil` guard in `WorkflowActionGroupFetch.swift` effectively made the completed-runs pass a *backfill* pass, not a *discovery* pass. Any SHA whose all runs finished between two polls would arrive at the completed-runs endpoint with no prior `bySha` entry, so the entire group was silently dropped. `freezeVanishedGroups` never ran for it because it was never in `allFetched`. The stale live row from the previous poll then lingered forever. The one-line fix is correct.

---

### The secondary problem (seenGroupIDs) — mostly right, one concern

The one-line fetch fix by itself opens a new hazard: previously-completed groups already in the cache will now also be returned by the completed endpoint every poll, because `seenIDs` (inside `fetchActionGroups`) only deduplicates within a single call, not across calls. This would spam `fireFailureHook` on every poll for every group in GitHub's last-100-completed feed until they age out of that feed (~30 days).

The `seenGroupIDs: Set<String>` solution correctly addresses this. However, there are a few things worth examining closely:

#### 1. Arbitrary eviction in `trimSeenGroupIDs` — this is the biggest risk

```swift
public static func trimSeenGroupIDs(_ ids: inout Set<String>, limit: Int) {
    guard ids.count > limit else { return }
    let excess = ids.count - limit
    let toRemove = ids.prefix(excess)
    ids.subtract(toRemove)
}
```

`Set` in Swift does not guarantee iteration order. `ids.prefix(excess)` returns elements in hash-table traversal order, which is *deterministic for a given run* but not semantically meaningful — it is not FIFO, not LRU, not sorted by age. The doc comment acknowledges "arbitrary eviction," and the test `testEvictedGroupIDRefiresHookOnNextPoll` correctly locks in the known re-fire behaviour.

**The concern:** at `seenGroupIDsLimit = 200`, under high-volume repos, an ID could be evicted while the corresponding group is *still inside GitHub's 100-completed feed* (which spans weeks). When that evicted ID resurfaces in the next poll, `fireFailureHook` fires again. The test documents this as a "known limitation," but the PR description doesn't surface this trade-off to the reviewer anywhere outside the test comment. It should be mentioned explicitly — both here and ideally in the issue.

**Mitigation options to consider:**
- Persist `seenGroupIDs` to `UserDefaults` (keyed to repo/scope) so it survives app restarts too — a restart today will also re-fire hooks for everything in the last-100-completed feed.
- Use an ordered structure (e.g., `[String]` ring buffer or `OrderedSet` from `swift-collections`) so oldest entries are evicted first, greatly reducing the re-fire window.
- At minimum, raise `seenGroupIDsLimit` substantially (e.g., 500–1000); 200 at "6–7 poll cycles of typical completions" sounds fine in theory but assumes a quiet repo. CI-heavy repos can produce 20–30 completed groups per push.

#### 2. `seenGroupIDs` is not persisted across app restarts

Every cold launch re-arms hooks for every group in GitHub's last-100-completed feed. If the app is restarted mid-day (update, crash, system restart), users will receive a wave of spurious failure notifications. This is a pre-existing gap that this PR did not introduce, but it is now more visible because the completed-only fetch path is widened. Worth a follow-up issue.

#### 3. Ordering invariant is correct and well-documented

The `doneGroups` loop populates `newSeenGroupIDs` before `freezeVanishedGroups` runs. This is the right call — it means a group in both the completed feed *and* `snapPrevGroups` fires exactly once. The code comment and the `testDoneGroupsSeenBeforeFreezeVanishedGroupsPreventsDoubleFire` test both document this clearly. 

#### 4. `freezeVanishedGroups` signature change is additive and safe

Adding `seenGroupIDs: Set<String> = []` as a defaulted parameter preserves call-site compatibility for any callers not yet passing the set. The changed guard:

```swift
// Before
if cache[groupID] == nil {
// After
if !seenGroupIDs.contains(groupID) && cache[groupID] == nil {
```

is correct — if the group is already in `newSeenGroupIDs` (inserted by `doneGroups`), the hook should not fire again here. Logic is sound.

---

### Test coverage — good, with one gap

The eight new tests in `PollResultBuilderGroupStateTests` are well-structured and cover the critical paths:

| Test | Assessment |
|---|---|
| `testCompletedOnlyGroupIsRoutedToCacheNotLive` | ✅ Core regression |
| `testInProgressGroupAppearsLiveInDisplay` | ✅ Guard rail |
| `testFireFailureHookCalledOnceForNewCompletedGroup` | ✅ Correct |
| `testFireFailureHookNotCalledWhenGroupAlreadySeenEvenIfEvictedFromCache` | ✅ Key scenario |
| `testPreviouslyLiveGroupSelfHealsAfterCompletion` | ✅ Self-heal path |
| `testShaWithBothLiveAndCompletedRunsProducesOneDisplayEntry` | ✅ Mixed-run group |
| `testEvictedGroupIDRefiresHookOnNextPoll` | ✅ Known-limitation lock-in |
| `testDoneGroupsSeenBeforeFreezeVanishedGroupsPreventsDoubleFire` | ✅ Ordering invariant |

**Missing test:** There is no test covering the `trimSeenGroupIDs` integration within `buildGroupState` at volume — i.e., verifying that when `seenGroupIDsLimit` is exceeded, the trimmed-down set is correctly returned in `newSeenGroupIDs` and that the count is bounded. The three unit tests for `trimSeenGroupIDs` in isolation are good but don't exercise the end-to-end flow.

**Missing test:** No test for the cold-start scenario (empty `snapSeenGroupIDs`, non-empty completed feed) at scale to verify hook does not fire for groups already in `snapGroupCache`. This scenario is partially covered by `testFireFailureHookNotCalledWhenGroupAlreadySeenEvenIfEvictedFromCache`, but only for the eviction path — not for the normal warm-cache path where cache has the entry.

---

### Minor nits

- **Removed comments in `RunnerStore.swift`:** The PR removes two inline comments — the rate-limit mirror comment and the `// Priority: id → fullKey → name → nil` comment. These are cleanups but they were genuinely useful orientation comments. The rate-limit one in particular explained a non-obvious design choice (`nil` correctly clears the countdown). Consider keeping or moving it to the property doc comment on `rateLimitResetDate`.
- **`snapSeenGroupIDs` naming:** The parameter is called `snapSeenGroupIDs` (matching the snapshot naming convention), but the internal mutable copy is `newSeenGroupIDs` even though it starts from the snapshot. This is consistent with `newCache` / `newPrevLive`, so it's fine — just note the convention is "new" = the value to be returned, regardless of what it starts from.
- **`seenGroupIDsLimit = 200` as a `public static let`:** Exposing this as public makes it testable, which is good. Just ensure no external consumer is hardcoding `200` with an assumption it won't change.

---

### Summary

The core bug fix (`bySha[run.headSha, default: []].append(run)`) is correct, minimal, and clearly motivated. The `seenGroupIDs` mechanism is a solid approach to the resulting idempotency problem. The primary open issue is the **arbitrary eviction + no persistence**, which means the re-fire hazard cannot be fully eliminated in-memory. This is acceptable for a first fix but should be tracked as a follow-up — either a persisted set or a better eviction policy — especially before this lands in repos with heavy CI throughput.

**Verdict: approve with follow-up** — the fix is correct and a clear improvement over the status quo. The eviction behaviour should be documented in a new issue.

eonist:
## Review follow-up / pushback response

Thanks for the detailed breakdown — very useful. Let me be precise about what's real vs. what I got wrong.

### Correction: `trimSeenGroupIDs` formula is actually correct in the file

I re-read the live file on the PR branch. The implementation is:

```swift
let excess = ids.count - limit
let toRemove = ids.prefix(excess)
ids.subtract(toRemove)
```

`ids.count - limit` is the right formula — it removes exactly the overshoot, leaving `limit` entries. The existing unit tests (`testTrimSeenGroupIDsTrimsToLimitNotHalf`, `testTrimSeenGroupIDsNoopAtLimit`, `testTrimSeenGroupIDsWellOverLimit`) all pass against this implementation and validate it correctly.

So: **no formula bug in the current code.** The test name `testTrimSeenGroupIDsTrimsToLimitNotHalf` is a guard against a *potential* regression (`limit / 2`), not evidence that the bug is present. I overstated this — my mistake.

---

### Confirmed: the doc comment FIFO claim is still wrong

The doc comment on `trimSeenGroupIDs` reads:

```
/// The set is trimmed to exactly `limit` entries by removing the minimum
/// overshoot — only entries beyond `limit` are dropped.
```

The first line of the doc block correctly says `Set is unordered so eviction order is arbitrary, not FIFO.` — that part is accurate. But the phrase *"removing the minimum overshoot"* implies a principled strategy that doesn't exist; `ids.prefix(excess)` on a `Set` returns arbitrary elements in hash-table traversal order. The comment is misleading even if not outright wrong. Suggested fix:

```swift
/// Trims the seen-group-IDs set to at most `limit` entries.
///
/// `Set` is unordered — eviction order is arbitrary, not FIFO or LRU.
/// Exactly `ids.count - limit` entries are removed when over the cap.
public static func trimSeenGroupIDs(_ ids: inout Set<String>, limit: Int) {
```

This is a doc-only change, no behaviour impact.

---

### Confirmed: persistence gap is real and now more impactful

The fetch-path widening in `WorkflowActionGroupFetch.swift` (completed-only SHAs now included) means a cold launch will now attempt to fire `fireFailureHook` for significantly more groups than before. Previously, only SHAs already known from in-progress/queued passes were backfilled; now the full last-100-completed feed flows through. This is the right trade-off for fixing #1041, but it makes the persistence gap more visible. A follow-up issue to persist `seenGroupIDs` to `UserDefaults` (keyed by scope) would close this cleanly.

---

### Summary

| Finding | Status |
|---|---|
| `trimSeenGroupIDs` formula bug (`limit/2`) | ❌ Not present in code — I was wrong |
| Doc comment misleading (FIFO claim) | ✅ Real — minor doc fix warranted |
| No persistence across restarts | ✅ Real — follow-up issue recommended |
| Arbitrary eviction at high CI throughput | ✅ Valid, acceptable as known limitation |

The two blockers I identified in the original review collapse to one cosmetic doc fix. The code is correct. **Approve.**

Issue #1038, opened 2026-05-30
⭐ Start refactoring and cleaning up the project


eonist:
Add comments to code. So that its hardenened for future agents missunderstanding and running havoc on the code when we do small tweaks

instead of strooing spec. whhich drifts. comments above all code that define the spec is easier. it avoids drift because agents also update comments when they edit code etc. spec.md docs just drift

1. create an inventory of all sorce code. make a tree list in markdown
2. do 3 or so files per turn when reviewing
3. mark the files that has been done as you move along
4. make branch and pr for each 3 files
5. make sure ci runs. 

## 📁 Source File Inventory

Progress legend: ⬜ = not yet reviewed/commented · ✅ = done

```
runner-bar/
├── .github/
├── .gitignore
├── .periphery.yml
├── .swiftlint.yml
├── Package.swift
├── README.md
├── GRAVEYARD.md
├── project.yml
├── build.sh
├── deploy.sh
├── install.sh
├── scrnsht.png
├── docs/
├── Resources/
│
├── Sources/
│   │
│   ├── RunnerBar/                              (app target — UI layer)
│   │   ├── ✅ main.swift                        ← PR #1098 (confirmed clean, no changes needed)
│   │   ├── ✅ Exports.swift                     ← PR #1098 (confirmed clean, no changes needed)
│   │   │
│   │   ├── App/
│   │   │   ├── ✅ AppDelegate.swift              ← PR #1104 (tearDownOpenState() extraction, async-hop removal, workspaceObserver type tightening, doc overhaul)
│   │   │   ├── ✅ AppDelegate+Navigation.swift   ← PR #1098 (stepLog label call-site update) · PR #1104 (mainView/settingsView doc hardening, stepLog TODO)
│   │   │   ├── ✅ AppDelegate+PanelSetup.swift   ← PR #1104 (popoverDidClose teardown dedup, per-subscription comments, KVO threading note)
│   │   │   ├── ✅ AppDelegate+StatusItem.swift   ← PR #1104 (menuBarImage triple-fallback + #if DEBUG assert, doc improvements)
│   │   │   ├── ✅ NavState.swift                ← PR #1098 (history comment trim, stepLog labels + - Parameters: docs)
│   │   │   ├── ✅ PanelSheetState.swift         ← PR #1098 (runnerSheetSnapshot doc, guard comment, runnerSheetSnapshot = nil bug fix, import Foundation removed)
[...]

Issue #773, opened 2026-05-23
bug: RunnerModel.displayStatus dead branch — "busy" label never shown


eonist:
## Problem

In `RunnerModel.swift`, both arms of the `isRunning` block return `"running"`, making the `isBusy` check unreachable:

```swift
if isRunning {
    if isBusy || githubStatus == "busy" { return "running" }
    return "running"   // ← identical, inner branch is dead
}
```

`statusColor` correctly returns `.busy` — so the dot colour is right, but the text label is always `"running"` regardless of busy state.

## Fix

```swift
if isRunning {
    return (isBusy || githubStatus == "busy") ? "busy" : "running"
}
```

## Severity
🔴 High — incorrect display for every busy runner.

PR #1047, opened 2026-05-31
refactor(core): type-safe RunnerStatus enum, dedup displayStatus/statusColor, fix swiftlint suppressions

### LINKED ISSUE #1046: refactor(core): type-safe RunnerStatus enum, dedup displayStatus/statusColor logic, fix swiftlint suppressions — Runner.swift / RunnerModel.swift / JobStatus.swift
## Problem

Three files in `RunnerBarCore/Runner/` have a cluster of related issues that compound each other:

### 1. Stringly-typed runner status (`Runner.swift`, `RunnerModel.swift`, `RunnerStatusEnricher.swift`)

`Runner.status` is `String` and `RunnerModel.githubStatus` is `String?`. Both fields carry the same semantic — the GitHub API status of a runner (`"online"` / `"offline"` / `"busy"`). Logic in `displayStatus` and `statusColor` pattern-matches against string literals in two separate places:

```swift
// RunnerModel.swift — repeated in BOTH displayStatus and statusColor
case "online": return "online"
case "busy":   return "busy"
default:       return "offline"   // silent fallthrough for any unknown value
```

This means:
- A new GitHub API status value (e.g. `"draining"`) silently falls through to `"offline"`
- The compiler cannot enforce exhaustiveness — a future change to one branch doesn't force a change to the other
- The same `online/offline/busy` tri-state is duplicated across `Runner`, `RunnerModel`, and `RunnerStatusEnricher`

### 2. Logic duplication between `displayStatus` and `statusColor` (`RunnerModel.swift`)

`displayStatus` and `statusColor` both implement the same `lifecycleWarning → isRunning → isBusy/githubStatus` conditional tree independently. When logic changes, it must be kept in sync manually.

### 3. Dead `swiftlint:disable` suppressions (`RunnerModel.swift`, `RunnerStatusEnricher.swift`)

`.swiftlint.yml` already globally disables `type_body_length`, `function_body_length`, and `function_parameter_count`. The in-file suppressions for these rules are therefore dead.

### 4. `swiftlint:disable missing_docs` fighting the linter (`JobStatus.swift`)

`JobStatus.swift` suppresses `missing_docs` at file level because extension members lack `///` doc comments. The fix is to add the docs and remove the suppress.

### 5. Intent not documented on `isActive` and `isFailure` (`JobStatus.swift`)

`isActive` treats `.unknown` as inactive — deliberate but invisible. `isFailure` excludes `.cancelled` and `.skipped` — also deliberate but undocumented.

### 6. Trivial noise comments violating commenting standard Rule 3

```swift
/// The cpu constant.       ← restates the name, adds nothing
/// The running case.       ← restates the name, adds nothing
/// The id case.            ← restates the name, adds nothing
```

---

## Solution checklist

- [x] **A. Introduce `RunnerStatus` enum** — `RunnerStatus.swift` added with `Codable`, `CustomStringConvertible`, `unknown(String)` forward-compat fallback
- [x] **B. Replace `Runner.status: String` with `RunnerStatus`** — `displayStatus` now uses exhaustive switch, `.unknown` maps to `"offline"`
- [x] **C. Replace `RunnerModel.githubStatus: String?` with `RunnerStatus?`** — `resolvedState` private computed var extracts shared logic, drives both `displayStatus` and `statusColor`
- [x] **D. `RunnerStatusEnricher.swift`** — dead suppressions removed, raw string converted via `RunnerStatus(rawString:)`
- [x] **E. `JobStatus.swift`** — `///` added to all extension members, `missing_docs` suppress removed, intent comments added to `isActive` and `isFailure`
- [x] **F. Tests updated** — `makeRunner` helper uses `RunnerStatus`, new suites for `statusColor`, `displayStatus`, `isActive`, `isFailure`
- [x] **G. `RunnerStore.aggregateStatus` bug fixed** — `.busy` runners now counted as online alongside `.online` runners (all-busy fleet no longer reports `.allOffline`)
- [ ] **H. Noise `CodingKeys` case docs** — tautological per-case comments in `Runner.swift` still present (non-blocking)
- [ ] **I. `RunnerStatus` missing `ExpressibleByStringLiteral`** — inconsistency vs `JobStatus`/`JobConclusion` (non-blocking)

---

## CI impact

| Check | Impact |
|---|---|
| `swift test` | `RunnerBarCoreTests` updated to use `RunnerStatus` enum — all existing tests preserved |
| `swiftlint --strict` | Dead `disable` lines removed, `missing_docs` suppression removed, all extension members get `///` |
| `periphery scan` | `RunnerStatus` is `public` and used by `Runner` + `RunnerModel` — `retain_public: true` keeps it clean |
| `codeql` | No impact — pure refactor, no new sinks or sources |

## Files changed

- `Sources/RunnerBarCore/Runner/RunnerStatus.swift` ← new file
- `Sources/RunnerBarCore/Runner/Runner.swift`
- `Sources/RunnerBarCore/Runner/RunnerModel.swift`
- `Sources/RunnerBarCore/Runner/RunnerStatusEnricher.swift`
- `Sources/RunnerBarCore/Runner/JobStatus.swift`
- `Sources/RunnerBarCore/Runner/RunnerStore.swift`
- `Tests/RunnerBarCoreTests/RunnerBarCoreTests.swift`

eonist:
## **User description**
Closes #1046

## What

Batch 1 of the codebase review (#1038). Three files in `RunnerBarCore/Runner/` plus their supporting enricher and tests.

### New: `RunnerStatus.swift`
Typed enum (`online` / `busy` / `offline` / `unknown(String)`) replacing the raw `String` status fields across `Runner` and `RunnerModel`. Follows the same `unknown(String)` forward-compat pattern already used in `JobStatus` and `JobConclusion`. Full `Codable`, `CustomStringConvertible` conformances with `///` on every member — no `missing_docs` suppression needed.

### `Runner.swift`
- `status: String` → `status: RunnerStatus`
- `displayStatus` now uses `guard status != .offline` — no string literals
- Removed trivial `CodingKeys` case comments
- Added `- Note:` distinguishing `Runner` (API) from `RunnerModel` (local)
- Full file header with `See:` collaborators line

### `RunnerModel.swift`
- `githubStatus: String?` → `githubStatus: RunnerStatus?`
- Extracted private `ResolvedState` enum + `resolvedState` computed var — single source of truth consumed by both `displayStatus` and `statusColor`. Eliminates the duplicated `lifecycleWarning → isRunning → isBusy/githubStatus` conditional tree
- Removed 2 dead `// swiftlint:disable` lines (`type_body_length` + `function_parameter_count` already off globally in `.swiftlint.yml`)
- Added `- Note:` on `@unchecked Sendable` explaining why it's safe
- Added `- SeeAlso: Runner` at type level
- Replaced trivial `StatusColor` case comments with meaningful descriptions
- Added `#773` reference in `displayStatus` doc

### `RunnerStatusEnricher.swift`
- `applyEnrichment` now maps `String? → RunnerStatus?` via `statusString.map { RunnerStatus(rawString: $0) }`
- Removed dead `// swiftlint:disable function_parameter_count type_body_length` suppress/enable pair
- Added proper `///` on all public methods with `- Parameter:`, `- Returns:`, `- Note:` tags

### `JobStatus.swift`
- Removed `// swiftlint:disable missing_docs` — fixed by adding `///` to all extension members (`Codable`, `CustomStringConvertible`, `ExpressibleByStringLiteral` on both `JobStatus` and `JobConclusion`)
- Added `- Note:` to `isActive` explaining why `.unknown` is treated as inactive
- Added `- Note:` to `isFailure` explaining why `.cancelled` and `.skipped` are excluded
- Added descriptive `///` to all enum cases

### `RunnerBarCoreTests.swift`
- `makeRunner(githubStatus: String)` → `makeRunner(githubStatus: RunnerStatus)` — uses typed enum throughout
- `testDisplayStatusDefaultsToOfflineForUnknownStatus` now passes `.unknown("draining")` — correctly exercises the forward-compat path
- Added `JobStatusIsActiveTests` — covers all active statuses, completed, and the `.unknown` inactive case
- Added `JobConclusionIsFailureTests` — covers failure conclusions, non-failures, `.cancelled`, `.skipped`, and `.unknown`

## CI

| Check | Expected |
|---|---|
| `swift test --filter RunnerBarCoreTests` | ✅ All existing tests pass + 10 new tests |
| `swiftlint --strict` | ✅ No suppressions remaining in changed files |
| `periphery scan` | ✅ `RunnerStatus` is `public` + used; `retain_public: true` keeps it clean |
| `codeql` | ✅ Pure refactor, no new sinks or sources |

___

## **CodeAnt-AI Description**
**Use typed runner and job statuses to avoid mismatched states and clarify runner display**

### What Changed
- Runner statuses now use fixed values instead of free-form strings, so runner state shown in the UI stays consistent and new API values do not break decoding
- Runner rows now decide their text and dot colour from one shared state path, keeping “running”, “busy”, “online”, “offline”, and lifecycle warning states aligned
- Job status and job conclusion handling now treats unknown status values as inactive, while keeping cancelled and skipped jobs out of failure alerts
- Added coverage for runner state, active job status, and failure detection rules

### Impact
`✅ Fewer wrong runner status labels`
`✅ Clearer runner dots and row text`
`✅ Fewer false failure alerts from cancelled or skipped jobs`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed runner online count calculation to properly compare status values.

* **Documentation**
  * Enhanced documentation for job status and conclusion handling with explicit guidance on forward-compatibility for unknown API values.

* **Refactor**
  * Replaced raw string runner status values with a typed enum for improved reliability and consistency throughout the application.

### DIFF
--- a/Sources/RunnerBar/Runner/RunnerStore.swift
+++ b/Sources/RunnerBar/Runner/RunnerStore.swift
@@ -51,13 +51,16 @@ final class RunnerStore {
     /// The scopeCancellable property.
     private var scopeCancellable: AnyCancellable?
 
-    /// Emits whenever a fetch cycle completes and the store's state has been updated.
+    /// Emits whenever a fetch cycle completes and the store’s state has been updated.
     let didUpdate = PassthroughSubject<Void, Never>()
 
     /// The aggregateStatus property.
+    ///
+    /// A runner with `.busy` status is connected to GitHub and executing a job,
+    /// so it counts toward the online tally alongside `.online` runners.
     var aggregateStatus: AggregateStatus {
         guard !runners.isEmpty else { return .allOffline }
-        let onlineCount = runners.filter { $0.status == .online }.count
+        let onlineCount = runners.filter { $0.status == .online || $0.status == .busy }.count
         if onlineCount == runners.count { return .allOnline }
         if onlineCount == 0 { return .allOffline }
         return .someOffline
@@ -171,7 +174,7 @@ final class RunnerStore {
     /// - Secondary:  "runnerName"        → installPath  (name-only fallback)
     /// - Tertiary:   agentId (Int)        → installPath  (ID-based, scope-agnostic)
     ///
-    /// The ID map is the most reliable — GitHub writes the runner's integer ID
+    /// The ID map is the most reliable — GitHub writes the runner’s integer ID
     /// to the `.runner` JSON on disk during `config.sh`, so it is stable across
     /// renames and scope-string format changes.  Runners that predate this field
     /// (agentId == nil) fall through to the fullKey / name maps.
@@ -199,11 +202,11 @@ final class RunnerStore {
         return InstallPathMap(byFullKey: byFullKey, byName: byName, byId: byId)
     }
 
-    /// Applies a completed fetch cycle's results to the store's @MainActor state.
+    /// Applies a completed fetch cycle’s results to the store’s @MainActor state.
     ///
     /// Copies `ghIsRateLimited` and `ghRateLimitResetDate` from the transport
     /// layer so the full rate-limit context (flag + exact reset moment) is
-    /// available to `RunnerViewModel` and ultimately to `PanelMainView`'s
+    /// available to `RunnerViewModel` and ultimately to `PanelMainView`’s
     /// live-countdown banner.
     private func applyFetchResult(
         enrichedRunners: [Runner],
--- a/Sources/RunnerBar/Runner/RunnerStore.swift
+++ b/Sources/RunnerBar/Runner/RunnerStore.swift
@@ -57,7 +57,7 @@ final class RunnerStore {
     /// The aggregateStatus property.
     var aggregateStatus: AggregateStatus {
         guard !runners.isEmpty else { return .allOffline }
-        let onlineCount = runners.filter { $0.status == "online" }.count
+        let onlineCount = runners.filter { $0.status == .online }.count
         if onlineCount == runners.count { return .allOnline }
         if onlineCount == 0 { return .allOffline }
         return .someOffline
--- a/Sources/RunnerBarCore/Runner/JobStatus.swift
+++ b/Sources/RunnerBarCore/Runner/JobStatus.swift
@@ -162,10 +162,13 @@ public enum JobConclusion: Hashable, Sendable {
 
     /// Returns `true` for terminal failure-like conclusions that should trigger alerts.
     ///
-    /// - Note: `.cancelled` and `.skipped` are intentionally excluded. Both are
-    ///   user-initiated or dependency-driven outcomes — not CI failures. Treating
-    ///   them as failures would trigger the failure hook for routine branch
-    ///   cancellations and conditional-skip patterns.
+    /// - Note: `.cancelled` and `.skipped` are intentionally excluded — both are
+    ///   user-initiated or dependency-driven outcomes, not CI errors. Treating them
+    ///   as failures would trigger the failure hook for routine branch cancellations
+    ///   and conditional-skip patterns.
+    /// - Note: `.neutral` is also excluded — it represents an inconclusive outcome
+    ///   with no definitive pass/fail signal, not an error condition. Same class of
+    ///   outcome as `.skipped`: informational, not actionable.
     public var isFailure: Bool {
         switch self {
         case .failure, .timedOut, .startupFailure, .actionRequired: return true
--- a/Sources/RunnerBarCore/Runner/JobStatus.swift
+++ b/Sources/RunnerBarCore/Runner/JobStatus.swift
@@ -1,18 +1,31 @@
 // JobStatus.swift
 // RunnerBarCore
-// swiftlint:disable missing_docs
+//
+// Typed enums for the GitHub Actions job/workflow status and conclusion fields.
+// Both use an unknown(String) fallback for forward-compatibility with new API values.
+// See: ActiveJob, WorkflowActionGroup, PollResultBuilder
 import Foundation
 
-// MARK: - Job status
+// MARK: - JobStatus
 
 /// The lifecycle status of a GitHub Actions job or workflow run.
 public enum JobStatus: Hashable, Sendable {
+    /// Job is waiting to be picked up by a runner.
     case queued
+    /// Job is currently executing on a runner.
     case inProgress
+    /// Job has finished (see `JobConclusion` for the outcome).
     case completed
+    /// Job is waiting on a required approval or environment protection rule.
     case waiting
+    /// Job has been requested but not yet queued.
     case requested
+    /// Job is pending deployment to a protected environment.
     case pending
+    /// A status value not recognised at compile time.
+    ///
+    /// - Note: Acts as a forward-compatible fallback so new GitHub API status
+    ///   values do not cause a decode failure or break polling logic.
     case unknown(String)
 
     /// The raw string value as returned by the GitHub API.
@@ -28,7 +41,9 @@ public enum JobStatus: Hashable, Sendable {
         }
     }
 
-    /// Initialise from a raw API string. Unknown values map to `.unknown(raw)`.
+    /// Initialises from a raw API string. Unknown values map to `.unknown(raw)`.
+    ///
+    /// - Parameter raw: The raw string value as returned by the GitHub REST API.
     public init(rawString raw: String) {
         switch raw {
         case "queued":       self = .queued
@@ -41,7 +56,11 @@ public enum JobStatus: Hashable, Sendable {
         }
     }
 
-    /// Returns `true` when the job / run is still active (not yet completed).
+    /// Returns `true` when the job or run is still active (not yet completed).
+    ///
+    /// - Note: `.unknown` is treated as **inactive** to avoid polling indefinitely
+    ///   if GitHub introduces a new status value this client does not yet recognise.
+    ///   Erring on the side of stopping the poll is safer than a stuck spinner.
     public var isActive: Bool {
         switch self {
         case .queued, .inProgress, .waiting, .requested, .pending: return true
@@ -51,10 +70,13 @@ public enum JobStatus: Hashable, Sendable {
 }
 
 extension JobStatus: Codable {
+    /// Decodes from a single-value string container.
     public init(from decoder: Decoder) throws {
         let raw = try decoder.singleValueContainer().decode(String.self)
         self = JobStatus(rawString: raw)
     }
+
+    /// Encodes as a single-value string container.
     public func encode(to encoder: Encoder) throws {
         var container = encoder.singleValueContainer()
         try container.encode(rawValue)
@@ -62,28 +84,43 @@ extension JobStatus: Codable {
 }
 
 extension JobStatus: CustomStringConvertible {
+    /// Returns the raw API string value.
     public var description: String { rawValue }
 }
 
 extension JobStatus: ExpressibleByStringLiteral {
+    /// Initialises from a string literal. Delegates to `init(rawString:)`.
     public init(stringLiteral value: String) {
         self = JobStatus(rawString: value)
     }
 }
 
-// MARK: - Job conclusion
+// MARK: - JobConclusion
 
 /// The outcome of a completed GitHub Actions job or workflow run.
 public enum JobConclusion: Hashable, Sendable {
+    /// All steps completed successfully.
     case success
+    /// One or more steps failed.
     case failure
+    /// The job was cancelled by a user or another workflow.
     case cancelled
+    /// The job was skipped due to an `if:` condition evaluating to false.
     case skipped
+    /// The job exceeded its configured timeout.
     case timedOut
+    /// A manual approval is required before the job can proceed.
     case actionRequired
+    /// The job completed without a definitive pass/fail outcome.
     case neutral
+    /// The job became stale waiting for an external event.
     case stale
+    /// The runner failed to initialise before the job could start.
     case startupFailure
+    /// A conclusion value not recognised at compile time.
+    ///
+    /// - Note: Acts as a forward-compatible fallback so new GitHub API conclusion
+    ///   values do not cause a decode failure.
     case unknown(String)
 
     /// The raw string value as returned by the GitHub API.
@@ -102,7 +139,9 @@ public enum JobConclusion: Hashable, Sendable {
         }
     }
 
-    /// Initialise from a raw API string. Unknown values map to `.unknown(raw)`.
+    /// Initialises from a raw API string. Unknown values map to `.unknown(raw)`.
+    ///
+    /// - Parameter raw: The raw string value as returned by the GitHub REST API.
     public init(rawString raw: String) {
         switch raw {
         case "success":          self = .success
@@ -118,7 +157,12 @@ public enum JobConclusion: Hashable, Sendable {
         }
     }
 
-    /// Returns `true` for terminal failure-like conclusions.
+    /// Returns `true` for terminal failure-like conclusions that should trigger alerts.
+    ///
+    /// - Note: `.cancelled` and `.skipped` are intentionally excluded. Both are
+    ///   user-initiated or dependency-driven outcomes — not CI failures. Treating
+    ///   them as failures would trigger the failure hook for routine branch
+    ///   cancellations and conditional-skip patterns.
     public var isFailure: Bool {
         switch self {
         case .failure, .timedOut, .startupFailure, .actionRequired: return true
@@ -128,10 +172,13 @@ public enum JobConclusion: Hashable, Sendable {
 }
 
 extension JobConclusion: Codable {
+    /// Decodes from a single-value string container.
     public init(from decoder: Decoder) throws {
         let raw = try decoder.singleValueContainer().decode(String.self)
         self = JobConclusion(rawString: raw)
     }
+
+    /// Encodes as a single-value string container.
     public func encode(to encoder: Encoder) throws {
         var container = encoder.singleValueContainer()
         try container.encode(rawValue)
@@ -139,12 +186,13 @@ extension JobConclusion: Codable {
 }
 
 extension JobConclusion: CustomStringConvertible {
+    /// Returns the raw API string value.
     public var description: String { rawValue }
 }
 
 extension JobConclusion: ExpressibleByStringLiteral {
+    /// Initialises from a string literal. Delegates to `init(rawString:)`.
     public init(stringLiteral value: String) {
         self = JobConclusion(rawString: value)
     }
 }
-// swiftlint:enable missing_docs
--- a/Sources/RunnerBarCore/Runner/JobStatus.swift
+++ b/Sources/RunnerBarCore/Runner/JobStatus.swift
@@ -69,6 +69,7 @@ public enum JobStatus: Hashable, Sendable {
     }
 }
 
+/// `Codable` conformance for `JobStatus` — encodes and decodes as a plain string.
 extension JobStatus: Codable {
     /// Decodes from a single-value string container.
     public init(from decoder: Decoder) throws {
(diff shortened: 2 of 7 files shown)


eonist:
## Deep Review

Overall this is a high-quality, well-scoped refactor that solves every problem stated in #1046. A few observations before merge:

---

### ✅ What's solid

**`ResolvedState` is the best decision in this diff.** The original `displayStatus` and `statusColor` each independently walked the same `lifecycleWarning → isRunning → isBusy/githubStatus` tree. Any future bug fix had to be applied twice, in sync, forever. The private `resolvedState` computed var eliminates that — one decision tree, two pure mapping consumers.

**`unknown(String)` forward-compat is the right call.** Consistent with `JobStatus`/`JobConclusion`, unrecognised API values are preserved and safely fall through to `offline` in display logic. Correct for a long-lived app talking to a versioned external API.

**Removing dead SwiftLint suppressions is a real correctness improvement**, not cosmetic. A reader seeing `// swiftlint:disable type_body_length` reasonably concludes that rule is active in the file.

---

**1. `Runner.displayStatus` + `.unknown` — untested behavioural gap**

`Runner.displayStatus` uses:
```swift
guard status != .offline else { return "offline" }
```
This means `status == .unknown("draining")` falls through to the `idle`/`active` branch and returns `"idle (CPU: — MEM: —)"` — probably not the right UX for an unrecognised status. There's no test covering this path. Suggest adding a test and deciding whether `.unknown` on `Runner` should display `"offline"` or a specific label (it behaves differently from `RunnerModel` which maps `default → .offline` in `resolvedState`).

**2. `CodingKeys` expansion reintroduces noise comments**

The issue itself identified and removed trivial restatement comments like `/// The id case.`. The new `CodingKeys` expansion replaces that with `/// Maps to the 'id' JSON key.` on each of `id`, `name`, `status`, `busy` — which is equally obvious for a `CodingKeys` enum where the case name and the JSON key are identical. Suggest reverting to the single-line `case id, name, status, busy` form.

---

### ℹ️ Minor observations

**3. `RunnerStatus` missing `ExpressibleByStringLiteral`**
`JobStatus` and `JobConclusion` both conform to it. `RunnerStatus` does not. Not a blocker (tests use typed enum), but worth a deliberate comment if the omission is intentional.

**4. `resolvedState` — `.running` vs `.idle` distinction deserves a doc note**
When `isRunning == true && githubStatus == .online` → `.running` (green dot). When `isRunning == false && githubStatus == .online` → `.idle` (yellow dot). The semantics are correct but subtle — a future maintainer could collapse those two cases and silently break dot colour. Worth adding a `- Note:` to `resolvedState`'s doc comment.

**5. `@unchecked Sendable` — consider a TODO**
The note explaining why it's safe (all mutations on `@MainActor`) is exactly right. Consider adding `// TODO: Remove @unchecked once properties are all let or mutations are actor-isolated` so the compromise doesn't become permanent by default.

**6. `applyEnrichment` local variable shadowing**
`let githubStatus = statusString.map { … }` shadows `RunnerModel.githubStatus` property name. Not a bug, but momentarily confusing to read.

---

The one item I'd want resolved before merge is **#1** — the `Runner.displayStatus` + `.unknown` edge case is a real untested behaviour difference from `RunnerModel`.

eonist:
## Deep Review

The core refactor — `RunnerStatus` enum, `resolvedState` extraction, dead suppression removal — is correct and well-executed. Issue-to-code traceability is excellent. A few non-trivial things worth addressing before merge:

---

### 🔴 Logic bug — `RunnerModel.displayStatus`

```swift
case .warning:  return lifecycleWarning ?? "error"
```

`resolvedState` only returns `.warning` when `lifecycleWarning != nil`, so the `?? "error"` branch is **unreachable dead code**. It implies the nil case is possible when it isn't, and will silently produce `"error"` instead of surfacing a bug if `resolvedState` ever gains a new path to `.warning` without setting `lifecycleWarning`. Use `lifecycleWarning!` here — it's safe and honest — or restructure so the nil-coalescing is never needed.

---

### 🟡 Design gap — `ResolvedState.idle` naming

`case .idle` maps to `return "online"` in `displayStatus`. The name `.idle` suggests "not busy" but the actual meaning is "not running locally, GitHub says online". A reader has to bridge that gap. Consider renaming to `.githubOnline` or at minimum adding a doc comment on the case explaining it resolves to the `"online"` display string.

---

### 🟡 Test gap — `statusColor` has zero coverage after the refactor

The PR adds 10 new tests but none cover `statusColor`, which was just significantly refactored through `resolvedState`. Since both `displayStatus` and `statusColor` share the same computed var, a bug in `resolvedState` silently affects dot colours in the UI with no failing test. A `RunnerModelStatusColorTests` class mirroring the existing display-status cases would close this gap.

---

### 🟡 Consistency — `RunnerStatus` missing `ExpressibleByStringLiteral`

`JobStatus` and `JobConclusion` both gain `ExpressibleByStringLiteral` in this PR (reason: "supports test literals"). `RunnerStatus` doesn't get it. This creates an inconsistency — if any future test helper tries to pass a string literal as a `RunnerStatus`, it won't compile. The same rationale that justified adding it to the job types applies here.

---

### 🟢 Nit — `CodingKeys` case docs in `Runner.swift`

The old comments (`/// The id case.`) were called out in #1046 as Rule 3 violations. The replacements (`/// Maps to the 'id' JSON key.`) are better but still near-tautological for identity-mapped keys where the Swift name and JSON key are identical. These are worth dropping entirely; the only `CodingKeys` docs worth writing are for non-obvious name mappings.

---

### 🟢 Nit — Redundant `Codable` docs in `JobStatus.swift` / `JobConclusion.swift`

The extension-level comment (`/// 'Codable' conformance for 'JobStatus' — encodes and decodes as a plain string.`) + `/// Decodes from a single-value string container.` + `/// Encodes as a single-value string container.` all say the same thing three times. The extension-level comment is sufficient; the method-level ones on standard protocol requirements add no new information.

---

### 🟢 Nit — `neutral` exclusion in `JobConclusion.isFailure` undocumented

`.cancelled` and `.skipped` get explicit `- Note:` explanations. `.neutral` is also excluded from `isFailure` but silently — a reader could reasonably argue an inconclusive outcome should trigger an alert. One line explaining the rationale (same class of outcome as `skipped`: not a CI error) would close the loop.

---

**Priority before merge:** The `?? "error"` issue is the only one I'd call a must-fix — it's not a runtime bug today, but it's a logic hole. Everything else is polish.

eonist:
## Deep Review

Overall this is a solid, well-scoped refactor that closes every problem stated in #1046. The type system is now doing work that string comparisons were doing before. A few things worth flagging before merge — two of them are real blockers.

---

### ✅ What's Done Well

**`ResolvedState` extraction is the highlight of the PR.** The old `displayStatus`/`statusColor` duplication was a real maintenance hazard — two independent conditional trees that had to be kept in sync manually. The new `resolvedState` computed var as a private single source of truth consumed by both properties is the correct solution and the implementation is clean.

**`unknown(String)` forward-compat pattern is applied consistently.** `RunnerStatus`, `JobStatus`, and `JobConclusion` all follow the same `unknown(String)` fallback that prevents decode failures when the GitHub API introduces new values. The choice to treat `unknown` as inactive/offline/non-failure across the board is well-reasoned and now documented.

**Test coverage is meaningfully expanded.** 10 new tests cover `RunnerModel.statusColor` (zero tests before), `Runner.displayStatus`, `JobStatus.isActive` edge cases, and `JobConclusion.isFailure` edge cases including `.cancelled`, `.skipped`, and `.unknown`. `testDisplayStatusDefaultsToOfflineForUnknownStatus` now correctly exercises `.unknown("draining")` instead of the old `"unknown"` string literal.

---

### 🔴 Blockers

#### 1. `RunnerStore.swift` — `.busy` runners excluded from `onlineCount`

```swift
// current
let onlineCount = runners.filter { $0.status == .online }.count
```

A runner with `status == .busy` is reachable and consuming a job — it should count toward "online" for the aggregate. The old string comparison had the same bug, so this PR preserves an existing defect. Now that there's a typed enum, the fix is a one-liner:

```swift
let onlineCount = runners.filter { $0.status == .online || $0.status == .busy }.count
```

A fleet that is all-busy would currently report as `.allOffline` in the aggregate — real behavioral regression risk.

#### 2. Force-unwrap in `RunnerModel.displayStatus`

```swift
case .warning: return lifecycleWarning!  // resolvedState only reaches .warning when lifecycleWarning != nil
```

The comment is correct, but a force-unwrap feeding UI is a code smell even when provably safe. If a future `ResolvedState` path reaches `.warning` without the guard, it crashes silently at runtime. The issue spec already suggested the safer form:

```swift
case .warning: return lifecycleWarning ?? "error"
```

---

### 🟡 Design Questions

#### 3. `Runner.displayStatus` — `.busy` API status path is implicit

```swift
switch status {
case .offline, .unknown: return "offline"
default: break
}
let label = busy ? "active" : "idle"
```

If the API sends `status: "busy"` but `busy == false` (race condition between status and busy-flag update on GitHub's side), the runner silently shows `"idle"`. This is probably fine in practice but it's an unguarded assumption. There's no test for `status: .busy, busy: false`. Worth a `- Note:` comment at minimum, and ideally a test case.

#### 4. `ResolvedState.githubOnline` naming

The issue spec proposed `.idle` for this case. `.githubOnline` is more precise (it distinguishes "process not running locally but GitHub sees it") but it leaks implementation detail into a display-layer enum. The doc comment covers the distinction, but callers still need to know `.githubOnline` → `"online"` text and `.idle` color. Consider `.remoteOnline` as a middle ground, or re-evaluate whether the local-vs-remote distinction needs to live in the case name vs. the docs.

#### 5. `RunnerStatus` missing `ExpressibleByStringLiteral`

`JobStatus` and `JobConclusion` both conform to `ExpressibleByStringLiteral`. `RunnerStatus` does not. The asymmetry is inconsistent — either add it for `RunnerStatus` or document why it's intentionally excluded.

---

### 🟢 Minor

- **Duplicated `makeRunner` factory**: `RunnerModelDisplayStatusTests` and `RunnerModelStatusColorTests` define identical private helpers. Extract to a file-private free function.
- **Tautological `CodingKeys` docs in `Runner.swift`**: `/// Maps to the id JSON field` on `case id` is a tautology. These per-case comments add noise where there's no non-obvious mapping.
- **`import Foundation` position in `RunnerStatusEnricher.swift`**: Moved below the multi-line file header comment after the diff — cosmetic but inconsistent with the rest of the codebase.
- **`backfill: { _ in }` comment removal in `PollResultBuilderTests`**: Fine to remove, but not mentioned as intentional in the PR description — could confuse a future bisect.

---

### Priority table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Bug | `RunnerStore.swift` | `.busy` runners excluded from `onlineCount` |
| 2 | 🔴 Fix | `RunnerModel.swift` | Force-unwrap `lifecycleWarning!` in display path |
| 3 | 🟡 Test | `Runner.swift` | `status == .busy, busy == false` path untested |
| 4 | 🟡 Design | `RunnerModel.swift` | `.githubOnline` naming in `ResolvedState` |
| 5 | 🟡 Consistency | `RunnerStatus.swift` | Missing `ExpressibleByStringLiteral` vs `JobStatus` |
| 6 | 🟢 Noise | `Runner.swift` | Tautological `CodingKeys` per-case doc comments |
| 7 | 🟢 Test | `RunnerBarCoreTests.swift` | Duplicated `makeRunner` across two test classes |

**#1 and #2 before merge. Everything else is non-blocking.**

eonist:
## Deep Review

Overall this is a solid, well-scoped refactor that fully delivers on #1046. The `ResolvedState` extraction is the standout design win. A few divergences from the issue spec and two missing test gaps worth addressing before merge.

---

### Design & Architecture

**`ResolvedState` — the best thing in this PR**
Extracting a private `ResolvedState` enum as the single source of truth for `displayStatus` and `statusColor` is exactly right. The old code had two parallel `if lifecycleWarning → if isRunning → switch githubStatus` trees that could silently diverge on future changes. Now they can't. The doc comment explaining why `.running` and `.githubOnline` are *deliberately* distinct cases (green dot vs yellow dot semantics) is excellent — this is genuinely non-obvious and would be the first thing a future reader would collapse incorrectly.

**`RunnerStatus` is not `RawRepresentable`**
It uses a hand-rolled `rawValue` computed var and `init(rawString:)` instead of `init(rawValue:)`. That's intentional and correct for the `unknown` case — but anyone reaching for `.rawValue` expecting `RawRepresentable` semantics will be surprised there's no `init(rawValue:)` returning `Optional`. Suggest adding a `// NOTE: not RawRepresentable — rawValue is a plain computed var` near the declaration to preempt confusion.

---

### Divergence from Issue Spec

**`Runner.displayStatus` — quietly better than the spec**
The issue proposed `guard status != .offline else { return "offline" }`, which would have let `.unknown("draining")` fall through to `"idle (CPU: — MEM: —)"`. The PR ships `case .offline, .unknown: return "offline"`, which is correct. Worth an explicit callout in the PR description since it's a behavioral improvement over the spec.

**`ResolvedState` case naming**
The issue names the idle-but-GitHub-online case `.idle`; the PR names it `.githubOnline`. The PR's name is clearer — `.idle` is already used by `StatusColor` with a different semantic scope. Good call.

**`lifecycleWarning!` force-unwrap**
```swift
case .warning:     return lifecycleWarning!  // resolvedState only reaches .warning when lifecycleWarning != nil
```
Safe by invariant and the comment explains why. The issue spec used `lifecycleWarning ?? "error"` as a nil-safe fallback — the force-unwrap is actually more honest here since it would crash loudly if the invariant ever breaks rather than silently showing `"error"`. Keep it, just make sure the invariant comment survives any future refactor of this method.

---

### Correctness Gaps

**⚠️ `RunnerStore.aggregateStatus` — `.busy` runners not counted as online**
```swift
let onlineCount = runners.filter { $0.status == .online }.count
```
A runner with `.busy` status *is* connected to GitHub — it's executing a job. Filtering only `.online` means a fleet of all-busy runners returns `aggregateStatus == .allOffline`. If `.busy` should contribute to the online count (which seems correct), the filter should be:
```swift
runners.filter { $0.status == .online || $0.status == .busy }.count
```
No test covers this edge case. This may be intentional (matching old string-based behavior where `"busy"` != `"online"`), but if so it deserves a `// Note:` explaining why.

**⚠️ `Runner.displayStatus` with `.busy` status but `busy: false`**
The switch in `Runner.displayStatus` falls through to `default: break` for `.busy`, then uses `busy ? "active" : "idle"`. So a runner with `status == .busy` and `busy == false` shows as `"idle (CPU: — MEM: —)"`. This is a plausible API inconsistency (status field vs busy flag disagree) and deserves either a test or a `// Note:` explaining the intended behavior.

---

### Missing Tests

- `RunnerModel.statusColor` with `isRunning: true, githubStatus: .busy` — the `resolvedState` returns `.busy` for this combination (via the `githubStatus == .busy` branch) but no test covers it explicitly.
- `RunnerStore.aggregateStatus` with a mix of `.online` and `.busy` runners — especially important given the gap above.

---

### Minor Nits

- `testDisplayStatusDefaultsToOfflineForUnknownStatus` — test name still says "UnknownStatus" generically; could be `testDisplayStatusUnknownStringMapsToOffline` for precision.
- Two `CodingKeys` cases (`name`, `status`) have identical boilerplate doc descriptions — fine for now, noise if the enum grows.

---

### Summary

| Area | Verdict |
|---|---|
| `RunnerStatus` enum design | ✅ Clean, follows existing pattern |
| `ResolvedState` extraction | ✅ Best design choice in the PR |
| `Runner.displayStatus` `.unknown` handling | ✅ Better than spec |
| `lifecycleWarning!` force-unwrap | ✅ Acceptable with comment |
| `RunnerStore.aggregateStatus` `.busy` gap | ⚠️ Possibly wrong — `.busy` runners not counted as online |
| `Runner.displayStatus` `.busy` + `busy: false` | ⚠️ Untested edge case |
| Dead SwiftLint suppressions removed | ✅ All three pairs gone |
| Test coverage | ✅ Good, two gaps noted |
| Docs & intent comments | ✅ Thorough |

PR #1065, opened 2026-05-31
Refactor batch 4: Scope + LogFetcher docs (#1063)

### LINKED ISSUE #1063: Refactor batch 4: RunnerBarCore/Scope + LogFetcher
Part of #1038

### Files
- `Sources/RunnerBarCore/Scope/ScopeEntry.swift`
- `Sources/RunnerBarCore/Scope/ScopePreferencesStore.swift`
- `Sources/RunnerBarCore/Services/LogFetcher.swift`

### Goals
- Fix file header if needed
- Remove any blanket `swiftlint:disable` suppressions — fix violations at source
- Add `///` doc comments to all public/internal declarations
- Rename any unclear local variables
- No logic changes

eonist:
## **User description**
Closes #1063
Part of #1038

### `ScopeEntry.swift` — no changes needed
Already fully documented with correct header, `///` on all declarations, and clear init docs.

### `ScopePreferencesStore.swift` — no changes needed
Already fully documented. All getters/setters have `///` comments, key builder is explained, cleanup contract is documented.

### `LogFetcher.swift` — docs improved
- `unzipBinaryPath`: weak one-liner replaced with a proper doc explaining *why* the path is hardcoded (always available on macOS)
- `ghRaw(_:timeout:)`: added full `///` doc explaining the transport-shim indirection, the S3 redirect behaviour, and parameter/return semantics
- `fetchActionLogs(group:)`: added `///` doc covering the parallel dispatch strategy, the `OSAllocatedUnfairLock` guard, sort ordering, and return semantics
- `unzipLogs(_:)`: expanded doc to cover the temp-dir lifecycle (`defer` cleanup), the `unzip -q` invocation, and the `[]` failure cases

No logic changes in any file.

___

## **CodeAnt-AI Description**
**Expand the log fetching docs so their behavior is clearer**

### What Changed
- Added clearer explanations for how log fetching works, including when it uses the system unzip tool and how GitHub log redirects are handled
- Documented the expected inputs, return values, and failure cases for single-job log fetches, grouped action log fetches, and ZIP extraction
- Clarified that grouped logs are fetched in parallel, combined in a consistent order, and cleaned up after temporary extraction

### Impact
`✅ Easier to understand log retrieval behavior`
`✅ Clearer failure expectations for log fetching`
`✅ Simpler maintenance of log-related code`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

### DIFF
--- a/Sources/RunnerBarCore/Services/LogFetcher.swift
+++ b/Sources/RunnerBarCore/Services/LogFetcher.swift
@@ -4,12 +4,21 @@ import Foundation
 import os
 
 // MARK: - Filesystem path constants
-/// The unzipBinaryPath constant.
+
+/// Absolute path to the system `unzip` binary, always present on macOS.
 private let unzipBinaryPath = "/usr/bin/unzip"
 
-/// Fetches raw bytes from a GitHub API endpoint via URLSession.
-/// Log endpoints 302-redirect to S3; URLSession follows the redirect automatically.
-/// Returns nil when no token is available or the request fails.
+// MARK: - Transport shim
+
+/// Fetches raw bytes from a GitHub API endpoint via the configured transport.
+///
+/// Delegates to `ghRawTransport()` so the underlying mechanism (URLSession or
+/// `gh` CLI) can be swapped without touching call sites.
+/// Log endpoints 302-redirect to S3; the transport follows the redirect automatically.
+/// - Parameters:
+///   - endpoint: A relative GitHub REST path, e.g. `"repos/owner/repo/actions/jobs/123/logs"`.
+///   - timeout: Maximum seconds to wait for a response. Defaults to `60`.
+/// - Returns: Raw response bytes, or `nil` when no token is available or the request fails.
 private func ghRaw(_ endpoint: String, timeout: TimeInterval = 60) -> Data? {
     ghRawTransport()(endpoint)
 }
@@ -17,7 +26,14 @@ private func ghRaw(_ endpoint: String, timeout: TimeInterval = 60) -> Data? {
 // MARK: - Job log (plain text, 1 call)
 
 /// Fetches the full plain-text log for a single job.
-/// `/actions/jobs/{id}/logs` 302-redirects to a short-lived S3 URL; gh follows it.
+///
+/// `/actions/jobs/{id}/logs` 302-redirects to a short-lived S3 URL; the transport follows it.
+/// Returns `nil` when `scope` is not in `owner/repo` form, the request fails,
+/// or the response body looks like a JSON error object (starts with `"{"`)
+/// - Parameters:
+///   - jobID: The GitHub Actions job ID.
+///   - scope: The `owner/repo` string identifying the repository.
+/// - Returns: Plain-text log content, or `nil` on failure.
 public func fetchJobLog(jobID: Int, scope: String) -> String? {
     guard scope.contains("/") else { return nil }
     guard let data = ghRaw("repos/\(scope)/actions/jobs/\(jobID)/logs"),
@@ -29,7 +45,14 @@ public func fetchJobLog(jobID: Int, scope: String) -> String? {
 // MARK: - Action logs (ZIP per run, N calls)
 
 /// Fetches and concatenates all job logs for every run in a group.
-/// Each run: 1 API call → ZIP → extract → read .txt files.
+///
+/// Issues one API call per run in parallel on `DispatchQueue.global(qos: .userInitiated)`.
+/// Each call retrieves a ZIP archive, extracts all `.txt` log files via `unzipLogs(_:)`,
+/// and appends the results to a shared accumulator guarded by `OSAllocatedUnfairLock`.
+/// The final output is sorted by filename so logs appear in a consistent order.
+/// - Parameter group: The `WorkflowActionGroup` whose runs should be fetched.
+/// - Returns: A single concatenated string with `=== <name> ===` section headers,
+///   or `nil` if `scope` is invalid, `runs` is empty, or all fetches fail.
 public func fetchActionLogs(group: WorkflowActionGroup) -> String? {
     let scope = group.repo
     guard scope.contains("/") else { return nil }
@@ -58,6 +81,14 @@ public func fetchActionLogs(group: WorkflowActionGroup) -> String? {
 // MARK: - ZIP extraction (uses /usr/bin/unzip — always available on macOS)
 
 /// Extracts all `.txt` files from a ZIP blob and returns `(name, text)` pairs.
+///
+/// Writes the ZIP to a unique temporary directory, runs `/usr/bin/unzip -q`,
+/// then enumerates the output directory for `.txt` files. The temporary directory
+/// is always removed on return via `defer`.
+/// - Parameter zipData: Raw ZIP archive bytes as returned by the GitHub logs API.
+/// - Returns: An array of `(name, text)` tuples where `name` is the relative path
+///   inside the archive (without `.txt` extension) and `text` is the file content.
+///   Returns `[]` if the write, unzip, or enumeration step fails.
 public func unzipLogs(_ zipData: Data) -> [(name: String, text: String)] {
     let fileManager = FileManager.default
     let tmp = fileManager.temporaryDirectory.appendingPathComponent(UUID().uuidString)
--- a/Sources/RunnerBarCore/Services/LogFetcher.swift
+++ b/Sources/RunnerBarCore/Services/LogFetcher.swift
@@ -16,8 +16,8 @@ private let unzipBinaryPath = "/usr/bin/unzip"
 /// `gh` CLI) can be swapped without touching call sites.
 /// Log endpoints 302-redirect to S3; the transport follows the redirect automatically.
 /// - Parameter endpoint: A relative GitHub REST path, e.g. `"repos/owner/repo/actions/jobs/123/logs"`.
-/// - Note: The `timeout` parameter is accepted for future use but is not yet forwarded
-///   to the underlying transport; the transport applies its own default timeout.
+/// - Parameter timeout: Reserved for transport implementations that support request timeouts.
+///   The current transport shim does not consume this value.
 /// - Returns: Raw response bytes, or `nil` when no token is available or the request fails.
 private func ghRaw(_ endpoint: String, timeout: TimeInterval = 60) -> Data? {
     ghRawTransport()(endpoint)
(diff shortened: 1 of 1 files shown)


eonist:
## Deep Review — PR #1065

### Summary verdict

This is a docs-only PR with zero logic changes and a single file touched (`LogFetcher.swift`). The issue checklist for `ScopeEntry.swift` and `ScopePreferencesStore.swift` is closed with a "no changes needed" note, which is the right call if those files were already clean. The docs added are accurate and genuinely useful. A few things to call out:

---

### ✅ What's done well

**`ghRaw` doc — transport shim indirection explained.** The previous one-liner just described the return value. The new doc explains *why* the function exists (swappable transport) and that `timeout` is currently unused at the call site (the transport ignores it). That's the kind of non-obvious behaviour that saves future readers from digging into `ghRawTransport()`.

**`fetchActionLogs` — concurrency contract documented.** Calling out `DispatchQueue.global(qos: .userInitiated)`, the `OSAllocatedUnfairLock` guard, and the sort ordering are all things that aren't visible from the function signature. Good.

**`unzipLogs` — lifecycle and failure modes documented.** `defer`-based temp dir cleanup, the `unzip -q` invocation, and the three failure cases for `[]` are all now surfaced. Correct and useful.

---

#### 1. `ghRaw` — `timeout` parameter is documented but silently ignored

```swift
/// - timeout: Maximum seconds to wait for a response. Defaults to `60`.
private func ghRaw(_ endpoint: String, timeout: TimeInterval = 60) -> Data? {
    ghRawTransport()(endpoint)  // timeout is not passed through
}
```

The doc says `timeout` controls the maximum wait. But the implementation calls `ghRawTransport()(endpoint)` — `timeout` is never forwarded. This means the doc is actively misleading: callers who read the doc will believe the timeout is honoured, but the transport's own default (or hardcoded value) wins. Either:
- Forward `timeout` to the transport (if the transport supports it), or
- Change the doc to say something like `/// - timeout: Reserved for future use; currently the transport applies its own timeout.`

This is the only real correctness issue in the PR.

#### 2. `fetchJobLog` — missing comma before `- Returns:`

```swift
/// or the response body looks like a JSON error object (starts with `"{"`)\n/// - Parameters:
```

There's a missing blank line separator between the description paragraph and the `- Parameters:` block. Swift-DocC / Xcode Quick Help requires a blank line between the prose description and the parameter block to render correctly. Without it, the parameter section may render as body text rather than a structured parameter list in Quick Help.

#### 3. `unzipLogs` — `name` return value semantics are slightly inaccurate

```swift
/// - Returns: An array of `(name, text)` tuples where `name` is the relative path
///   inside the archive (without `.txt` extension)
```

This is correct only if the actual enumeration code strips the `.txt` extension — worth verifying against the implementation. If the extension is included, the doc is wrong. If it's stripped, clarify whether the path uses `/` or `_` as the separator (GitHub log ZIPs use numeric prefixes like `1_Build.txt`). A concrete example would make this unambiguous: e.g. `"1_Build"` for a file named `1_Build.txt`.

---

### 🟡 Nits

- **`unzipBinaryPath` doc** — `"always present on macOS"` is accurate for the current macOS versions this app targets, but a link to or mention of the minimum deployment target would make this claim auditable. Fine as-is for an internal constant.
- **`fetchActionLogs` — `nil` when `runs` is empty** is documented but the code actually returns `nil` via the `group.runs.isEmpty` guard or equivalent upstream — worth confirming that path is actually there and not just inferred from the `DispatchGroup` logic completing with zero work (which would return an empty string, not `nil`).
- The `// MARK: - Transport shim` added in the diff is for `ghRaw`, which is a private function. It's a minor structural addition — fine, but the MARK label may be slightly misleading since the *actual* transport shim is `ghRawTransport()` in a different file.

---

### Bottom line

One real issue: the `timeout` doc is actively wrong about what the parameter does. Everything else is accurate. Fix the timeout doc, add the missing blank line before `- Parameters:` in `fetchJobLog`, and this is clean to merge.

eonist:
## Review — Batch 4: LogFetcher docs

Docs-only PR, 1 file changed. Fully delivers on #1063's goal for `LogFetcher.swift`. No logic touched, no regressions possible. A few doc accuracy issues worth noting.

---

### 🟡 Worth fixing before merge

#### 1. `ghRaw(_:timeout:)` — `timeout` parameter is documented but not passed through

The new doc says:
> `timeout`: Maximum seconds to wait for a response. Defaults to `60`.

But the implementation is:
```swift
private func ghRaw(_ endpoint: String, timeout: TimeInterval = 60) -> Data? {
    ghRawTransport()(endpoint)
}
```

`timeout` is accepted as a parameter but never forwarded to `ghRawTransport()`. The doc therefore overpromises — a caller passing `timeout: 5` will get no effect. Either:
- Remove `timeout` from the doc (and add a `// TODO:` or `- Note:` explaining it's not yet wired), or
- Add a `- Note: The `timeout` parameter is accepted for API compatibility but not yet forwarded to the underlying transport.`

This is pre-existing, but the new doc now actively misleads callers.

---

#### 2. `fetchActionLogs` — return doc says `nil` when `runs` is empty, but the code returns `""`

The new doc says:
> Returns … `nil` if `scope` is invalid, `runs` is empty, or all fetches fail.

Check the actual implementation path for an empty `runs` array — if `group.runs` is empty the `DispatchGroup` still completes and the lock-guarded accumulator returns an empty string `""`, not `nil`. If that's the case the doc is wrong on that specific case. Worth verifying and correcting to match actual behaviour.

---

### ✅ What's done well

- **`unzipBinaryPath` doc** — the old one-liner `/// The unzipBinaryPath constant.` was pure noise. The replacement (`/// Absolute path to the system unzip binary, always present on macOS.`) explains the *why* behind the hardcoded path, which is exactly what belongs in a doc comment.
- **`fetchActionLogs` parallel dispatch note** — documenting `DispatchQueue.global(qos: .userInitiated)`, the `OSAllocatedUnfairLock` guard, and the sort ordering is genuinely useful for anyone debugging a race or understanding the output order. Good level of detail.
- **`unzipLogs` temp-dir lifecycle** — calling out the `defer` cleanup and the `[]` failure cases is exactly the contract a caller needs.
- **Transport shim indirection** in `ghRaw` doc — explaining that the underlying mechanism can be swapped without touching call sites is valuable architectural context.
- Correct DocC tick-quoting throughout (`\`ghRawTransport()\``, `` `nil` ``, etc.).
- `fetchJobLog` missing-parameter-line fix (the `- Parameters:` block was absent before) is a clean addition.

---

**Verdict:** One doc actively misleads (`timeout` not forwarded), one may be factually wrong (empty `runs` → `nil` vs `""`). Both are in the new text this PR adds, so they should be corrected here rather than in a follow-up. Otherwise clean to merge.

eonist:
## Review: approve with one minor nit

Documentation-only change to `LogFetcher.swift` — no logic touched. Every behavioral claim in the new docstrings was verified against the implementation. 👍

### Accurate (verified against code)
- **`fetchActionLogs`** — parallel per-run dispatch on `DispatchQueue.global(qos: .userInitiated)`, accumulator guarded by `OSAllocatedUnfairLock`, output `.sorted { $0.name < $1.name }`. All match.
- **`unzipLogs`** — temp-dir write, `/usr/bin/unzip -q`, `.txt` enumeration, name via `deletingPathExtension().path`, and `defer` cleanup all confirmed.
- **`fetchJobLog`** — the "JSON error object (starts with `{`)" note maps to the real `if text.hasPrefix("{") { return nil }` guard.
- Return-on-failure paths (`[]` on write/unzip/enumeration failure) all match.

### One issue worth fixing before merge
`ghRaw` now documents a `timeout` parameter ("Maximum seconds to wait… Defaults to `60`"), but the parameter is **dead** — the body is just `ghRawTransport()(endpoint)`, which never receives `timeout`. The docstring now asserts a guarantee the code doesn't provide. (Pre-existing unused param, but the new doc makes it misleading.)

Pick one:
- Drop the `timeout` line from the docstring (matches current behavior), or
- Forward it: `ghRawTransport()(endpoint, timeout)` — only if the transport signature supports it.

Suggest the doc-only fix here since this is a docs PR; wiring the timeout is a behavioral change for its own commit.

### Non-blocking
`name` being the archive-relative path minus `.txt` is described precisely — no change needed.

eonist:
## Review — PR #1065

Clean docs-only batch overall, but two of the new docs overstate what the code actually guarantees.

---

### 🟡 Should fix

**1. `ghRaw(_:timeout:)` documents a timeout that is ignored**

The new doc says `timeout` is the "Maximum seconds to wait for a response," but the implementation never uses the parameter:

```swift
private func ghRaw(_ endpoint: String, timeout: TimeInterval = 60) -> Data? {
    ghRawTransport()(endpoint)
}
```

Since the issue explicitly says "No logic changes," either remove the timeout parameter doc or describe it as reserved/currently unused by this shim. Suggested wording:

```swift
/// - Parameter timeout: Reserved for transport implementations that support request timeouts.
///   The current transport shim does not consume this value.
```

**2. `fetchActionLogs(group:)` overclaims deterministic ordering**

The new doc says the final output is "sorted by filename so logs appear in a consistent order." That is only true when filenames are unique. Fetches append concurrently into a lock-protected array, and the sort key is only `name` — so duplicate filenames across runs inherit nondeterministic append order. For a docs-only PR, soften this to:

```swift
/// The final output is sorted by filename for stable ordering when names are unique.
```

If true determinism across repeated filenames matters, the sort should carry `runID` as a tiebreaker — but that would be a logic change outside this batch's stated scope.

---

### 🟢 Nice to fix

**3. `fetchActionLogs` is synchronous — the doc should say so**

The function dispatches per-run work in parallel but immediately calls `dispatchGroup.wait()`, so it blocks the calling thread until all fetches complete. The current doc emphasises the parallel dispatch without noting the synchronous return contract. Add:

```swift
/// This function is synchronous: it blocks the calling thread until all per-run fetches complete.
```

That matters for UI call sites — this should not be called from the main thread for groups with many runs or slow log downloads.

**4. `unzipBinaryPath` says "always present on macOS"**

Absolute wording for a shelled-out system binary. Suggest:

```swift
/// Absolute path to the system `unzip` binary expected on supported macOS versions.
```

---

### ✅ Looks good

The scope files (`ScopeEntry.swift`, `ScopePreferencesStore.swift`) are correctly left untouched — both already have the right `RunnerBarCore` header and `///` docs on exposed declarations. The `LogFetcher.swift` additions are directionally useful, especially around the transport shim lifecycle, GitHub log redirect handling, ZIP extraction, and temp-file cleanup. No logic has changed. Items 1–2 are the only things worth fixing before merge.

eonist:
## Review follow-up — verdict after checking branch

**Finding 1 (`timeout` doc) — ✅ confirmed real, now fixed.** Replaced the misleading `- timeout:` parameter line with a `- Note:` explaining it's accepted for future use but not yet forwarded to the transport. Good call.

**Finding 2 (`fetchActionLogs` empty runs → `nil`) — ✅ false alarm.** There's an explicit `guard !runIDs.isEmpty else { return nil }` in the implementation, so the doc was accurate. Retracted.

**Finding 3 (`unzipLogs` name semantics) — not a real issue.** The `deletingPathExtension().path` strip is real and a concrete example (`"1_Build"` for `1_Build.txt`) was added to make it unambiguous. No change needed there.

**PR is clean. Ready to merge when CI is green.**

Issue #477, opened 2026-05-18
Theory A: shell() timeout ignored — allWorkerMetrics() hangs forever blocking entire poll


eonist:
## Theory

`allWorkerMetrics()` in `RunnerMetrics.swift` calls:

```swift
let output = shell("ps aux", timeout: 5)
```

But `Shell.swift` ignores the timeout parameter entirely:

```swift
// timeout is intentionally ignored — Shell.run uses Process.waitUntilExit.
func shell(_ command: String, timeout _: TimeInterval = 20) -> String {
    Shell.run(command).output
}
```

`Shell.run` calls `process.waitUntilExit()` with no deadline. If `ps aux` hangs — which macOS can do when a process is in an uninterruptible wait state, or zsh startup is slow — the entire background closure in `RunnerStore.fetch()` blocks forever. `buildJobState` and `buildGroupState` never run. `onChange` never fires.

Introduced by commit [5f5e9cb](https://github.com/eoncode/runner-bar/commit/5f5e9cbdb811372bc9208a0e77c77468a4b1990a) which renamed `timeout` to `_`, silently killing enforcement.

## Evidence
- `fetchRunners` logs immediately ✅
- `allWorkerMetrics()` is called next inside `fetchAndEnrichRunners()`, same closure
- `buildJobState` never logs — meaning `fetchAndEnrichRunners()` never returned
- No `onChange fired` log anywhere

## Fix

Implement a real timeout in `Shell.run` using `DispatchQueue` + `process.terminate()`:

```swift
static func run(_ command: String, timeout: TimeInterval = 20) -> Result {
    let process = makeProcess(command)
    let (outPipe, errPipe) = attachPipes(to: process)
    do { try process.run() } catch {
        return Result(output: error.localizedDescription, exitCode: -1)
    }
    let didFinish = DispatchSemaphore(value: 0)
    DispatchQueue.global().async {
        process.waitUntilExit()
        didFinish.signal()
    }
    if didFinish.wait(timeout: .now() + timeout) == .timedOut {
        process.terminate()
        return Result(output: "", exitCode: -1)
    }
    return Result(output: readOutput(from: outPipe), exitCode: process.terminationStatus)
}
```

Parent: #476

PR #1066, opened 2026-05-31
Refactor batch 5: ProcessRunner + Utilities docs (#1064)

### LINKED ISSUE #1064: Refactor batch 5: ProcessRunner + RunnerBarCore/Utilities
Part of #1038

### Files
- `Sources/RunnerBarCore/Services/ProcessRunner.swift`
- `Sources/RunnerBarCore/Utilities/GHBinaryLocator.swift`
- `Sources/RunnerBarCore/Utilities/Logger.swift`
- `Sources/RunnerBarCore/Utilities/SystemStats.swift`

### Goals
- Fix file header if needed
- Remove any blanket `swiftlint:disable` suppressions — fix violations at source
- Add `///` doc comments to all public/internal declarations
- Rename any unclear local variables
- No logic changes

### Note
Completes the entire `RunnerBarCore` target. After this, the `RunnerBar` UI layer begins.

eonist:
## **User description**
Closes #1064
Part of #1038

### `ProcessRunner.swift`
- `Result` struct doc: replaced `/// The collected output and exit status from a subprocess invocation.` (was missing entirely)
- `Result.data`: replaced noise doc `/// The data constant.` → `/// Collected stdout bytes, or \`nil\` if the process failed to launch or produced no output.`
- `Result.exitCode`: replaced noise doc `/// The exitCode constant.` → `/// Process exit code. \`Int32.max\` indicates a launch failure rather than a process exit.`
- All other docs (shell-injection migration note, `#477` timeout warning, `run` parameters) were already excellent — no changes

### `Logger.swift`
- Fixed file header: `// RunnerBar` → `// RunnerBarCore` (file lives in the `RunnerBarCore` target)
- `logger` constant: replaced `/// The logger constant.` → `/// Shared \`os.Logger\` instance used by all \`log(_:)\` call sites in RunnerBarCore.`
- `log(_:file:line:)` doc was already complete — no changes

### `SystemStats.swift`
- Fixed file header: `// RunnerBar` → `// RunnerBarCore`
- All property, init, computed var, and `zero` docs were already fully documented — no other changes

### `GHBinaryLocator.swift`
- Listed in issue #1064 but does not exist in the repository — removed from scope. Issue body updated accordingly.

No logic changes in any file.

___

## **CodeAnt-AI Description**
**Clarify process, logging, and system stats documentation**

### What Changed
- Replaced vague process result comments with clear descriptions of collected output and exit codes
- Updated logging and system stats file headers to match the correct target name
- Reworded the shared logger comment to explain its role across RunnerBarCore

### Impact
`✅ Clearer API docs`
`✅ Easier code navigation`
`✅ Fewer confusing comments`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **Internal Improvements**
  * Enhanced error handling documentation for process execution, clarifying failure scenarios and stderr behavior
  * Updated branding references in internal documentation
  * System statistics properties are now immutable for improved data integrity

### DIFF
--- a/Sources/RunnerBarCore/Services/ProcessRunner.swift
+++ b/Sources/RunnerBarCore/Services/ProcessRunner.swift
@@ -27,11 +27,11 @@ import Foundation
 /// approach guarantees termination within `timeout` seconds even in that case.
 /// Do NOT replace this with a bare `waitUntilExit()` call.
 public enum ProcessRunner {
-    /// A value type representing Result.
+    /// The collected output and exit status from a subprocess invocation.
     public struct Result {
-        /// The data constant.
+        /// Collected stdout bytes, or `nil` if the process failed to launch or produced no output.
         public let data: Data?
-        /// The exitCode constant.
+        /// Process exit code. `Int32.max` indicates a launch failure rather than a process exit.
         public let exitCode: Int32
         /// Convenience: decoded UTF-8 string of `data`, or empty string.
         public var output: String { data.flatMap { String(data: $0, encoding: .utf8) } ?? "" }
--- a/Sources/RunnerBarCore/Services/ProcessRunner.swift
+++ b/Sources/RunnerBarCore/Services/ProcessRunner.swift
@@ -48,7 +48,8 @@ public enum ProcessRunner {
     ///   - arguments: Command-line arguments.
     ///   - stdin: Optional data written to the process's standard input.
     ///   - workingDirectory: Optional working directory for the process.
-    ///   - mergeStderr: When true, stderr is merged into stdout (default false).
+    ///   - mergeStderr: When `true`, stderr is merged into stdout.
+    ///     When `false` (default), stderr is discarded — it is not captured or returned.
     ///   - timeout: Seconds before the process is force-terminated.
     /// - Returns: A `Result` with the collected stdout data and exit code.
     ///   `exitCode` is `Int32.max` on process-launch failure.
@@ -67,7 +68,10 @@ public enum ProcessRunner {
 
         let outPipe = Pipe()
         task.standardOutput = outPipe
-        task.standardError = mergeStderr ? outPipe : Pipe()
+        // When mergeStderr is false, use nullDevice so stderr is cleanly discarded.
+        // A throwaway Pipe() would fill its buffer on verbose stderr output and
+        // block the child process indefinitely.
+        task.standardError = mergeStderr ? outPipe : FileHandle.nullDevice
 
         // Wire stdin pipe when body data is provided.
         let inputPipe: Pipe?
@@ -79,6 +83,12 @@ public enum ProcessRunner {
             inputPipe = nil
         }
 
+        // nonisolated(unsafe): Swift 6 concurrency workaround — all concurrent reads/writes
+        // are serialised through `lock`. The final read after `readabilityHandler = nil` is
+        // safe because no other thread can access `outputData` at that point.
+        // TODO: replace readabilityHandler + NSLock with readDataToEndOfFile() after
+        // waitUntilExit() — streaming is unnecessary given the background-thread call
+        // contract. Tracked in <issue link once created>.
         nonisolated(unsafe) var outputData = Data()
         let lock = NSLock()
         outPipe.fileHandleForReading.readabilityHandler = { handle in
--- a/Sources/RunnerBarCore/Services/ProcessRunner.swift
+++ b/Sources/RunnerBarCore/Services/ProcessRunner.swift
@@ -29,7 +29,10 @@ import Foundation
 public enum ProcessRunner {
     /// The collected output and exit status from a subprocess invocation.
     public struct Result {
-        /// Collected stdout bytes, or `nil` if the process failed to launch or produced no output.
+        /// Collected stdout bytes, or `nil` when the process failed to launch
+        /// or when the process ran successfully but produced no stdout.
+        /// - Note: `nil` does not imply failure — use `exitCode` to distinguish
+        ///   a launch failure (`Int32.max`) from a successful process that produced no output.
         public let data: Data?
         /// Process exit code. `Int32.max` indicates a launch failure rather than a process exit.
         public let exitCode: Int32
--- a/Sources/RunnerBarCore/Utilities/Logger.swift
+++ b/Sources/RunnerBarCore/Utilities/Logger.swift
@@ -1,5 +1,5 @@
 // Logger.swift
-// RunnerBar
+// RunnerBarCore
 import Foundation
 import os
 
@@ -10,7 +10,7 @@ import os
 // subsystem/category filtering and are zero-cost in release builds
 // (.debug level is compiled out by the OS when not actively streaming).
 
-/// The logger constant.
+/// Shared `os.Logger` instance used by all `log(_:)` call sites in RunnerBarCore.
 private let logger = Logger(
     subsystem: "com.eoncode.runner-bar",
     category: "general"
--- a/Sources/RunnerBarCore/Utilities/SystemStats.swift
+++ b/Sources/RunnerBarCore/Utilities/SystemStats.swift
@@ -1,5 +1,5 @@
 // SystemStats.swift
-// RunnerBar
+// RunnerBarCore
 import Foundation
 
 /// Snapshot of CPU and memory metrics sampled at a point in time.
--- a/Sources/RunnerBarCore/Utilities/SystemStats.swift
+++ b/Sources/RunnerBarCore/Utilities/SystemStats.swift
@@ -4,16 +4,16 @@ import Foundation
 
 /// Snapshot of CPU and memory metrics sampled at a point in time.
 public struct SystemStats: Sendable {
-    /// CPU usage as a fraction in 0…100 (percentage).
-    public var cpuPct: Double
+    /// CPU usage percentage (0–100).
+    public let cpuPct: Double
     /// Used memory in gigabytes.
-    public var memUsedGB: Double
+    public let memUsedGB: Double
     /// Total physical memory in gigabytes.
-    public var memTotalGB: Double
+    public let memTotalGB: Double
     /// Used disk space in gigabytes.
-    public var diskUsedGB: Double
+    public let diskUsedGB: Double
     /// Total disk space in gigabytes.
-    public var diskTotalGB: Double
+    public let diskTotalGB: Double
 
     /// Creates a new instance.
     public init(


eonist:
## Deep Review — PR #1066

### Summary verdict

XS-sized PR, docs and headers only, no logic changes. All three doc replacements are accurate against the actual implementation. Clean to merge.

---

### ✅ What's done well

**`Result.data` doc** — `"or nil if the process failed to launch or produced no output"` is exactly right. The implementation returns `Result(data: nil, exitCode: Int32.max)` on launch failure, and `Result(data: outputData.isEmpty ? nil : outputData, ...)` on normal exit. Both nil paths are covered.

**`Result.exitCode` doc** — `"Int32.max indicates a launch failure rather than a process exit"` matches the catch block precisely. Useful distinction since `Int32.max` (2147483647) would otherwise look like a plausible but confusing exit code to a caller.

**`logger` constant doc** — `"Shared os.Logger instance used by all log(_:) call sites in RunnerBarCore"` is accurate and more useful than the old `"The logger constant."` The subsystem/category `// comment` block above it already explains the zero-cost debug level detail, so the doc doesn't need to repeat it.

**Header fixes** — `// RunnerBar` → `// RunnerBarCore` in both `Logger.swift` and `SystemStats.swift` is a correct hygiene fix. `ProcessRunner.swift` already had the right header.

**`GHBinaryLocator.swift` scoping** — correctly dropped from scope when the file doesn't exist. No phantom changes.

---

### Nothing to block merge

The only thing worth noting for future batches: `Result.data`'s doc says `"or nil ... if the process produced no output"` — which is accurate (`outputData.isEmpty ? nil : outputData`), but a caller might find it surprising that a zero-byte stdout is indistinguishable from a launch failure when reading `data` alone. The `exitCode` disambiguates (`Int32.max` = launch failure, `0` = success with no output), but that contract lives across two properties with no cross-reference. A `/// - Note: Use exitCode to distinguish launch failure (Int32.max) from a successful process that produced no output.` on `data` would make this fully self-contained. Minor — not a blocker.

Ready to merge. 🟢

eonist:
## Review — Batch 5: ProcessRunner + Utilities docs

XS PR, 6+6, 3 files, docs and headers only. No logic touched.

---

### Findings

#### 1. `Result.data` doc — `nil` on "no output" conflates two distinct states

```swift
/// Collected stdout bytes, or `nil` if the process failed to launch or produced no output.
```

This bundles two meaningfully different cases into one `nil`:
- Launch failure (`exitCode == Int32.max`) — something went wrong before the process ran
- Process ran fine but wrote nothing to stdout — a valid, successful empty output

A caller trying to distinguish "did this fail?" from "did this produce output?" has to also read `exitCode`. The doc as written implies both cases are equivalent failures, which isn't accurate. Worth splitting: e.g. `/// Collected stdout bytes, or \`nil\` when the process failed to launch. Empty output from a successful process is represented as \`nil\` rather than \`Data()\`.` — or, if `Data()` is actually possible for empty-but-successful, document that instead. Either way the current phrasing conflates the two.

This is a pre-existing semantic ambiguity surfaced by the new doc, not introduced by this PR — but since the doc is being written now, it's the right moment to nail the contract.

---

#### 2. `Result` struct doc — minor: doesn't mention `output` convenience property

The struct-level doc (`/// The collected output and exit status from a subprocess invocation.`) is a clear improvement. The existing `output` convenience var already has its own doc (`/// Convenience: decoded UTF-8 string of \`data\`, or empty string.`), so no change needed there — just noting it's complete.

---

### ✅ What's done well

- **Header fixes** (`// RunnerBar` → `// RunnerBarCore`) in both `Logger.swift` and `SystemStats.swift` — straightforward correctness fix, no ambiguity.
- **`logger` constant doc** — `/// Shared \`os.Logger\` instance used by all \`log(_:\`)\` call sites in RunnerBarCore.` is exactly the right level of detail for a `private` constant: explains scope and role without restating the implementation.
- **`Result.exitCode` doc** — calling out `Int32.max` as a sentinel for launch failure (not a real process exit) is precisely the non-obvious contract a caller needs. Good.
- **Noise-doc replacement** (`/// The data constant.`, `/// The exitCode constant.`, `/// A value type representing Result.`) — all three were pure tautologies, all three are now useful. Clean sweep.
- Correctly scoped: `GHBinaryLocator.swift` excluded after confirming it doesn't exist in the repo. Good hygiene.

---

**Verdict:** One genuine doc accuracy issue (finding 1 — `nil` conflation). Everything else is clean. Verify the actual `data` contract in `ProcessRunner.run` (is `Data()` ever returned for a zero-output success, or is it always `nil`?) and tighten the doc accordingly. Ready to merge after that.

eonist:
## Review addendum — full file audit

I previously only reviewed what the diff changed. Here's a proper audit of all three files for code improvement opportunities, not just docs.

---

### `ProcessRunner.swift` — one code finding

**`NSLock` on a background-only path could be `OSAllocatedUnfairLock`**

```swift
nonisolated(unsafe) var outputData = Data()
let lock = NSLock()
outPipe.fileHandleForReading.readabilityHandler = { handle in
    let chunk = handle.availableData
    guard !chunk.isEmpty else { return }
    lock.lock(); outputData.append(chunk); lock.unlock()
}
```

`NSLock` is an Objective-C object and slightly heavier than `OSAllocatedUnfairLock`. The rest of the codebase already uses `OSAllocatedUnfairLock` (e.g. `iso8601Lock` in `WorkflowActionGroupFetch.swift`, the lock in `fetchActionLogs`). This is a minor consistency nit — both are correct — but worth aligning for uniformity.

**`readabilityHandler` + `readDataToEndOfFile` drain** — already noted as correct. No issue.

---

### `Logger.swift` — clean, nothing to improve

The `URL(fileURLWithPath:).deletingPathExtension().lastPathComponent` approach for extracting the filename is correct and idiomatic. The `privacy: .public` annotations on all interpolated values are intentional and right for a dev tool. No code improvements needed.

---

### `SystemStats.swift` — two findings

**1. `var` stored properties should be `let`**

```swift
public var cpuPct: Double
public var memUsedGB: Double
// ...
```

`SystemStats` is a snapshot — it represents a point-in-time sample and is never mutated after creation. All five stored properties are `var`, which means a caller can silently overwrite `stats.cpuPct = 99` without any compiler warning. They should be `let`. The `zero` static and the `init` both work fine with `let`.

**2. `diskFreePct` is redundant given `diskPressure`**

```swift
public var diskPressure: Double { ... return diskUsedGB / diskTotalGB }
public var diskFreePct: Double  { ... return ((diskTotalGB - diskUsedGB) / diskTotalGB) * 100 }
```

`diskFreePct` is exactly `(1 - diskPressure) * 100`. If both are needed by callers, they should at least be expressed in terms of each other to make the relationship explicit and avoid drift:

```swift
public var diskFreePct: Double {
    guard diskTotalGB > 0 else { return 0 }
    return (1 - diskPressure) * 100
}
```

Same output, but makes it clear the two properties are consistent by construction.

---

### Summary

| File | Finding | Severity |
|---|---|---|
| `ProcessRunner.swift` | `NSLock` → `OSAllocatedUnfairLock` for consistency | Nit |
| `SystemStats.swift` | Stored properties should be `let`, not `var` | ⚠️ Real |
| `SystemStats.swift` | `diskFreePct` should derive from `diskPressure` | Nit |

The `var` → `let` change on `SystemStats` is the only one worth acting on before merge — it's a public API correctness issue for a type that is semantically immutable.

eonist:
## Supplemental review — full file read (code + docs)

Apologies for the first pass — I reviewed only the diff. After reading all three files in full, here's what the code side looks like.

---

### `ProcessRunner.swift` — code observations

#### 1. `NSLock` used instead of `OSAllocatedUnfairLock`

```swift
nonisolated(unsafe) var outputData = Data()
let lock = NSLock()
outPipe.fileHandleForReading.readabilityHandler = { handle in
    let chunk = handle.availableData
    guard !chunk.isEmpty else { return }
    lock.lock(); outputData.append(chunk); lock.unlock()
}
```

The rest of the codebase (e.g. `WorkflowActionGroupFetch.swift`, `LogFetcher.swift`) uses `OSAllocatedUnfairLock` consistently. `NSLock` works fine but is inconsistent with the established pattern and slightly heavier. Worth aligning, especially since this is a hot path called from `readabilityHandler` on a background thread.

#### 2. `inputPipe` binding is more complex than needed

```swift
let inputPipe: Pipe?
if stdin != nil {
    let p = Pipe()
    task.standardInput = p
    inputPipe = p
} else {
    inputPipe = nil
}
```

This can be simplified to:
```swift
let inputPipe: Pipe? = stdin.map { _ in
    let p = Pipe()
    task.standardInput = p
    return p
}
```
or just an `if let stdin` block that avoids the separate binding entirely. Minor, but the current form is more verbose than idiomatic Swift.

#### 3. `Result.data` nil-for-empty contract confirmed — doc finding 1 from previous comment still stands

From the return site:
```swift
return Result(data: outputData.isEmpty ? nil : outputData, exitCode: exitCode)
```

So `data` is `nil` for *both* launch failure (`exitCode == Int32.max`) and successful-but-empty stdout (`exitCode == 0`). The `output` convenience var already papers over this (`?? ""`), but the doc saying `"or nil if the process failed to launch or produced no output"` is accurate — it just buries the distinction. Given the explicit `Int32.max` sentinel already documented on `exitCode`, the `data` doc should at minimum say: `/// Collected stdout bytes, or \`nil\` on launch failure or empty output. Check \`exitCode\` to distinguish the two.`

---

### `Logger.swift` — clean

No code issues. The `URL(fileURLWithPath:).deletingPathExtension().lastPathComponent` for stripping the file path to just the filename is correct and idiomatic. `log(_:file:line:)` doc is excellent as-is.

---

### `SystemStats.swift` — one minor observation

#### 4. `diskFreePct` is redundant given `diskPressure`

```swift
/// Disk usage as a fraction in 0…1.
public var diskPressure: Double { ... return diskUsedGB / diskTotalGB }

/// Percentage of disk space that is FREE (0–100).
public var diskFreePct: Double { ... return ((diskTotalGB - diskUsedGB) / diskTotalGB) * 100 }
```

`diskFreePct` is `(1 - diskPressure) * 100`. Not a bug, but if both are used in the UI it's worth a doc note linking them: `/// Equivalent to \`(1 - diskPressure) * 100\`.` — makes it obvious they stay in sync and that there's no independent calculation.

#### 5. `cpuPct` doc says "0…100" but there's no clamping

```swift
/// CPU usage as a fraction in 0…100 (percentage).
public var cpuPct: Double
```

The doc promises `0…100` but it's a plain `var` with no enforcement. If the sampling source (wherever `SystemStats` gets populated) ever returns a value outside that range (common with `host_processor_info` on brief spikes), the UI could display `> 100%`. Worth either adding a clamped computed var, or noting in the doc that the range is *expected* not *enforced*.

---

### Updated verdict

| | |
|---|---|
| 🟡 Code | Align `NSLock` → `OSAllocatedUnfairLock` in `ProcessRunner` for consistency |
| 🟡 Code | Simplify `inputPipe` binding |
| 🟡 Doc | `Result.data` nil doc — add pointer to `exitCode` to distinguish the two nil causes |
| Nit | `diskFreePct` doc: note it equals `(1 - diskPressure) * 100` |
| Nit | `cpuPct` doc: clarify range is expected, not enforced |

None of these block merge, but findings 1–3 are genuine improvements worth a small follow-up. The PR as scoped (docs + headers) is correct — these are pre-existing code issues the issue tracking (#1063/#1038) may want to pick up.

eonist:
## Deep Review — Refactor Batch 5: ProcessRunner + Utilities

Overall: **solid and merge-ready with minor caveats**, but there are a few things worth calling out beyond what the PR description says.

---

### ✅ What's genuinely good

**`ProcessRunner.swift` — `nil` disambiguation in `Result.data`**

The expanded `nil` note is the most meaningful doc change in this whole PR:

```swift
/// - Note: `nil` does not imply failure — use `exitCode` to distinguish
///   a launch failure (`Int32.max`) from a successful process that produced no output.
```

This is exactly the kind of pitfall that causes silent bugs in callers who write `guard let data = result.data else { return }` and silently swallow a successful-but-silent process. Good catch.

**`FileHandle.nullDevice` over throwaway `Pipe()`**

This is a real bug fix hiding inside a "docs-only" PR. The old code:

```swift
task.standardError = mergeStderr ? outPipe : Pipe()
```

…creates an orphaned `Pipe()` whose read end is never drained. On any process that writes significant stderr, the pipe buffer fills (typically 64 KB on macOS), the write syscall blocks, and the child process hangs indefinitely. Replacing it with `FileHandle.nullDevice` is the correct fix. This deserves a callout in the PR description as a **bug fix**, not just a comment/docs change. It's currently buried and will be invisible in the changelog.

**`SystemStats` `var → let`**

Also a real improvement smuggled in as a docs change. `SystemStats` is a snapshot — it should never be mutated after capture. Promoting all five properties to `let` enforces that at the type level and makes `Sendable` conformance trivially correct (no mutation = no data race risk). Worth noting that this is technically a **source-breaking change** for any call site that does `var stats = SystemStats(...)` and later mutates a field — though for an internal snapshot type that's almost certainly fine.

---

**1. The TODO comment lacks a tracking issue link**

```swift
// TODO: replace readabilityHandler + NSLock with readDataToEndOfFile() after
// waitUntilExit() — streaming is unnecessary given the background-thread call
// contract. Tracked in <issue link once created>.
```

The placeholder `<issue link once created>` should be replaced before merge or immediately after. Unlinked TODOs get lost. Either create the issue now and drop the link in, or remove the TODO until you're ready to track it.

**2. `Int32.max` sentinel is a leaky abstraction — worth a follow-up**

Both the `Result.data` note and the `exitCode` doc reference `Int32.max` as the launch-failure sentinel:

```swift
/// Process exit code. `Int32.max` indicates a launch failure rather than a process exit.
```

This is documented correctly, but the *design* has a smell: `Int32.max` (2147483647) is technically a valid exit code on some platforms. A typed enum like `enum ProcessResult { case success(Data?, Int32); case launchFailure }` would make this unambiguous and eliminate the sentinel entirely. The doc improvement here is good, but it quietly surfaces a design debt. Worth a follow-up issue if one doesn't already exist.

**3. `mergeStderr` doc now has an implicit contradiction**

```
/// - mergeStderr: When `true`, stderr is merged into stdout.
///   When `false` (default), stderr is discarded — it is not captured or returned.
```

This is accurate for the *current* implementation, but callers expecting stderr-capture when `mergeStderr: false` will be surprised. The doc says "discarded" but doesn't mention that there's no way to get stderr separately. Consider adding: `Pass \`true\` to capture it via \`Result.data\`.` — makes the only escape hatch explicit.

**4. `nonisolated(unsafe)` comment is good, but the concurrency contract isn't enforced**

```swift
// nonisolated(unsafe): Swift 6 concurrency workaround — all concurrent reads/writes
// are serialised through `lock`.
```

The comment is accurate and the `NSLock` usage looks correct from the diff context. However, the stated refactor (use `readDataToEndOfFile()` after `waitUntilExit()`) would actually remove the need for both `nonisolated(unsafe)` *and* the lock. The comment + TODO together are fine, but make sure the TODO issue is filed before this gets lost in the backlog.

---

### 📋 Issue #1064 scope compliance

| Goal | Status |
|---|---|
| Fix file headers | ✅ `RunnerBar` → `RunnerBarCore` in Logger + SystemStats |
| Remove blanket `swiftlint:disable` | ✅ None added/present |
| Add `///` doc comments to all public/internal declarations | ✅ All targeted declarations covered |
| Rename unclear local variables | N/A — no renames needed in these files |
| No logic changes | ⚠️ Technically false — `Pipe()→FileHandle.nullDevice` and `var→let` are behavioural changes (both correct) |
| `GHBinaryLocator.swift` | ✅ Correctly removed from scope (file doesn't exist) |

The "no logic changes" claim in the issue goal isn't quite met, but both deviations are improvements, not regressions.

---

### Minor nits

- `SystemStats` doc for `cpuPct` changed from `/// CPU usage as a fraction in 0…100 (percentage).` to `/// CPU usage percentage (0–100).` — the original said "fraction" but the range was 0–100, which is contradictory. The new wording is cleaner. ✅
- Logger file header fix (`RunnerBar` → `RunnerBarCore`) is a trivial but correct hygiene fix.

---

### Summary

This is mostly docs, but it contains **two real fixes** (`FileHandle.nullDevice` pipe drain bug, `var→let` immutability) that should be acknowledged explicitly. The `Int32.max` sentinel is design debt worth a follow-up issue. The unlinked TODO is the only thing I'd want resolved before merge.

eonist:
## Deep Review — PR #1066 (updated diff)

The PR now contains real changes beyond the docs-only first version. Reviewing against issue #1064's goals: fix headers, remove SwiftLint suppressions, add `///` docs, rename unclear locals, no logic changes (except where bugs were found).

---

### ✅ Real bug fix — stderr deadlock

```swift
// before
task.standardError = mergeStderr ? outPipe : Pipe()
// after
task.standardError = mergeStderr ? outPipe : FileHandle.nullDevice
```

This is the most important change in the batch. A throwaway `Pipe()` has a finite kernel buffer (~64 KB on macOS). Any child process that writes more than that to stderr will block waiting for the read-end to drain — but nothing is reading it, so the process hangs indefinitely. `FileHandle.nullDevice` discards cleanly with no buffer. The inline comment explaining exactly why is correct and should stay. Good fix.

---

### ✅ `Result.data` doc — improved but now slightly overclaims

```swift
/// Collected stdout bytes, or `nil` when the process failed to launch
/// or when the process ran successfully but produced no stdout.
/// - Note: `nil` does not imply failure — use `exitCode` to distinguish
///   a launch failure (`Int32.max`) from a successful process that produced no output.
```

The `- Note:` is a good addition and correctly documents the `Int32.max` sentinel. However, look at the actual return site:

```swift
return Result(data: outputData.isEmpty ? nil : outputData, exitCode: exitCode)
```

This means a process that runs successfully and writes one byte returns `data: Data([byte])`, but a process that runs successfully and writes *nothing* returns `data: nil` — which the note says is indistinguishable from a launch failure without checking `exitCode`. The doc is now accurate, but it's documenting a design smell. The follow-up issue you're planning to open (nil vs Data() return semantics) is the right place to fix this. The current doc is the best you can honestly write given the current implementation — no change needed here, just confirming the note is accurate.

---

### ✅ `nonisolated(unsafe)` comment — correct and complete

```swift
// nonisolated(unsafe): Swift 6 concurrency workaround — all concurrent reads/writes
// are serialised through `lock`. The final read after `readabilityHandler = nil` is
// safe because no other thread can access `outputData` at that point.
// TODO: replace readabilityHandler + NSLock with readDataToEndOfFile() after
// waitUntilExit() — streaming is unnecessary given the background-thread call
// contract. Tracked in <issue link once created>.
```

The safety argument is sound: `readabilityHandler = nil` is set before the semaphore wait completes, so by the time the final `outputData` read happens, the handler is torn down and no concurrent write is possible. The TODO correctly identifies `readDataToEndOfFile()` as the right successor — much simpler, no lock needed, and eliminates `nonisolated(unsafe)` entirely. Remember to update `<issue link once created>` once the follow-up issue is open.

---

### ✅ `SystemStats` — `var → let`

All five stored properties correctly changed to `let`. `SystemStats` is a `Sendable` snapshot type — `var` on a `Sendable` struct is a contradiction: it implies mutation is valid after construction, which it isn't here. This also means `SystemStats` now satisfies `Sendable` purely structurally (all stored `let` properties of `Sendable` types) without any `@unchecked` annotation. Clean fix.

The `cpuPct` doc fix (`"fraction in 0…100"` → `"percentage (0–100)"`) is also correct — `memoryPressure` and `diskPressure` correctly say `"fraction in 0…1"`, so now the language is consistent across the type.

---

### 🟡 One loose end — `<issue link once created>`

The TODO in the `nonisolated(unsafe)` comment has a placeholder. Before or immediately after merge, open the follow-up issue and update the comment with the real link. Leaving the placeholder in main is fine temporarily but it should be resolved within the same session.

---

### Not in scope — confirmed correct decisions

- `NSLock → Mutex` deferred ✅ — coupled to the readabilityHandler removal, right call
- `nil vs Data()` semantics deferred ✅ — coupled to the above, right call  
- `stdin != nil` redundancy skipped ✅ — pre-existing style, out of scope
- Logger single-category comment skipped ✅ — design decision, not a bug

---

**Verdict: clean, merge when CI is green.** Update the TODO link post-merge.

eonist:
## Deep Review — Refactor Batch 5: ProcessRunner + Utilities Docs (#1064)

Overall verdict: **✅ Approve with one must-fix and a few worth-discussing notes.** The doc and header fixes are genuinely good. But the PR also sneaked in a real behaviour change and a meaningful correctness fix that deserve explicit acknowledgment — and the former has a semantic gap worth closing.

---

### 1. `ProcessRunner.swift` — behaviour change: `Pipe()` → `FileHandle.nullDevice` (⚠️ must document in PR + changelog)

This is the most important thing in the diff and it is **not doc-only**. The old code:

```swift
task.standardError = mergeStderr ? outPipe : Pipe()
```

The new code:

```swift
task.standardError = mergeStderr ? outPipe : FileHandle.nullDevice
```

The inline comment explains the *why* correctly — a `Pipe()` with an unread read-end will block the child once its kernel pipe buffer fills (~64 KB on macOS). `FileHandle.nullDevice` is the right fix. But:

- **This is a latent bug fix, not a doc change.** Any process that wrote significant stderr would previously have silently deadlocked `waitUntilExit()` (after the timeout, at best). Calling this "no logic changes" in the PR description is inaccurate.
- The in-code comment is excellent. Consider also tracking this as a separate fixed bug (e.g. a `Fixes #XXXX` reference), so the change is discoverable through issue history rather than only through `git blame`.
- **Minor semantic gap:** `FileHandle.nullDevice` discards stderr bytes. If a caller sets `mergeStderr: false` and the process exits non-zero, there is no way to retrieve the error message. Consider whether a future `Result` version should carry `stderrData: Data?` when `mergeStderr == false`, or at least document the tradeoff explicitly at the call site. The current doc already says "stderr is discarded — it is not captured or returned", which is good, but the consequence (you cannot diagnose failures) could be called out in the `- Returns:` block too.

---

### 2. `ProcessRunner.Result.data` doc — small accuracy issue

```swift
/// Collected stdout bytes, or `nil` when the process failed to launch
/// or when the process ran successfully but produced no stdout.
/// - Note: `nil` does not imply failure — use `exitCode` to distinguish
///   a launch failure (`Int32.max`) from a successful process that produced no output.
```

This is a significant improvement over `/// The data constant.`. One nit: the doc says "`nil` … when the process ran successfully but produced no stdout", but the implementation returns `nil` specifically when `outputData.isEmpty`:

```swift
return Result(data: outputData.isEmpty ? nil : outputData, exitCode: exitCode)
```

If a process writes nothing to stdout and exits 0, callers get `data: nil, exitCode: 0`. That is technically what the doc says, but "produced no stdout" is ambiguous — it could mean the pipe was never written to *or* the process wrote zero bytes. Worth being explicit: **"`nil` when stdout was empty (zero bytes), regardless of exit code."** The current wording is fine for human readers but could mislead someone writing a `guard data != nil` check, thinking they're also guarding against launch failure.

---

### 3. `nonisolated(unsafe)` TODO comment — great transparency, one gap

```swift
// TODO: replace readabilityHandler + NSLock with readDataToEndOfFile() after
// waitUntilExit() — streaming is unnecessary given the background-thread call
// contract. Tracked in <issue link once created>.
```

The comment is valuable — it explains the future direction clearly. However, `<issue link once created>` is a placeholder that will likely never get filled in. **Before merging, either:**
- Open the tracking issue now and paste the link, or
- Replace the placeholder with `Tracked: open a follow-up issue` so it reads as intentional rather than forgotten.

Leaving placeholder text in merged code is a maintenance smell.

---

### 4. `SystemStats.swift` — `var` → `let` (another undeclared logic change)

```swift
// before
public var cpuPct: Double
public var memUsedGB: Double
// ...

// after
public let cpuPct: Double
public let memUsedGB: Double
// ...
```

Again: this is **not a doc change**. It changes the public API surface. Any caller that previously mutated a `SystemStats` value (e.g. for mocking/testing purposes by creating a mutable copy and overwriting a field) will now fail to compile. Given that `SystemStats` is `public` and `Sendable`, this is the *correct* design decision — a snapshot struct should be immutable — but it warrants a call-out in the PR description under "API changes" rather than being silent.

Check: does the test suite contain any `var stats = SystemStats(...); stats.cpuPct = 42` patterns? If so those are now broken and must be addressed.

---

### 5. `Logger.swift` header fix — correct, no concerns

`// RunnerBar` → `// RunnerBarCore` is accurate; the file lives in `RunnerBarCore`. No issues. The doc improvement from `/// The logger constant.` to `/// Shared \`os.Logger\` instance used by all \`log(_:)\` call sites in RunnerBarCore.` is a material improvement and correctly describes the singleton role.

---

### 6. `GHBinaryLocator.swift` — removed from scope

The PR description notes this file doesn't exist in the repository. That's fine — removing dead scope is good. But it's worth verifying issue #1064 was updated to reflect this (the PR description says it was). Consider adding a brief comment in the issue noting *why* it was removed (file was deleted in a previous refactor? never existed?), both for future readers and for the batch-5 completeness record.

---

### 7. Doc style consistency — minor

The existing `run` parameter docs use `\`true\`` and `\`false\`` with backticks. The new `mergeStderr` doc follows this:

```
/// - mergeStderr: When `true`, stderr is merged into stdout.
///   When `false` (default), stderr is discarded…
```

That's consistent. ✅ The `Result` struct doc uses `\`nil\`` and `\`Int32.max\`` with backticks throughout — also consistent. ✅

---

### Summary

| Item | Severity | Status |
|---|---|---|
| `Pipe()` → `nullDevice` is a bug fix, not a doc change | ⚠️ Mislabelled | Update PR description |
| `var` → `let` on `SystemStats` is an API change | ⚠️ Mislabelled | Update PR description; verify no test mutations |
| `Result.data` nil semantics slightly ambiguous | 🔵 Minor | Tighten wording |
| `<issue link once created>` placeholder in code | 🔵 Minor | Resolve before merge |
| stderr not capturable on failure with `mergeStderr: false` | 💡 Future consideration | Document or track |

The two behaviour changes are genuinely correct and beneficial — but labelling a PR "no logic changes" when it includes a deadlock fix and a public API mutability change makes it harder to audit in `git log`. Everything else is solid work.

PR #1069, opened 2026-06-01
Batch 6: docs + headers — ScopeEntry, GitHubConstants, GitHubTransportShim (#1068)

### LINKED ISSUE #1068: Batch 6: docs + headers — ScopeEntry, ScopePreferencesStore, GitHubConstants, GitHubTransportShim
Part of #1038 — final batch to complete `RunnerBarCore`.

## Files in scope

- `Sources/RunnerBarCore/Scope/ScopeEntry.swift`
- `Sources/RunnerBarCore/Scope/ScopePreferencesStore.swift`
- `Sources/RunnerBarCore/GitHub/GitHubConstants.swift`
- `Sources/RunnerBarCore/GitHub/GitHubTransportShim.swift`

---

### 1. `ScopeEntry.swift` — 3 changes

- **`scope: var` → `let`** — no call site in the codebase mutates `scope` after construction (confirmed by full grep). `ScopeStore` only creates new entries or removes by `id`. Changed to `let` with a doc note explaining the remove+add contract.
- **`Sendable` conformance added** — all stored properties (`UUID`, `String`, `Bool`) are `Sendable`. Conformance is structurally correct and eliminates potential Swift 6 cross-actor warnings at call sites where `ScopeEntry` crosses actor boundaries.
- **`isEnabled` stays `var`** — legitimately mutated by `ScopeStore.setEnabled(_:_:)`.

### 2. `ScopePreferencesStore.swift` — no changes needed

Already in excellent shape:
- All public methods have clear `///` docs with nil/Bool? contract documented at type level and per property.
- `cleanUp(scope:)` doc correctly cross-references `ScopeStore.remove(id:)`.
- Key naming scheme (`scope.<scope>.<field>`) is documented.
- Issue number traceability via MARK comments.
- File header correct.

**One minor observation (not acted on):** `failureHookEnabled(for:)` returns `Bool` (not `Bool?`) meaning unset = `false`, which is the correct default. But it's inconsistent with `notifyOnSuccess`/`notifyOnFailure` which return `Bool?` for three-state semantics. Worth a future doc note or alignment pass, but not a blocker.

**Verdict: no changes. Marked ✅ as-is.**

### 3. `GitHubConstants.swift` — 1 fix

- **File header `// RunnerBar` → `// RunnerBarCore`** — file lives in `Sources/RunnerBarCore/GitHub/`.
- `NOSONAR` annotations correct and intentional.
- Both constants documented.

### 4. `GitHubTransportShim.swift` — 2 comment fixes

- **Circular lock comments removed** — `// Safety: protected by transportLock` on `transportLock` itself was circular. Replaced with `// Serialises all reads and writes to the active transport closure.`
- **`ghRawTransport()` doc note added** — clarifies it returns the closure itself, not a result. Intentional design (LogFetcher holds the closure for multiple calls) but non-obvious without the note.
- **`import os` confirmed necessary** — `OSAllocatedUnfairLock` is in the `os` module, consistent with all other files using it in the codebase.

---

## Out of scope / follow-up

- **Pre-configure assertion** — `ghAPI` / `ghRawTransport` called before `configureGHAPI` / `configureGHRaw` silently return `nil`. A DEBUG-only `assertionFailure` would catch misconfiguration early. Deferred — behaviour change, deserves its own PR.
- **`failureHookEnabled` Bool vs Bool? alignment** — see `ScopePreferencesStore` note above.

## Summary of changes landed in PR #1069

| File | Changes | Logic? |
|---|---|---|
| `ScopeEntry.swift` | `scope` → `let`; `Sendable` added; doc note | API change (additive + restrictive) |
| `ScopePreferencesStore.swift` | None | — |
| `GitHubConstants.swift` | Header fix | Docs only |
| `GitHubTransportShim.swift` | Lock comments fixed; `ghRawTransport` note | Docs only |

eonist:
Part of #1038. Closes #1068.

Final batch to complete `RunnerBarCore`. 3 files changed, 1 confirmed no changes needed.

### `ScopeEntry.swift` — `scope` `var` → `let` + `Sendable` + doc
- `scope` was `var` with no mutation sites anywhere in the codebase (`ScopeStore` only ever creates new entries or removes by `id` — never reassigns `scope` on an existing entry). Changed to `let`.
- Added `Sendable` conformance — all stored properties are `let`/`var` of `Sendable` types (`UUID`, `String`, `Bool`), so conformance is structurally correct and eliminates any Swift 6 cross-actor warnings at call sites.
- Added doc note on `scope` explaining the immutability contract and the remove+add pattern.
- `isEnabled` remains `var` — legitimately mutated by `ScopeStore.setEnabled(_:_:)`.

### `GitHubConstants.swift` — header fix only
- File header `// RunnerBar` → `// RunnerBarCore` (file lives in `Sources/RunnerBarCore/GitHub/`)
- No other changes.

### `GitHubTransportShim.swift` — lock comment fixes + `ghRawTransport` doc
- Removed circular lock safety comments (`// Safety: protected by transportLock` on `transportLock` itself) — replaced with `// Serialises all reads and writes to…`
- Added `- Note:` to `ghRawTransport()` clarifying it returns the closure itself, not the result of a call — intentional (LogFetcher holds the closure for multiple calls) but previously non-obvious.
- `import os` confirmed necessary — `OSAllocatedUnfairLock` is in the `os` module, consistent with all other files in the codebase.

### `ScopePreferencesStore.swift` — no changes
- Already fully documented with correct headers and clear `nil`/`Bool?` contract docs. Confirmed clean ✅.

## Logic / API changes
- `ScopeEntry.scope`: `var` → `let` — source-level API change. No call site mutates `scope` directly (confirmed by grep across full codebase). `Codable` round-trips unaffected.
- `ScopeEntry: Sendable` — additive conformance, no breaking changes.

## Not in scope
- `ScopeStore.swift` `swiftlint:disable orphaned_doc_comment` — separate batch (RunnerBar UI layer)
- `failureHookEnabled` `Bool` vs `Bool?` consistency in `ScopePreferencesStore` — tracked in #1068 for future consideration
- Pre-configure assertion for unconfigured transport calls — behaviour change, separate issue

### DIFF
--- a/Sources/RunnerBarCore/GitHub/GitHubConstants.swift
+++ b/Sources/RunnerBarCore/GitHub/GitHubConstants.swift
@@ -1,5 +1,5 @@
 // GitHubConstants.swift
-// RunnerBar
+// RunnerBarCore
 import Foundation
 
 // MARK: - Shared GitHub URI constants
--- a/Sources/RunnerBarCore/GitHub/GitHubTransportShim.swift
+++ b/Sources/RunnerBarCore/GitHub/GitHubTransportShim.swift
@@ -27,15 +27,15 @@ public typealias GHRawTransport = @Sendable (_ endpoint: String) -> Data?
 // MARK: - Module-level state
 
 /// The active JSON transport closure.
-/// Safety: protected by transportLock.
+/// Serialises all reads and writes to the active transport closure.
 private let transportLock = OSAllocatedUnfairLock<GHAPITransport>(initialState: { _ in nil })
 
 /// The active raw-bytes transport closure (log endpoints).
-/// Safety: protected by rawTransportLock.
+/// Serialises all reads and writes to the active raw transport closure.
 private let rawTransportLock = OSAllocatedUnfairLock<GHRawTransport>(initialState: { _ in nil })
 
 /// Closure that reports the current rate-limit state.
-/// Safety: protected by rateLimitLock.
+/// Serialises all reads and writes to the rate-limit closure.
 private let rateLimitLock = OSAllocatedUnfairLock<@Sendable () -> Bool>(initialState: { false })
 
 // MARK: - Configuration
@@ -70,6 +70,7 @@ func ghAPI(_ endpoint: String) -> Data? {
 
 /// Returns the configured raw-bytes transport closure.
 /// Used by `LogFetcher` to fetch log data without importing the app target.
+/// - Note: Returns the closure itself, not the result of a call — callers invoke it directly.
 func ghRawTransport() -> GHRawTransport {
     rawTransportLock.withLock { $0 }
 }
--- a/Sources/RunnerBarCore/Scope/ScopeEntry.swift
+++ b/Sources/RunnerBarCore/Scope/ScopeEntry.swift
@@ -17,6 +17,7 @@ public struct ScopeEntry: Identifiable, Codable, Equatable, Hashable, Sendable {
     /// remove the existing entry and add a new one via `ScopeStore`.
     public let scope: String
     /// When `false`, `RunnerStore` skips this scope during polling.
+    /// `var`: toggled by `ScopeStore.setEnabled(_:_:)` — the only intended mutation site.
     public var isEnabled: Bool
 
     /// Creates a new `ScopeEntry` with a fresh random `id`.
--- a/Sources/RunnerBarCore/Scope/ScopeEntry.swift
+++ b/Sources/RunnerBarCore/Scope/ScopeEntry.swift
@@ -9,7 +9,7 @@ import Foundation
 /// `scope` is either `"owner/repo"` (repository) or `"myorg"` (organisation).
 /// `isEnabled` controls whether `RunnerStore` polls this scope; disabled scopes
 /// are retained in the list but silently skipped during fetch.
-public struct ScopeEntry: Identifiable, Codable, Equatable, Hashable {
+public struct ScopeEntry: Identifiable, Codable, Equatable, Hashable, Sendable {
     /// Stable identity for use in SwiftUI lists and `Codable` round-trips.
     public let id: UUID
     /// The GitHub scope string — either `"owner/repo"` or an org name.
--- a/Sources/RunnerBarCore/Scope/ScopeEntry.swift
+++ b/Sources/RunnerBarCore/Scope/ScopeEntry.swift
@@ -13,7 +13,9 @@ public struct ScopeEntry: Identifiable, Codable, Equatable, Hashable {
     /// Stable identity for use in SwiftUI lists and `Codable` round-trips.
     public let id: UUID
     /// The GitHub scope string — either `"owner/repo"` or an org name.
-    public var scope: String
+    /// `let`: the scope string is immutable after construction. To change a scope,
+    /// remove the existing entry and add a new one via `ScopeStore`.
+    public let scope: String
     /// When `false`, `RunnerStore` skips this scope during polling.
     public var isEnabled: Bool
 


eonist:
## Deep Review — Batch 6: ScopeEntry, GitHubConstants, GitHubTransportShim (#1068)

**Verdict: ✅ Approve — ready to merge.** CI is all green (Swift Test ✅, SwiftLint ✅, Periphery ✅, SonarCloud ✅, review ✅; ui-test skipped as expected for a docs/refactor-only branch). Notes below range from "worth knowing" to one minor wording nit.

---

### 1. `ScopeEntry.swift` — `scope: var → let` ✅ correct call

The issue (#1068) explicitly flagged the decision point and the PR made the right call. `scope` has no mutation sites — `ScopeStore` only creates or removes entries, never reassigns `scope` in-place — so `let` is the semantically correct choice.

The new doc comment is good:
```swift
/// `let`: the scope string is immutable after construction. To change a scope,
/// remove the existing entry and add a new one via `ScopeStore`.
```

**One minor nit:** the doc is on the property, which is fine for API consumers reading generated docs. But `ScopeStore` itself doesn't cross-reference this constraint anywhere. If someone extends `ScopeStore` later and tries to add a `rename(id:newScope:)` method, they'll find out at compile time — which is exactly the point of `let`. So this is fine as-is. Just noting the cascade is self-enforcing.

**`isEnabled` stays `var`** — correct. `ScopeStore.setEnabled(_:_:)` mutates it in place via index lookup. Keeping it mutable avoids a copy-remove-reinsert on every toggle.

---

### 2. `GitHubConstants.swift` — header fix ✅ trivial and correct

`// RunnerBar` → `// RunnerBarCore`. File lives in `Sources/RunnerBarCore/GitHub/`. Nothing more to say.

---

### 3. `GitHubTransportShim.swift` — lock comment fix ✅ + closure-return note ✅

**Circular comment fix — correct and meaningful.** The old comments were:
```swift
/// Safety: protected by transportLock.
private let transportLock = OSAllocatedUnfairLock<...>(...)
```
This was genuinely confusing — the comment described the lock as being protected by itself. The new wording:
```swift
/// Serialises all reads and writes to the active transport closure.
```
…accurately describes what the lock *does* rather than circularly referencing its own name. This is a real readability improvement, not cosmetic.

**`ghRawTransport()` closure-return note — correct, necessary.** Without the note:
```swift
/// Returns the configured raw-bytes transport closure.
/// Used by `LogFetcher` to fetch log data without importing the app target.
```
…it reads ambiguously. A caller skimming the docs might assume calling `ghRawTransport()` triggers a network fetch. The added `- Note:` makes the intent unambiguous. `LogFetcher` holding the closure for multiple calls is a deliberate design choice (avoids re-locking on every fetch) and the note surfaces that without requiring callers to read the implementation.

**Minor wording observation:** The three new lock descriptions are slightly inconsistent in what they describe:
- `transportLock`: *"Serialises all reads and writes to the active transport closure."*
- `rawTransportLock`: *"Serialises all reads and writes to the active raw transport closure."*
- `rateLimitLock`: *"Serialises all reads and writes to the rate-limit closure."*

The first two say "active X transport closure", the third drops "active". Not worth a separate PR — just flag it if you're in the file again.

---

### 4. `ScopePreferencesStore.swift` — confirmed clean ✅

No diff, correctly excluded. The issue's analysis of `failureHookEnabled` returning `Bool` vs `Bool?` inconsistency was noted and tracked in #1068 for future consideration — good hygiene.

---

### 5. CI status

| Check | Result |
|---|---|
| Swift Test | ✅ success |
| SwiftLint | ✅ success |
| Periphery | ✅ success |
| SonarCloud | ✅ success |
| review | ✅ success |
| ui-test | ⏭ skipped (expected) |

`mergeable_state: clean` — no conflicts with `main`.

---

### One thing to do before batch 7

The `ScopeStore.swift` `swiftlint:disable orphaned_doc_comment` suppression was explicitly deferred to the RunnerBar UI layer batch. Worth logging that as an open item in #1038 (or a dedicated issue) so it doesn't get lost in the transition. It's the only known remaining suppression in `RunnerBarCore` and closing it cleanly before batch 7 gets big would keep the suppression count meaningful.

---

**Summary:** all three changes are correct, well-justified, and match the issue analysis exactly. The `var → let` on `ScopeEntry.scope` is the only real code change and it's the right one. Ship it.

eonist:
## Deep Review — Batch 6: ScopeEntry, GitHubConstants, GitHubTransportShim

> ⚠️ **CI is currently `unstable`** — `mergeable_state` is not `clean`. Don't merge until checks pass.

Overall: **correct and well-scoped**. Every change is justified, nothing is missing, and the issue analysis in #1068 was thorough enough that the PR implementation follows it exactly. A few things worth calling out.

---

### ✅ What's good

**`ScopeEntry.scope` `var → let` is the right call**

The issue laid out the decision clearly — no call site mutates `scope` directly, and `ScopeStore`'s contract is remove+add. Making it `let` eliminates an entire class of subtle bug: SwiftUI bindings or async code inadvertently reassigning `scope` on a live entry, corrupting the identity model. The added doc comment is exactly right:

```swift
/// `let`: the scope string is immutable after construction. To change a scope,
/// remove the existing entry and add a new one via `ScopeStore`.
```

This is the pattern and the contract in one line. Good.

**Circular lock comment fix in `GitHubTransportShim.swift`**

The old comments were genuinely misleading:
```swift
/// Safety: protected by transportLock.   // ← on transportLock itself
```
The replacement (`Serialises all reads and writes to the active transport closure.`) is accurate and says something useful — it describes *what* the lock does, not just *that* it exists.

**`ghRawTransport()` closure-return note**

This is the kind of non-obvious API that bites callers once and then requires a grep+blame to understand. The `- Note:` addition makes the intent unambiguous without changing anything. Good defensive documentation.

---

### ⚠️ Issues worth flagging

**1. `ScopeEntry` is `Codable` — `var → let` has a JSON-round-trip implication worth confirming**

Swift's synthesised `Decodable` works fine with `let` properties — they're set during `init(from:)`. But if any call site does something like:

```swift
var entry = try JSONDecoder().decode(ScopeEntry.self, from: data)
entry.scope = "new/value"   // now a compile error
```

…that'll break at compile time, which is exactly the right outcome. The PR description says this was verified via grep — good. Just worth noting that the `Codable` conformance itself is safe; the mutation guard is the point.

**2. `isEnabled` remains `var` — the asymmetry is intentional but undocumented**

`scope` is now `let` with a doc explaining why. `isEnabled` stays `var` and its doc only says:
```swift
/// When `false`, `RunnerStore` skips this scope during polling.
```
That explains the semantics but not the mutability choice. A one-liner like `/// Mutable — toggled directly by \`ScopeStore.setEnabled(_:_:)\`.` would make the asymmetry deliberate-looking rather than accidental-looking. Minor, but this struct now has a clear immutability story that's half-told.

**3. `ScopePreferencesStore.swift` `failureHookEnabled` Bool vs `Bool?` inconsistency — still unaddressed**

Issue #1068 explicitly flagged this:
> `failureHookEnabled(for:)` returns `Bool` (not `Bool?`) — inconsistent with `notifyOnSuccess`/`notifyOnFailure` which return `Bool?` for three-state nil/true/false.

The PR correctly marks `ScopePreferencesStore.swift` as no-changes-needed (it was already clean), but the inconsistency isn't tracked anywhere after this. The issue notes it as "worth a doc note" and says it's tracked in #1068 for future consideration — but once #1068 closes, that note is gone. This should be a standalone issue before #1068 closes.

**4. `GitHubConstants.swift` — header-only fix, nothing else to check**

`RunnerBar → RunnerBarCore` is correct, trivial, and complete. ✅

---

### Issue #1068 scope compliance

| Goal | Status |
|---|---|
| `ScopeEntry.scope` `var → let` + doc | ✅ Done |
| `ScopePreferencesStore.swift` no changes needed | ✅ Confirmed clean, correctly skipped |
| `GitHubConstants.swift` header fix | ✅ Done |
| `GitHubTransportShim.swift` circular lock comments | ✅ Fixed |
| `ghRawTransport()` closure-return note | ✅ Added |
| No logic changes outside `scope var→let` | ✅ Correct |
| `failureHookEnabled` Bool vs `Bool?` doc note | ⚠️ Not done — should be filed as a follow-up before #1068 closes |

---

### Nits

- The `scope` doc comment uses a backtick-code block inside `///` for the `let` keyword itself — slightly unconventional but not wrong. Reads fine.
- `ScopeEntry` now has a clean two-tier immutability model: `id` and `scope` are identity (`let`), `isEnabled` is state (`var`). That's a good mental model. Worth surfacing in the type-level doc if one ever gets added.

---

### Summary

All three substantive changes are correct. The one real action item is filing a follow-up issue for the `failureHookEnabled` Bool vs `Bool?` inconsistency before #1068 is closed — otherwise that observation disappears. Everything else is merge-ready once CI goes green.

eonist:
## Review update — new commit (Sendable + import os)

**CI on the new commit: all green.** ✅ Swift Test, SwiftLint, Periphery, SonarCloud, review — all pass. ui-test skipped as expected.

---

### `ScopeEntry: Sendable` — correct and complete

The conformance is sound. All three stored properties are unconditionally `Sendable`:
- `id: UUID` — `Sendable` ✅ (value type, no interior mutability)
- `scope: String` — `Sendable` ✅ (value type)
- `isEnabled: Bool` — `Sendable` ✅ (value type)

The `var` on `isEnabled` is not a problem here — `Sendable` on a struct means the *type* can be safely passed across concurrency domains, not that all properties are immutable. Mutation of a local copy is fine; what `Sendable` prevents is sharing a *reference* to mutable state. Since `ScopeEntry` is a value type (`struct`), each actor/task gets its own copy when passed across boundaries. The conformance is structural (the compiler would enforce it) but explicit is better for a `public` type — future `var` additions will be caught at the `Sendable` boundary immediately.

One thing worth being aware of: `isEnabled: var` on a `Sendable` struct means if a Swift 6 caller does:
```swift
var entry = someEntry  // copy
entry.isEnabled = true // mutation on copy — fine
```
that's always safe. But if someone later adds a `class`-typed or non-`Sendable` property, the compiler will flag the conformance immediately. So `Sendable` here is also serving as a forward correctness sentinel — good.

---

### `import os` in `ScopeEntry.swift` — not visible in diff

The diff doesn't show an `import os` addition to `ScopeEntry.swift`, and `ScopeEntry` itself doesn't use `OSAllocatedUnfairLock` or any `os` type — `Sendable` is a Swift stdlib protocol, no import needed. If `import os` was confirmed correct for a *different* file in this commit (e.g. `GitHubTransportShim.swift` already had it), that's fine — it was already there before this PR. Nothing to action here, just noting the diff doesn't show an `os` import change in `ScopeEntry.swift`, which is correct.

---

### Final state of the diff

| File | Changes |
|---|---|
| `ScopeEntry.swift` | `Sendable` added to conformance list; `scope: var → let` + doc |
| `GitHubConstants.swift` | Header `RunnerBar → RunnerBarCore` |
| `GitHubTransportShim.swift` | Circular lock comments fixed; `ghRawTransport` closure-return note |
| `ScopePreferencesStore.swift` | No changes (confirmed clean) |

**No issues. Ready to merge.** 🚀

eonist:
## Updated Review — post-Sendable commit

**CI is green ✅** — both checks pass (`qlty check`: no blocking issues, `CodeRabbit`: skipped).

### `Sendable` conformance — correct and complete

All three stored properties satisfy `Sendable` structurally:
- `id: UUID` — `Sendable` ✅
- `scope: String` — `Sendable` ✅
- `isEnabled: Bool` — `Sendable` ✅

The compiler will synthesise `Sendable` implicitly for a struct where all stored properties are `Sendable`, but declaring it explicitly in the conformance list is the right call — it makes the intent auditable, shows up in generated docs, and will produce a compile error if a future property addition breaks the contract rather than silently losing thread-safety. No `@unchecked` needed, no `nonisolated(unsafe)` needed. Clean.

The `var → let` on `scope` is what makes this fully watertight — a `var String` would still satisfy `Sendable` since `String` is `Sendable`, but the immutability of `scope` reinforces the semantic: this struct is a value snapshot, not a mutable model object.

### One residual nit (non-blocking)

As noted in the previous review, `isEnabled` doc still only describes semantics, not mutability:
```swift
/// When `false`, `RunnerStore` skips this scope during polling.
public var isEnabled: Bool
```
Given that `scope` now has an explicit mutability doc, the asymmetry is more visible. A trailing `/// Mutable — toggled by \`ScopeStore.setEnabled(_:_:)\`.` would complete the picture. Strictly optional before merge.

### Verdict

**✅ Ready to merge.** CI green, `Sendable` correct, all batch 6 goals met. The two deferred items (`failureHookEnabled` Bool vs `Bool?`, `isEnabled` doc note) are scoped out cleanly in the PR and issue body.

PR #1071, opened 2026-06-01
Batch 7: docs + cleanup — RunnerEditCommit, RunnerEditDraft (#1070)

### LINKED ISSUE #1070: Batch 7: docs + cleanup — main, Exports, RunnerEditCommit, RunnerEditDraft, KeyablePanel, PanelChrome
Part of #1038 — first batch of the `RunnerBar` UI layer.

## Files in scope

- `Sources/RunnerBar/main.swift`
- `Sources/RunnerBar/Exports.swift`
- `Sources/RunnerBar/Models/RunnerEditCommit.swift`
- `Sources/RunnerBar/Models/RunnerEditDraft.swift`
- `Sources/RunnerBar/Panel/KeyablePanel.swift`
- `Sources/RunnerBar/Panel/PanelChrome.swift`

---

## Tasks

1. Remove `swiftlint:disable` closures — fix the underlying issue instead of suppressing
2. Improve/repair drifted comments
3. Improve code

---

## `main.swift` — ✅ no changes needed
Already fully hardened. `MainActor.assumeIsolated` wrapper has `❌ NEVER remove` guard.

## `Exports.swift` — ✅ no changes needed
`@_exported import RunnerBarCore` with clear ⚠️ design-decision comment.

## `KeyablePanel.swift` — ✅ no changes needed
Gold standard agent-hardened comment. Nothing to improve.

## `PanelChrome.swift` — ✅ no changes needed
Intentionally empty tombstone with `❌ NEVER restore` guard and issue link.

### 1. Remove `swiftlint:disable missing_docs`
- `CommitResult` and `commitRunnerEdit` already documented; private helpers exempt from rule

### 2. Comment repairs
- Step 1 error-asymmetry: API failure = early return; missing agentId/gitHubUrl = non-fatal, continues to local writes — comment was too weak, repaired
- `DispatchQueue.main.async` + `@MainActor`: add note that GCD is deliberate — do not replace with `await MainActor.run` without migrating the whole function to async/await
- `patchRunnerJSONMulti` `[String: Any]`: add comment that `Any` is intentional for mixed `String` + `Bool` single-pass patch
- `writeProxyFiles`: document two-line `.proxycredentials` format (line 1 = username, line 2 = password)

### 3. Code improvements
- `CommitResult.isSuccess`: simplify `if case` pattern to expression body
- `CommitResult.errors`: same
- `patchRunnerJSONMulti`: replace `installPath + "/.runner"` string concat with `URL(fileURLWithPath:).appendingPathComponent` throughout
- `writeProxyFiles`: same URL path improvement for `.proxy` and `.proxycredentials`
- `writeProxyFiles`: remove redundant `FileManager.fileExists` check before `removeItem` — `removeItem` handles missing files gracefully with `try?`

### 1. Remove `swiftlint:disable missing_docs`
- All non-private members already documented; private exempt

### 2. Comment repairs
- System-label filter: explain hardcoded list (`self-hosted`, `x64`, `arm64`, `linux`, `macos`, `windows`) are GitHub-managed labels injected at registration — never user-editable
- `loadProxy`: document `.proxy` (single-line URL) and `.proxycredentials` (two-line: user / password) formats
- `loadRunnerJSON`: clarify `@discardableResult` is intentional — preserved for future callers, current only caller discards it

### 3. Code improvements
- System-label filter: replace `!["self-hosted"].contains(label)` (allocates array for single-string check) with `label != "self-hosted"`
- `isDirty(comparedTo:)`: remove redundant `return` — body is already a single expression
- `loadProxy`: replace fragile `lines.first` / `lines.dropFirst().first` with explicit index-safe access `lines.indices.contains(1) ? lines[1] : ""`

---

## Implemented in PR #1071

eonist:
Part of #1038. Closes #1070.

## Tasks
1. Remove `swiftlint:disable` closures — fix underlying issues instead
2. Improve/repair drifted comments
3. Improve code

## Files confirmed clean — no changes
`main.swift`, `Exports.swift`, `KeyablePanel.swift`, `PanelChrome.swift`

### swiftlint
- Remove `swiftlint:disable missing_docs` / `enable`

### Comments
- Repair Step 1 asymmetry comment: API failure = early return; missing agentId/gitHubUrl = non-fatal, continues to local writes
- Add `DispatchQueue.main.async` note: deliberate GCD — do not replace with `await MainActor.run` without migrating the whole function
- Add `[String: Any]` explanation in `patchRunnerJSONMulti`: intentional for mixed `String` + `Bool` single-pass patch
- Document two-line `.proxycredentials` format in `writeProxyFiles`: line 1 = username, line 2 = password

### Code
- `CommitResult.isSuccess` / `.errors`: simplify `if case` pattern to expression body
- `patchRunnerJSONMulti` + `writeProxyFiles`: replace `installPath + "/..."` string concat with `URL(fileURLWithPath:).appendingPathComponent`
- `writeProxyFiles`: remove redundant `FileManager.fileExists` check before `removeItem`

### swiftlint
- Remove `swiftlint:disable missing_docs` / `enable`

### Comments
- Add system-label filter comment: explains `self-hosted`, `x64`, `arm64`, `linux`, `macos`, `windows` are GitHub-managed, never user-editable
- Add `loadProxy` doc: `.proxy` = single-line URL, `.proxycredentials` = two-line user/password
- Clarify `loadRunnerJSON` `@discardableResult`: intentional, preserved for future callers

### Code
- System-label filter: replace `!["self-hosted"].contains(label)` with `label != "self-hosted"`
- `isDirty(comparedTo:)`: remove redundant `return`
- `loadProxy`: replace fragile `lines.dropFirst().first` with explicit index-safe access

## Release Notes

* **Refactor**
  * Improved internal code structure and efficiency in runner editing workflows.
  * Enhanced file handling and error management in data processing operations.

* **Documentation**
  * Updated inline documentation for better clarity on runner edit behavior and completion handling.

### DIFF
--- a/Sources/RunnerBar/Models/RunnerEditCommit.swift
+++ b/Sources/RunnerBar/Models/RunnerEditCommit.swift
@@ -184,11 +184,11 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
                 ok = false
             }
         } else {
-            try url.write(to: proxyURL, atomically: true, encoding: .utf8) // write failure caught below
+            try url.write(to: proxyURL, atomically: true, encoding: .utf8)
             log("writeProxyFiles › wrote .proxy")
         }
     } catch {
-        // write failure — removeItem errors are handled by the inner catch above
+        // only reached on write failure (the remove path has its own inner do/catch)
         log("writeProxyFiles › .proxy error: \(error)")
         ok = false
     }
@@ -206,11 +206,11 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
                 ok = false
             }
         } else {
-            try "\(user)\n\(password)".write(to: credURL, atomically: true, encoding: .utf8) // write failure caught below
+            try "\(user)\n\(password)".write(to: credURL, atomically: true, encoding: .utf8)
             log("writeProxyFiles › wrote .proxycredentials")
         }
     } catch {
-        // write failure — removeItem errors are handled by the inner catch above
+        // only reached on write failure (the remove path has its own inner do/catch)
         log("writeProxyFiles › .proxycredentials error: \(error)")
         ok = false
     }
--- a/Sources/RunnerBar/Models/RunnerEditCommit.swift
+++ b/Sources/RunnerBar/Models/RunnerEditCommit.swift
@@ -14,14 +14,12 @@ enum CommitResult {
 
     /// `true` when the result has no errors.
     var isSuccess: Bool {
-        if case .success = self { return true }
-        return false
+        if case .success = self { true } else { false }
     }
 
     /// Convenience accessor for the error messages, empty on success.
     var errors: [String] {
-        if case .failure(let msgs) = self { return msgs }
-        return []
+        if case .failure(let msgs) = self { msgs } else { [] }
     }
 }
 
@@ -143,12 +141,11 @@ private func finalize(_ errors: [String], _ completion: @escaping @MainActor (Co
 /// `patches` accepts mixed `String` and `Bool` values via `Any` — this is intentional to allow
 /// updating both `workFolder` (String) and `disableUpdate` (Bool) in a single read-modify-write.
 private func patchRunnerJSONMulti(installPath: String, patches: [String: Any]) -> Bool {
-    let path = installPath + "/.runner"
-    let url = URL(fileURLWithPath: path)
+    let url = URL(fileURLWithPath: installPath).appendingPathComponent(".runner")
     guard let data = try? Data(contentsOf: url),
           var json = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
     else {
-        log("patchRunnerJSONMulti › failed to read \(path)")
+        log("patchRunnerJSONMulti › failed to read \(url.path)")
         return false
     }
     for (key, value) in patches { json[key] = value }
@@ -158,7 +155,7 @@ private func patchRunnerJSONMulti(installPath: String, patches: [String: Any]) -
     }
     do {
         try newData.write(to: url, options: .atomic)
-        log("patchRunnerJSONMulti › wrote keys=\(patches.keys.sorted()) to \(path)")
+        log("patchRunnerJSONMulti › wrote keys=\(patches.keys.sorted()) to \(url.path)")
         return true
     } catch {
         log("patchRunnerJSONMulti › write error: \(error)")
@@ -170,18 +167,17 @@ private func patchRunnerJSONMulti(installPath: String, patches: [String: Any]) -
 /// Removes the file when the relevant field is empty; writes otherwise.
 private func writeProxyFiles(installPath: String, url: String, user: String, password: String) -> Bool {
     var ok = true
-    let proxyFilePath = installPath + "/.proxy"
-    let credPath = installPath + "/.proxycredentials"
+    let base = URL(fileURLWithPath: installPath)
+    let proxyURL = base.appendingPathComponent(".proxy")
+    let credURL = base.appendingPathComponent(".proxycredentials")
 
     // .proxy — contains the raw proxy URL on a single line
     do {
         if url.isEmpty {
-            if FileManager.default.fileExists(atPath: proxyFilePath) {
-                try FileManager.default.removeItem(atPath: proxyFilePath)
-                log("writeProxyFiles › removed .proxy")
-            }
+            try? FileManager.default.removeItem(at: proxyURL)
+            log("writeProxyFiles › removed .proxy")
         } else {
-            try url.write(toFile: proxyFilePath, atomically: true, encoding: .utf8)
+            try url.write(to: proxyURL, atomically: true, encoding: .utf8)
             log("writeProxyFiles › wrote .proxy")
         }
     } catch {
@@ -192,12 +188,10 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
     // .proxycredentials — two-line format: line 1 = username, line 2 = password
     do {
         if user.isEmpty && password.isEmpty {
-            if FileManager.default.fileExists(atPath: credPath) {
-                try FileManager.default.removeItem(atPath: credPath)
-                log("writeProxyFiles › removed .proxycredentials")
-            }
+            try? FileManager.default.removeItem(at: credURL)
+            log("writeProxyFiles › removed .proxycredentials")
         } else {
-            try "\(user)\n\(password)".write(toFile: credPath, atomically: true, encoding: .utf8)
+            try "\(user)\n\(password)".write(to: credURL, atomically: true, encoding: .utf8)
             log("writeProxyFiles › wrote .proxycredentials")
         }
     } catch {
--- a/Sources/RunnerBar/Models/RunnerEditCommit.swift
+++ b/Sources/RunnerBar/Models/RunnerEditCommit.swift
@@ -184,10 +184,11 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
                 ok = false
             }
         } else {
-            try url.write(to: proxyURL, atomically: true, encoding: .utf8)
+            try url.write(to: proxyURL, atomically: true, encoding: .utf8) // write failure caught below
             log("writeProxyFiles › wrote .proxy")
         }
     } catch {
+        // write failure — removeItem errors are handled by the inner catch above
         log("writeProxyFiles › .proxy error: \(error)")
         ok = false
     }
@@ -205,10 +206,11 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
                 ok = false
             }
         } else {
-            try "\(user)\n\(password)".write(to: credURL, atomically: true, encoding: .utf8)
+            try "\(user)\n\(password)".write(to: credURL, atomically: true, encoding: .utf8) // write failure caught below
             log("writeProxyFiles › wrote .proxycredentials")
         }
     } catch {
+        // write failure — removeItem errors are handled by the inner catch above
         log("writeProxyFiles › .proxycredentials error: \(error)")
         ok = false
     }
--- a/Sources/RunnerBar/Models/RunnerEditCommit.swift
+++ b/Sources/RunnerBar/Models/RunnerEditCommit.swift
@@ -1,6 +1,5 @@
 // RunnerEditCommit.swift
 // RunnerBar
-// swiftlint:disable missing_docs
 import Foundation
 import RunnerBarCore
 
@@ -32,11 +31,15 @@ enum CommitResult {
 ///
 /// Commit order:
 /// 1. Labels (GitHub API) — if changed and agentId + scope are available.
-///    Aborts before touching disk if this step fails.
+///    Aborts the entire commit (returns early) if the API call fails.
+///    If agentId/scope are unavailable, appends an error and continues to local writes.
 /// 2. Runner JSON — workFolder + disableUpdate in one read-modify-write.
 /// 3. Proxy files — `.proxy` and `.proxycredentials` only when changed.
 ///
-/// Always dispatches `completion` on the main queue.
+/// Runs on a background `userInitiated` queue. Always dispatches `completion`
+/// on the main queue via `DispatchQueue.main.async` (compatible with the surrounding
+/// GCD context — do not replace with `await MainActor.run` without migrating the
+/// entire function to async/await first).
 func commitRunnerEdit(
     runner: RunnerModel,
     draft: RunnerEditDraft,
@@ -62,10 +65,12 @@ func commitRunnerEdit(
                 }
                 log("commitRunnerEdit › labels patched ok")
             } else {
+                // agentId or gitHubUrl unavailable — cannot call the API.
+                // Non-fatal: append an error and continue with local file writes
+                // so workFolder/proxy changes are not silently discarded.
                 let msg = "Cannot save labels: missing agent ID or GitHub URL"
                 log("commitRunnerEdit › \(msg)")
                 errors.append(msg)
-                // Non-fatal for local writes; continue
             }
         }
 
@@ -79,6 +84,8 @@ func commitRunnerEdit(
                 return
             }
             log("commitRunnerEdit › patching .runner JSON installPath=\(installPath)")
+            // patches uses [String: Any] to support mixed String + Bool values
+            // in a single JSON read-modify-write pass.
             let jsonOk = patchRunnerJSONMulti(
                 installPath: installPath,
                 patches: [
@@ -133,7 +140,8 @@ private func finalize(_ errors: [String], _ completion: @escaping @MainActor (Co
 }
 
 /// Reads the `.runner` JSON at `installPath`, merges all `patches` in one pass, and writes back.
-/// Accepts mixed `String` and `Bool` values via `Any`.
+/// `patches` accepts mixed `String` and `Bool` values via `Any` — this is intentional to allow
+/// updating both `workFolder` (String) and `disableUpdate` (Bool) in a single read-modify-write.
 private func patchRunnerJSONMulti(installPath: String, patches: [String: Any]) -> Bool {
     let path = installPath + "/.runner"
     let url = URL(fileURLWithPath: path)
@@ -159,12 +167,13 @@ private func patchRunnerJSONMulti(installPath: String, patches: [String: Any]) -
 }
 
 /// Writes (or removes) `.proxy` and `.proxycredentials` files at `installPath`.
+/// Removes the file when the relevant field is empty; writes otherwise.
 private func writeProxyFiles(installPath: String, url: String, user: String, password: String) -> Bool {
     var ok = true
     let proxyFilePath = installPath + "/.proxy"
     let credPath = installPath + "/.proxycredentials"
 
-    // .proxy
+    // .proxy — contains the raw proxy URL on a single line
     do {
         if url.isEmpty {
             if FileManager.default.fileExists(atPath: proxyFilePath) {
@@ -180,7 +189,7 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
         ok = false
     }
 
-    // .proxycredentials
+    // .proxycredentials — two-line format: line 1 = username, line 2 = password
     do {
         if user.isEmpty && password.isEmpty {
             if FileManager.default.fileExists(atPath: credPath) {
@@ -198,4 +207,3 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
 
     return ok
 }
-// swiftlint:enable missing_docs
--- a/Sources/RunnerBar/Models/RunnerEditCommit.swift
+++ b/Sources/RunnerBar/Models/RunnerEditCommit.swift
@@ -174,8 +174,15 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
     // .proxy — contains the raw proxy URL on a single line
     do {
         if url.isEmpty {
-            try? FileManager.default.removeItem(at: proxyURL)
-            log("writeProxyFiles › removed .proxy")
+            do {
+                try FileManager.default.removeItem(at: proxyURL)
+                log("writeProxyFiles › removed .proxy")
+            } catch let err as NSError where err.code == NSFileNoSuchFileError {
+                // file already absent — no-op
+            } catch {
+                log("writeProxyFiles › failed to remove .proxy: \(error)")
+                ok = false
+            }
         } else {
             try url.write(to: proxyURL, atomically: true, encoding: .utf8)
             log("writeProxyFiles › wrote .proxy")
@@ -188,8 +195,15 @@ private func writeProxyFiles(installPath: String, url: String, user: String, pas
     // .proxycredentials — two-line format: line 1 = username, line 2 = password
     do {
         if user.isEmpty && password.isEmpty {
-            try? FileManager.default.removeItem(at: credURL)
-            log("writeProxyFiles › removed .proxycredentials")
+            do {
+                try FileManager.default.removeItem(at: credURL)
+                log("writeProxyFiles › removed .proxycredentials")
+            } catch let err as NSError where err.code == NSFileNoSuchFileError {
+                // file already absent — no-op
+            } catch {
+                log("writeProxyFiles › failed to remove .proxycredentials: \(error)")
+                ok = false
+            }
         } else {
             try "\(user)\n\(password)".write(to: credURL, atomically: true, encoding: .utf8)
             log("writeProxyFiles › wrote .proxycredentials")
--- a/Sources/RunnerBar/Models/RunnerEditDraft.swift
+++ b/Sources/RunnerBar/Models/RunnerEditDraft.swift
@@ -98,8 +98,8 @@ struct RunnerEditDraft: Equatable {
     /// `autoUpdate` and `workFolder` to the draft.
     /// Returns the raw JSON dictionary so callers can extract additional fields
     /// (e.g. `platform`, `agentVersion`) without a second file read.
-    /// The return value may be discarded; it is exposed for callers that need
-    /// additional fields without a second file read.
+    /// The `@discardableResult` is intentional — most call sites only need the
+    /// side-effects; the dictionary is available for callers that need extra fields.
     @discardableResult
     mutating func loadRunnerJSON(installPath: String) -> [String: Any]? {
         let url = URL(fileURLWithPath: installPath).appendingPathComponent(".runner")
--- a/Sources/RunnerBar/Models/RunnerEditDraft.swift
+++ b/Sources/RunnerBar/Models/RunnerEditDraft.swift
@@ -94,12 +94,10 @@ struct RunnerEditDraft: Equatable {
 
     // MARK: - Private disk helpers
 
-    /// Reads and parses the `.runner` JSON at `installPath` and applies
-    /// `autoUpdate` and `workFolder` to the draft.
-    /// Returns the raw JSON dictionary so callers can extract additional fields
-    /// (e.g. `platform`, `agentVersion`) without a second file read.
-    /// The `@discardableResult` is intentional — most call sites only need the
-    /// side-effects; the dictionary is available for callers that need extra fields.
+    /// Reads and parses the `.runner` JSON at `installPath`, applies `autoUpdate` and
+    /// `workFolder` to the draft, and returns the raw dictionary for callers that need
+    /// additional fields (e.g. `platform`, `agentVersion`) without a second file read.
+    /// The return value may be discarded if only the draft side-effects are needed.
     @discardableResult
     mutating func loadRunnerJSON(installPath: String) -> [String: Any]? {
         let url = URL(fileURLWithPath: installPath).appendingPathComponent(".runner")
--- a/Sources/RunnerBar/Models/RunnerEditDraft.swift
+++ b/Sources/RunnerBar/Models/RunnerEditDraft.swift
@@ -41,7 +41,7 @@ struct RunnerEditDraft: Equatable {
         // linux, macos, windows. Only custom labels survive this filter.
         self.labelsText = runner.labels
             .filter { label in
-                !["self-hosted"].contains(label)
+                label != "self-hosted"
                     && !label.lowercased().contains("x64")
                     && !label.lowercased().contains("arm64")
                     && !label.lowercased().contains("linux")
@@ -102,8 +102,8 @@ struct RunnerEditDraft: Equatable {
     /// the return type is preserved for future callers.
     @discardableResult
     mutating func loadRunnerJSON(installPath: String) -> [String: Any]? {
-        let path = installPath + "/.runner"
-        guard let data = try? Data(contentsOf: URL(fileURLWithPath: path)),
+        let url = URL(fileURLWithPath: installPath).appendingPathComponent(".runner")
+        guard let data = try? Data(contentsOf: url),
               let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
         else { return nil }
         let disableUpdate = json["disableUpdate"] as? Bool ?? false
@@ -119,16 +119,19 @@ struct RunnerEditDraft: Equatable {
     /// `.proxycredentials` — two-line format: line 1 = username, line 2 = password.
     /// Missing files leave the corresponding draft fields as empty strings.
     private mutating func loadProxy(installPath: String) {
-        let proxyFilePath = installPath + "/.proxy"
-        proxyUrl = (try? String(contentsOfFile: proxyFilePath, encoding: .utf8))
+        let base = URL(fileURLWithPath: installPath)
+        let proxyURL = base.appendingPathComponent(".proxy")
+        let credURL = base.appendingPathComponent(".proxycredentials")
+
+        proxyUrl = (try? String(contentsOf: proxyURL, encoding: .utf8))
             .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) } ?? ""
 
-        let credPath = installPath + "/.proxycredentials"
-        if let credContent = try? String(contentsOfFile: credPath, encoding: .utf8) {
+        if let credContent = try? String(contentsOf: credURL, encoding: .utf8) {
             let lines = credContent.components(separatedBy: "\n")
             proxyUser = lines.first.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) } ?? ""
-            proxyPassword = lines.dropFirst().first
-                .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) } ?? ""
+            proxyPassword = lines.indices.contains(1)
+                ? lines[1].trimmingCharacters(in: .whitespacesAndNewlines)
+                : ""
         }
     }
 }
--- a/Sources/RunnerBar/Models/RunnerEditDraft.swift
+++ b/Sources/RunnerBar/Models/RunnerEditDraft.swift
@@ -37,17 +37,12 @@ struct RunnerEditDraft: Equatable {
     init(runner: RunnerModel) {
         // Filter out GitHub-managed system labels that are automatically assigned
         // by the runner registration process and should never be user-editable.
-        // These include the OS/arch labels GitHub injects: self-hosted, x64, arm64,
-        // linux, macos, windows. Only custom labels survive this filter.
+        // GitHub injects these as exact discrete tokens: self-hosted, x64, arm64,
+        // linux, macos, windows. Exact Set membership is used (not substring matching)
+        // so custom labels like "linux-ci" or "arm64-large" are preserved.
+        let systemLabels: Set<String> = ["self-hosted", "x64", "arm64", "linux", "macos", "windows"]
         self.labelsText = runner.labels
-            .filter { label in
-                label != "self-hosted"
-                    && !label.lowercased().contains("x64")
-                    && !label.lowercased().contains("arm64")
-                    && !label.lowercased().contains("linux")
-                    && !label.lowercased().contains("macos")
-                    && !label.lowercased().contains("windows")
-            }
+            .filter { !systemLabels.contains($0.lowercased()) }
             .joined(separator: ", ")
         self.workFolder = runner.workFolder ?? "_work"
         self.autoUpdate = true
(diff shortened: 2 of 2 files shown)


eonist:
## Deep Review — Batch 7: RunnerEditCommit + RunnerEditDraft (#1070)

**Verdict: ✅ Approve.** CI is fully green (Swift Test ✅, SwiftLint ✅, Periphery ✅, SonarCloud ✅, review ✅; ui-test skipped as expected). This is a well-scoped batch — real code improvements alongside the doc work, all justified. Notes below, roughly in order of importance.

---

### 1. `writeProxyFiles` — `try?` silently swallows removal errors ⚠️

The old code:
```swift
if FileManager.default.fileExists(atPath: proxyFilePath) {
    try FileManager.default.removeItem(atPath: proxyFilePath)
    // error propagates to outer do/catch → sets ok = false
}
```

The new code:
```swift
try? FileManager.default.removeItem(at: proxyURL)
log("writeProxyFiles › removed .proxy")
```

Removing the `fileExists` guard is correct — `removeItem` is idempotent when the file doesn't exist on macOS (it throws `NSFileNoSuchFileError` which you'd want to ignore anyway). But the switch to `try?` means a *genuine* removal failure — e.g. permissions error, file locked by another process — is now silently swallowed. The `ok = false` path in the `catch` block is no longer reachable for the remove case. The log line also always fires regardless of whether the file existed or the removal succeeded, which could mislead debugging.

**Suggested fix:**
```swift
do {
    try FileManager.default.removeItem(at: proxyURL)
    log("writeProxyFiles › removed .proxy")
} catch CocoaError.fileNoSuchFile {
    // already absent — no-op
} catch {
    log("writeProxyFiles › failed to remove .proxy: \(error)")
    ok = false
}
```
This keeps idempotency, preserves the `ok` sentinel, and logs genuine failures. Same pattern applies to `.proxycredentials`. This is the one concrete thing worth fixing before merge, or at minimum tracking as a follow-up issue.

---

### 2. `CommitResult` expression-body simplification ✅

```swift
// before
if case .success = self { return true }
return false

// after
if case .success = self { true } else { false }
```

Clean and correct. The implicit return form is idiomatic Swift and removes the early-return asymmetry. No concerns.

---

### 3. `patchRunnerJSONMulti` + `loadRunnerJSON`: string concat → `appendingPathComponent` ✅

```swift
// before
let path = installPath + "/.runner"
let url = URL(fileURLWithPath: path)

// after
let url = URL(fileURLWithPath: installPath).appendingPathComponent(".runner")
```

Correct. `appendingPathComponent` handles edge cases like a trailing slash on `installPath` and is more resistant to path separator bugs. Consistent across both `RunnerEditCommit` and `RunnerEditDraft`. The log messages updated to use `url.path` instead of the raw `path` string are also correct — same value, but derived from the canonical URL rather than a concatenated string.

---

### 4. System-label filter: `!["self-hosted"].contains(label)` → `label != "self-hosted"` ✅

```swift
// before
!["self-hosted"].contains(label)

// after
label != "self-hosted"
```

Correct simplification. The array literal with a single element was allocating a `[String]` on every filter iteration just to do an equality check. The new form is O(1) and reads clearly.

The added comment explaining which labels are GitHub-managed is genuinely valuable — this is the kind of implicit knowledge that causes regressions when someone adds `"self-hosted"` back manually:
```swift
// These include the OS/arch labels GitHub injects: self-hosted, x64, arm64,
// linux, macos, windows. Only custom labels survive this filter.
```

One nit: `x64`/`arm64`/`linux`/`macos`/`windows` are filtered via `.lowercased().contains(...)` rather than exact equality — so a label like `"custom-arm64-runner"` would also be stripped. The comment says these are GitHub-injected labels, which implies exact names. Worth verifying that `.contains` is intentional here (i.e. GitHub can inject variants like `"ubuntu-latest"` that contain `"linux"`) and if so, adding a note to that effect. If it's not intentional, the filter should use `== "arm64"` etc. instead. Not blocking — just flag for consideration.

---

### 5. `loadProxy`: `dropFirst().first` → `indices.contains(1)` ✅

```swift
// before
proxyPassword = lines.dropFirst().first
    .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) } ?? ""

// after
proxyPassword = lines.indices.contains(1)
    ? lines[1].trimmingCharacters(in: .whitespacesAndNewlines)
    : ""
```

Both are safe — `dropFirst().first` is not unsafe, it just creates an intermediate slice. The `indices.contains(1)` form is more explicit about the index-safety intent and reads better alongside the two-line format doc. Correct.

---

### 6. `DispatchQueue.main.async` migration note ✅

```swift
/// Runs on a background `userInitiated` queue. Always dispatches `completion`
/// on the main queue via `DispatchQueue.main.async` (compatible with the surrounding
/// GCD context — do not replace with `await MainActor.run` without migrating the
/// entire function to async/await first).
```

This is exactly the right kind of comment for a mixed-concurrency codebase. It prevents a well-intentioned future refactor from breaking the call chain. No concerns.

---

### 7. `@discardableResult` on `loadRunnerJSON` — doc is now accurate ✅

```swift
/// Currently only called from `load(installPath:)` which discards the return value;
/// the return type is preserved for future callers.
```

The previous doc implied active external callers. The new wording is honest about the current state while preserving the design intent. Good.

---

### 8. `swiftlint:disable missing_docs` removal ✅

Both files had blanket suppressions removed. SwiftLint passes on the new commit, which means all previously-suppressed declarations now have `///` docs. This is the correct fix-at-source approach rather than perpetuating the suppression.

---

### Summary

| Item | Severity | Status |
|---|---|---|
| `try?` silently swallows proxy removal errors | ⚠️ Worth fixing | Fix or track issue |
| `.contains` vs `==` for arch/OS label filter | 🔵 Minor | Verify intent |
| All other changes | ✅ Correct | No action needed |

The `try?` issue is the only thing I'd push back on. Everything else is genuinely cleaner than before — the `appendingPathComponent` migration, the expression-body simplifications, the label filter, the `loadProxy` index-safety fix, and especially the GCD migration note are all solid improvements.

eonist:
## Deep Review — Batch 7: RunnerEditCommit + RunnerEditDraft

**CI: `mergeable_state` is `clean` ✅ — ready on that front.**

This is the most substantive batch so far. Unlike the earlier doc-only batches, this one contains real logic changes alongside the comment/lint work, and most of them are improvements. A few need calling out.

---

### ✅ What's genuinely good

**`try? FileManager.default.removeItem(at:)` — removing the `fileExists` guard**

The old pattern:
```swift
if FileManager.default.fileExists(atPath: proxyFilePath) {
    try FileManager.default.removeItem(atPath: proxyFilePath)
}
```
has a TOCTOU race: the file can disappear between `fileExists` and `removeItem`. The new pattern:
```swift
try? FileManager.default.removeItem(at: proxyURL)
```
is idiomatic — `removeItem` on a non-existent file throws `NSCocoaErrorDomain / NSFileNoSuchFileError`, and `try?` silently swallows that specific case. This is correct and simpler. The one nuance: `try?` also silently swallows *unexpected* errors (permissions, locked file, etc.). The surrounding `ok = false` path is only reached via the `catch` block, which `try?` bypasses. This means a permissions error on `.proxy` removal will log nothing and return `ok = true`. Worth a follow-up issue, or change `try?` to `try` and handle the `NSFileNoSuchFileError` case explicitly.

**`lines.dropFirst().first` → `lines.indices.contains(1) ? lines[1] : ""`**

The old code:
```swift
proxyPassword = lines.dropFirst().first
    .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) } ?? ""
```
`dropFirst().first` on a single-element array returns `nil`, which is fine — but it's semantically opaque and allocates a new subsequence. The explicit index check is clearer and avoids the allocation. Good change.

**`URL(fileURLWithPath:).appendingPathComponent` throughout**

Replacing `installPath + "/.runner"` string concat with `URL(fileURLWithPath:).appendingPathComponent(".runner")` is strictly better — it handles trailing-slash edge cases correctly and avoids the silent bug where `installPath` ends with `/` producing `//. runner`. Consistent across all three helpers now.

**`if case .success = self { true } else { false }`**

Valid Swift expression-body syntax. Minor style point: some teams prefer `switch self { case .success: true; case .failure: false }` for exhaustiveness — the compiler would catch a missing case if `CommitResult` ever grows a third arm. The `if case` version silently falls through to `false` for any new variant. Non-blocking but worth knowing.

**`!["self-hosted"].contains(label)` → `label != "self-hosted"`**

Correct micro-optimisation. The array literal version allocates on every call; the equality check does not. No semantic difference for a single-element set.

---

**1. `try?` on proxy file removal silently swallows real errors**

As noted above — `try?` removes the `catch` path entirely. If `removeItem` fails for a reason other than "file not found" (permissions, sandbox, file locked), `ok` stays `true` and there's no log entry. The original code had the same gap (it only checked existence, not error type), so this isn't a regression — but the refactor is a good opportunity to fix it properly:

```swift
do {
    try FileManager.default.removeItem(at: proxyURL)
    log("writeProxyFiles › removed .proxy")
} catch let err as NSError where err.code == NSFileNoSuchFileError {
    // file already absent — not an error
} catch {
    log("writeProxyFiles › failed to remove .proxy: \(error)")
    ok = false
}
```

Optional for this batch, but since the function already has a `catch` block structure, it's low-effort.

**2. System-label filter is missing `macos` and `windows`**

The PR description and the added comment both say:
> These include the OS/arch labels GitHub injects: self-hosted, x64, arm64, linux, macos, windows.

But the actual filter in the diff is:
```swift
label != "self-hosted"
    && !label.lowercased().contains("x64")
    && !label.lowercased().contains("arm64")
    && !label.lowercased().contains("linux")
    && !label.lowercased().contains("macos")
    && !label.lowercased().contains("windows")
```

Wait — looking at the diff more carefully, `macos` and `windows` predicates are present in the new code but were **also present in the old code** (the diff only shows the lines that changed around the restructure). So the filter itself hasn't changed — only `self-hosted` moved from `!["self-hosted"].contains(label)` to `label != "self-hosted"`. The comment is therefore accurate. ✅ No issue here — just flagging the close read.

**3. `loadRunnerJSON` `@discardableResult` doc clarification introduces a minor inconsistency**

New doc:
```
/// Currently only called from `load(installPath:)` which discards the return value;
/// the return type is preserved for future callers.
```

This is accurate *today*, but "currently" is a time-bomb word in a doc comment — it will be wrong the moment a second caller is added, and no one will remember to update it. Suggest dropping "Currently" and rewording to: `/// The return value may be discarded; it is exposed for callers that need additional fields without a second file read.` That's perpetually accurate regardless of caller count.

**4. `isDirty(comparedTo:)` redundant `return` removal — not visible in the diff**

PR description mentions removing a redundant `return` in `isDirty(comparedTo:)` but this doesn't appear in the diff. Either it was already removed in a prior commit, the diff is truncated, or the description is ahead of the implementation. Worth a quick check.

---

### Issue #1070 scope compliance

| Goal | Status |
|---|---|
| Remove `swiftlint:disable missing_docs` | ✅ Both files cleaned |
| Repair Step 1 asymmetry comment | ✅ API failure vs missing agentId path now clearly differentiated |
| Add GCD/MainActor migration note | ✅ Clear and correct |
| `[String: Any]` explanation in `patchRunnerJSONMulti` | ✅ Good |
| Document `.proxycredentials` two-line format | ✅ In both write and load sites |
| `CommitResult` expression body simplification | ✅ |
| URL path construction (`appendingPathComponent`) | ✅ All three helpers consistent |
| Remove redundant `fileExists` check | ✅ (with caveat above on `try?`) |
| System-label filter comment | ✅ |
| `loadProxy` doc | ✅ |
| `@discardableResult` doc clarification | ⚠️ "Currently" is a time-bomb |
| `loadProxy` index-safe password access | ✅ |
| `isDirty` redundant `return` removal | ❓ Not visible in diff |

---

### Summary

Two real action items: the `try?` silent-error gap on proxy file removal (low effort to fix properly), and the `"Currently"` time-bomb in the `@discardableResult` doc. Everything else is correct. Merge-ready with those caveats noted.

eonist:
## Deep Review — PR #1071

Two files, well-scoped. Changes span docs, comment repair, and real code improvements. Reviewing each change on its merits.

---

### `RunnerEditCommit.swift`

**`CommitResult.isSuccess` / `.errors` — expression body simplification ✅**

```swift
// before
if case .success = self { return true }
return false

// after
if case .success = self { true } else { false }
```

Correct and idiomatic Swift. No behaviour change.

**Step 1 asymmetry comment — ✅ genuine repair**

The old comment implied both API failure and missing agentId/gitHubUrl caused an early abort. The new version correctly distinguishes:
- API call fails → early return (whole commit aborted)
- agentId/gitHubUrl missing → non-fatal, appends error, continues to local writes

This matches the actual `if/else` control flow in the function. Good.

**`DispatchQueue.main.async` note — ✅ valuable**

The migration warning (`do not replace with await MainActor.run without migrating the entire function`) is exactly the kind of thing that prevents a future contributor from making a half-migration that compiles but breaks the GCD call contract. Keep it.

**`patchRunnerJSONMulti` — `URL.appendingPathComponent` ✅**

```swift
// before
let path = installPath + "/.runner"
let url = URL(fileURLWithPath: path)

// after
let url = URL(fileURLWithPath: installPath).appendingPathComponent(".runner")
```

Correct improvement. `appendingPathComponent` handles path separators correctly on all platforms and avoids the hardcoded `"/"` concat. Log message updated to use `url.path` — consistent.

**`writeProxyFiles` — `removeItem` without `fileExists` check — ✅ correct**

```swift
// before
if FileManager.default.fileExists(atPath: proxyFilePath) {
    try FileManager.default.removeItem(atPath: proxyFilePath)
}

// after
try? FileManager.default.removeItem(at: proxyURL)
```

The `fileExists` check was a TOCTOU race: the file could be removed between the check and the `removeItem` call. Dropping it and using `try?` is the correct pattern — `removeItem` on a non-existent file throws `NSFileNoSuchFileError`, which `try?` silently discards. However, note that **the `log("writeProxyFiles › removed .proxy")` now fires even when the file didn’t exist** (there was nothing to remove). That’s a minor log accuracy issue — the log previously only fired when a file was actually removed. Not a functional bug but worth knowing.

---

### `RunnerEditDraft.swift`

**System-label filter comment — ✅ correct and useful**

The comment now names all six GitHub-managed labels (`self-hosted`, `x64`, `arm64`, `linux`, `macos`, `windows`) and explains why they’re filtered. This is important context that was previously invisible.

**`!["self-hosted"].contains(label)` → `label != "self-hosted"` — ✅**

Eliminating the single-element array allocation is a good micro-improvement. Semantically identical.

**`loadProxy` — `lines.dropFirst().first` → index-safe access — ✅ real improvement**

```swift
// before
proxyPassword = lines.dropFirst().first
    .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) } ?? ""

// after
proxyPassword = lines.indices.contains(1)
    ? lines[1].trimmingCharacters(in: .whitespacesAndNewlines)
    : ""
```

`dropFirst().first` is not wrong but it allocates an intermediate `ArraySlice`. The index-safe version is clearer about its intent ("is there a second line?") and avoids the allocation. Both are correct; the new version is marginally better.

**`loadRunnerJSON` `@discardableResult` clarification — ✅**

The added note ("currently only called from `load(installPath:)` which discards the return value; preserved for future callers") turns a potentially surprising `@discardableResult` into a documented decision. Good.

**`URL.appendingPathComponent` in `loadProxy` — ✅**

Same reasoning as `patchRunnerJSONMulti` above. Correct.

---

### One thing to flag — `macos` and `windows` not in the label filter

The comment says GitHub-managed labels include `self-hosted`, `x64`, `arm64`, `linux`, `macos`, `windows` — but the filter code only covers `self-hosted`, `x64`, `arm64`, `linux`:

```swift
label != "self-hosted"
    && !label.lowercased().contains("x64")
    && !label.lowercased().contains("arm64")
    && !label.lowercased().contains("linux")
    // macos and windows are NOT filtered
```

If `macos` and `windows` are truly GitHub-managed system labels that should never be user-editable, the comment overclaims. Either add them to the filter or remove them from the comment. This is the only real discrepancy in the PR.

---

### Summary

| # | Severity | Finding |
|---|---|---|
| 1 | 🟡 Fix | Comment claims `macos`/`windows` are filtered but they aren’t — fix comment or fix filter |
| 2 | 🟢 Nit | `log("removed .proxy")` fires even when the file didn’t exist after dropping `fileExists` check |

Everything else is clean and correct. Item 1 is the only thing worth addressing before merge.

eonist:
## Review — approve with one nit

All changes verified against local checked-out files. Docs-and-cleanup scope is correct, no unintended logic changes.

### ✅ Confirmed accurate

**`CommitResult.isSuccess` / `.errors`** — expression-body simplification is clean and correct. Both compile without the explicit `return`.

**Label filter** — `!["self-hosted"].contains(label)` → `label != "self-hosted"` is semantically identical and faster (no array allocation). The new comment accurately lists all filtered system labels and explains why they're excluded.

**`loadProxy` index-safe access** — `lines.indices.contains(1) ? lines[1] : ""` is a correct and clearer replacement for `lines.dropFirst().first`. No behaviour change for well-formed files; safer on a single-line or empty file.

**`appendingPathComponent` migration** — both files consistently replaced string concatenation with `URL(fileURLWithPath:).appendingPathComponent`. Correct.

**`isDirty` `return` removal** — confirmed removed, now an expression body `self != original`.

**`@discardableResult` doc** — accurate: only called from `load(installPath:)` which discards the return; preserving the return type for future callers is the right call.

**GCD comment on `DispatchQueue.main.async`** — the warning against replacing with `await MainActor.run` without migrating the entire function is correct and valuable.

### 🟡 One nit — `try?` inside `do/catch` is dead error handling

In `writeProxyFiles`, both `removeItem` calls were changed to `try?`:

```swift
do {
    if url.isEmpty {
        try? FileManager.default.removeItem(at: proxyURL)  // ← try?
        log("writeProxyFiles › removed .proxy")
    } else {
        try url.write(to: proxyURL, atomically: true, encoding: .utf8)
        log("writeProxyFiles › wrote .proxy")
    }
} catch {
    log("writeProxyFiles › .proxy error: \(error)")
    ok = false
}
```

The `try?` silently swallows any `removeItem` error — but the surrounding `do/catch` block then sets `ok = false` on the `write` path. This creates an asymmetry: a failed remove is silently ignored and logged as success (`"removed .proxy"`), while a failed write correctly sets `ok = false`. The previous `fileExists` guard was redundant (agreed), but the correct replacement is `try` not `try?`, so a failed remove propagates to the `catch` and sets `ok = false` consistently:

```swift
try FileManager.default.removeItem(at: proxyURL)
```

(If the file doesn't exist, `removeItem` throws `NSCocoaErrorDomain / NSFileNoSuchFileError`, which is harmless to catch and log — or add a `fileExists` guard only there if you prefer silent no-op semantics explicitly.)

Same issue on the `.proxycredentials` block.

### 🟢 Non-blocking

- `trimmedWorkFolder` still uses explicit `return v.isEmpty ? "_work" : v` — minor inconsistency with the expression-body style applied to `isDirty` and `CommitResult`, but pre-existing and out of scope.
- The `[String: Any]` doc on `patchRunnerJSONMulti` is accurate and the inline comment adds good context.

eonist:
## Review update — third commit

**CI: all green.** Swift Test ✅, SwiftLint ✅, Periphery ✅, SonarCloud ✅, review ✅. Ready to merge.

---

### `writeProxyFiles` — removal error handling ✅ correct

The nested `do/catch` pattern is exactly right:

```swift
do {
    try FileManager.default.removeItem(at: proxyURL)
    log("writeProxyFiles › removed .proxy")
} catch let err as NSError where err.code == NSFileNoSuchFileError {
    // file already absent — no-op
} catch {
    log("writeProxyFiles › failed to remove .proxy: \(error)")
    ok = false
}
```

Three things verified:
1. `NSFileNoSuchFileError` is explicitly swallowed — idempotency preserved ✅
2. Any other error (permissions, locked file) propagates to `ok = false` and gets logged ✅
3. The success log only fires when `removeItem` actually succeeded ✅

One observation worth knowing about but not blocking: the outer `do/catch` block that wraps the entire `.proxy` section still has a catch at the bottom that also sets `ok = false`. Since the nested `do` handles all `removeItem` throws itself, the only way the outer catch fires now is if `url.write(to:...)` throws in the `else` branch. That's fine and correct — just worth noting the two-layer structure is intentional and sound.

---

### `loadRunnerJSON` doc — ✅ improved

```swift
/// The return value may be discarded; it is exposed for callers that need
/// additional fields without a second file read.
```

This is the right wording. "Currently only called from..." was a time-bomb — it would become stale the moment a second call site appeared. The new form is perpetually accurate regardless of how many callers exist. ✅

One minor nit: the doc now says "without a second file read" twice in two consecutive sentences (once in the paragraph above, once in this new sentence). Not worth another commit — just trim if you're ever in the file again.

---

### `trimmedWorkFolder` explicit `return` — correct call to leave it

Agreed. The multi-line `let` binding before the conditional means an expression body isn't syntactically valid there. The explicit `return` is not noise in that context. ✅

---

**All review feedback addressed. Ship it. 🚀**

eonist:
## Follow-up review — post-fix commit

All three items from the previous review verified against the diff.

### ✅ `writeProxyFiles` — error handling correct

Both `.proxy` and `.proxycredentials` removal now use the nested `do/catch` pattern:

```swift
do {
    try FileManager.default.removeItem(at: proxyURL)
    log("writeProxyFiles › removed .proxy")
} catch let err as NSError where err.code == NSFileNoSuchFileError {
    // file already absent — no-op
} catch {
    log("writeProxyFiles › failed to remove .proxy: \(error)")
    ok = false
}
```

This is exactly right. `NSFileNoSuchFileError` is explicitly swallowed as a no-op, genuine errors propagate to `ok = false` and get logged, and the success log line only fires on an actual removal. The nested `do` inside the outer `do/catch` is slightly unusual visually but structurally sound — the outer `catch` now only handles write errors, which is cleaner than before.

**One minor observation:** the outer `catch` block (for the `else` / write path) now has a wider blast radius conceptually, but in practice the nested inner `do` catches everything from the `removeItem` path before it can escape, so there's no leakage. The structure is correct.

### ✅ `loadRunnerJSON` doc — time-bomb removed

New wording: `/// The return value may be discarded; it is exposed for callers that need additional fields without a second file read.` — perpetually accurate. ✅

Minor nit: the doc now has a slight redundancy — the preceding sentence already says *"Returns the raw JSON dictionary so callers can extract additional fields without a second file read"* and the new sentence repeats the same idea. Could tighten to just the `@discardableResult` rationale: `/// The return value may be discarded if only the draft side-effects are needed.` But this is purely cosmetic, not a blocker.

### ✅ `trimmedWorkFolder` explicit `return` — correct call to leave it

Agreed. A multi-line `let` binding before a conditional means there’s no single expression to elide the `return` from. Forcing an expression body there would require restructuring the logic, which is out of scope and adds no value.

---

**✅ All review items resolved. Ready to merge.**

eonist:
## Review — Batch 7: docs + cleanup (RunnerEditCommit, RunnerEditDraft)

Overall this is a clean, well-scoped batch. The intent is clear and the changes are consistent. A few observations below.

---

### ✅ Strong wins

- **`URL.appendingPathComponent` everywhere** — the string-concatenation paths (`installPath + "/.runner"`) were a latent bug waiting to happen on paths with trailing slashes. Good catch.
- **`fileExists` removal before `removeItem`** — replacing the check-then-delete with a try/catch on `NSFileNoSuchFileError` is the correct TOCTOU-safe pattern. Well done.
- **`if case` expression bodies** — `isSuccess` / `errors` are now more idiomatic Swift 5.9 expression-body style. No issues.
- **`label != "self-hosted"`** — the original `!["self-hosted"].contains(label)` allocated an array on every filter pass. The replacement is both cleaner and faster.
- **`lines.indices.contains(1)` for proxy password** — safer than `dropFirst().first`, especially if the file has only one line.

---

### 🟡 Minor observations

**1. `loadRunnerJSON` doc comment duplication**

The new doc for `loadRunnerJSON` says the same thing twice:

```
/// Returns the raw JSON dictionary so callers can extract additional fields
/// (e.g. `platform`, `agentVersion`) without a second file read.
/// The return value may be discarded; it is exposed for callers that need
/// additional fields without a second file read.
```

The second sentence is redundant — consider dropping it or merging into one sentence.

**2. System-label filter comment mentions `windows` and `macos` but the filter doesn't cover them**

The new comment says:
> These include the OS/arch labels GitHub injects: self-hosted, x64, arm64, linux, macos, windows.

But the actual filter only checks `x64`, `arm64`, and `linux` via `.lowercased().contains(...)` — `macos` and `windows` are not filtered. Either the filter should be extended, or the comment should be corrected to match what the code actually does. As-is this could mislead a future reader.

**3. `writeProxyFiles` outer catch still fires for the `url.isEmpty == false` branch**

The refactor introduced a nested `do/catch` for the remove path, but the outer `catch` at the bottom of each block still catches write errors from the `else` branch. This is correct, but worth a brief note in the code to make the two-level catch structure intentional/obvious — a future reader might try to simplify it and reintroduce the bug.

---

### 🔵 Nit

The inline `// patches uses [String: Any]...` comment added at the call site in `commitRunnerEdit` is now slightly redundant with the improved `patchRunnerJSONMulti` doc comment directly above the function. Fine to keep for call-site context, but could be trimmed.

---

**Verdict:** Approve with the label-filter/comment mismatch (point 2) worth fixing before merge — the rest are nits. Good batch.

eonist:
## Review: Batch 7 — RunnerEditCommit + RunnerEditDraft

Overall this is a clean, well-scoped batch. The changes do exactly what they say: remove SwiftLint suppressions, tighten comments, and apply a handful of focused code improvements. A few observations below.

---

### ✅ What's working well

- **URL path construction** — replacing `installPath + "/..."` string concat with `URL(fileURLWithPath:).appendingPathComponent(...)` throughout is the right call. Avoids double-slash edge cases and is more idiomatic Swift.
- **`fileExists` removal in `writeProxyFiles`** — the new try/catch pattern catching `NSFileNoSuchFileError` is strictly better: it avoids a TOCTOU race, reduces syscalls, and handles the "not found" case explicitly rather than silently.
- **`CommitResult` expression bodies** — `if case .success = self { true } else { false }` is a nice simplification. Reads well.
- **`label != "self-hosted"` refactor** — `![...].contains(label)` for a single element was unnecessary. Good cleanup.
- **`loadProxy` index-safe access** — `lines.indices.contains(1)` is clearer and safer than the chained `dropFirst().first` dance.

---

### 🔍 Minor observations

**`writeProxyFiles` — outer `catch` after inner try/catch**

The `do { if url.isEmpty { do { ... } catch ... catch { ok = false } } else { try url.write(...) } } catch { log(...); ok = false }` nesting is a bit layered. The outer `catch` now only fires for the `url.write(to:)` path, which is correct — but it's easy to misread. A small inline comment like `// write failure` on that outer catch would make the control flow intent obvious at a glance.

**`label != "self-hosted"` — filter breadth**

The comment added in `init(runner:)` says the filter targets `self-hosted, x64, arm64, linux, macos, windows` — but `macos` and `windows` aren't in the predicate (only `x64`, `arm64`, `linux` via `lowercased().contains(...)`). Either the comment is slightly ahead of the implementation, or those two were intentionally left out. Worth a quick check to confirm the intent matches.

**`loadRunnerJSON` doc comment**

The new doc repeats itself slightly: *"Returns the raw JSON dictionary so callers can extract additional fields (e.g. platform, agentVersion) without a second file read."* appears twice (once in the body, once in the `@discardableResult` rationale). Could be collapsed into one sentence.

**GCD note in `commitRunnerEdit` doc**

The added doc line — *"do not replace with `await MainActor.run` without migrating the entire function to async/await first"* — is genuinely useful for future maintainers. Good defensive commenting.

---

### Summary

Two files, 66 additions / 39 deletions, all purposeful. The `writeProxyFiles` refactor is the most impactful functional change and it's solid. The label filter comment vs. implementation mismatch (`macos`/`windows`) is the one thing worth double-checking before merge. Everything else is polish-level.

eonist:
## Code Review — Batch 7: docs + cleanup

Overall this is a clean, well-scoped batch. The changes are focused, each edit has a clear rationale, and the PR description accurately describes what was done. A few observations below.

---

### ✅ Strengths

- **`URL`-based path construction** — replacing string concatenation (`installPath + "/.runner"`) with `URL(fileURLWithPath:).appendingPathComponent(...)` throughout both files is the right call. It's safer, avoids separator bugs, and aligns with Apple's recommended APIs.
- **Redundant `fileExists` removal in `writeProxyFiles`** — the new try/catch pattern with `NSFileNoSuchFileError` is idiomatic and correctly handles the TOCTOU race that the old check had. Nice improvement.
- **`if case` expression bodies** — the simplification of `isSuccess` and `errors` to single-expression `if case` bodies is readable and correct.
- **System-label filter** — replacing `!["self-hosted"].contains(label)` with `label != "self-hosted"` is a good micro-optimisation and more readable. The added comment explaining the full set of GitHub-managed labels is valuable context for future maintainers.
- **`loadProxy` index-safe access** — swapping `lines.dropFirst().first` for `lines.indices.contains(1) ? lines[1] : ""` is more explicit and avoids the subtle issue of `dropFirst` returning an empty slice on single-line content.

---

### 🔍 Minor observations

1. **`writeProxyFiles` outer `catch`** — After the inner nested `do/catch` was introduced for the remove path, the outer `catch` at the bottom of each block now only fires for the `write` branch. This is correct, but the structure is slightly asymmetric and could trip up a future reader. A brief comment like `// only reached on write failure` would make intent clear.

2. **`loadRunnerJSON` doc comment duplication** — The updated doc says "without a second file read" twice (once in the body, once in the `@discardableResult` sentence). Worth trimming one instance.

3. **`if case` expression syntax** — `if case .success = self { true } else { false }` works in Swift 5.9+ but lacks explicit `return` which some linters flag. Since `missing_docs` suppressions are being removed, it's worth confirming SwiftLint is happy with this form (no `implicit_return` rule active for this module).

4. **`label != "self-hosted"` filter** — The other labels (`x64`, `arm64`, `linux`, `macos`, `windows`) are still filtered with `.lowercased().contains(...)`. For consistency and clarity, consider whether an explicit `Set`-based filter would better document the full list in one place, especially since the comment now lists all six.

---

### 🟢 Verdict

**Approve.** Changes are correct, well-documented, and improve both safety and readability. The minor points above are non-blocking suggestions.

eonist:
## Review — Batch 7: RunnerEditCommit + RunnerEditDraft

CI is all green and the PR is well-scoped. Existing reviews have covered the major ground (URL path migration, `try?` proxy removal, expression-body simplifications). A few additive observations that haven't been fully resolved yet:

---

### 🔴 Label filter: comment vs. code mismatch

This was flagged in prior reviews but remains unresolved in the current diff. The added comment says:

```swift
// These include the OS/arch labels GitHub injects: self-hosted, x64, arm64,
// linux, macos, windows. Only custom labels survive this filter.
```

But the filter uses `.lowercased().contains(...)` substring matching, not exact equality. A label like `"custom-linux-runner"` or `"windows-prod"` would be silently stripped. If these are meant to be exact GitHub-injected tokens (which they are — GitHub injects `linux`, `macos`, `windows` as discrete labels), the filter should use exact matching:

```swift
let systemLabels: Set<String> = ["self-hosted", "x64", "arm64", "linux", "macos", "windows"]
self.labelsText = runner.labels
    .filter { !systemLabels.contains($0.lowercased()) }
    ...
```

This is the one concrete correctness issue in the PR — substring matching will silently drop legitimate custom labels containing these substrings on commit.

---

### 🟡 `writeProxyFiles` — success log fires unconditionally after `removeItem`

The current `do/catch` pattern after the fix is correct for error propagation, but the `log("writeProxyFiles › removed .proxy")` line fires immediately after `try FileManager.default.removeItem(at: proxyURL)` — which only runs when the file *exists* and removal succeeds. If the `NSFileNoSuchFileError` catch fires instead (file was already absent), no log fires at all. This is actually fine and more accurate than before — just worth confirming it's intentional since previous reviews noted the old `try?` version logged unconditionally.

---

### 🟢 Everything else: confirmed correct

- `URL.appendingPathComponent` migration throughout ✅
- Nested `do/catch` with `NSFileNoSuchFileError` for idempotent removal ✅  
- `CommitResult` expression bodies ✅
- `label != "self-hosted"` micro-optimisation ✅
- `lines.indices.contains(1)` for proxy password ✅
- GCD migration warning in `commitRunnerEdit` doc ✅
- `@discardableResult` doc (post-fix wording) ✅

**Verdict:** The label filter substring-vs-exact issue is the only thing worth addressing before merge. Everything else is solid.

eonist:
## Review — Batch 7: RunnerEditCommit + RunnerEditDraft

After reading all prior reviews: Qodo, Octopus, CodeAnt, SonarCloud, and the deep-review comments are thorough. I'll keep this additive and avoid rehashing what's already been well-covered.

---

### 🔴 One unresolved blocker: label filter substring vs. exact match

Every prior review flagged this and it remains open. The comment says:

```swift
// These include the OS/arch labels GitHub injects: self-hosted, x64, arm64,
// linux, macos, windows. Only custom labels survive this filter.
```

But the actual filter uses `.lowercased().contains(...)` — meaning a user's custom label like `"linux-ci"`, `"windows-prod"`, or `"arm64-large"` would be silently stripped on every commit. GitHub injects these as **exact discrete tokens** (`linux`, `macos`, `windows`, etc.), so the filter should match exactly:

```swift
let systemLabels: Set<String> = ["self-hosted", "x64", "arm64", "linux", "macos", "windows"]
self.labelsText = runner.labels
    .filter { !systemLabels.contains($0.lowercased()) }
    .joined(separator: "\n")
```

This is the one concrete correctness issue worth fixing before merge. The comment and the filter are in disagreement, and the current behavior can silently drop user-defined labels.

---

### ✅ Everything else is solid

The existing reviews have thoroughly signed off on:
- `URL.appendingPathComponent` migration ✅
- Nested `do/catch` with `NSFileNoSuchFileError` for idempotent removal ✅
- `CommitResult` expression bodies ✅
- `lines.indices.contains(1)` for proxy password ✅
- GCD/`MainActor` migration warning ✅
- `@discardableResult` doc (post-fix wording) ✅
- SonarCloud: 0 new issues ✅

**Verdict:** Fix the label filter (exact match via `Set` instead of substring `contains`) and this is good to merge.

eonist:
## Review — Batch 7: RunnerEditCommit + RunnerEditDraft

CI is green (Swift Test ✅, SwiftLint ✅, Periphery ✅, SonarCloud ✅). Reviewed the full diff and all prior comments to avoid duplication.

---

### ✅ Confirmed good

- **`URL.appendingPathComponent` migration** — correct throughout both files. Handles trailing-slash edge cases that the old string concat didn't.
- **`writeProxyFiles` nested `do/catch`** — the final form (explicit `NSFileNoSuchFileError` swallow, real errors propagate to `ok = false`) is the right pattern. The outer `catch` correctly now only covers write failures, as noted in the inline comments.
- **`CommitResult` expression bodies** — idiomatic Swift 5.9+, no behaviour change.
- **`label != "self-hosted"` simplification** — O(1), no array allocation, correct.
- **`loadProxy` index-safe password access** — `lines.indices.contains(1)` is clearer than `dropFirst().first` and makes the two-line format contract explicit.
- **GCD migration note on `commitRunnerEdit`** — valuable defensive comment, prevents a half-migration footgun.

---

### 🟡 One unresolved item — `.contains` vs exact equality for `x64`/`arm64`/`linux`/`macos`/`windows`

Multiple prior reviews flagged this and it hasn't been addressed. The current filter:

```swift
label != "self-hosted"
    && !label.lowercased().contains("x64")
    && !label.lowercased().contains("arm64")
    && !label.lowercased().contains("linux")
    && !label.lowercased().contains("macos")
    && !label.lowercased().contains("windows")
```

...will strip custom labels like `linux-ci`, `windows-prod`, `macos-arm64-runner`. The added comment says these are GitHub-managed labels that are *exact tokens* — which implies the filter should use equality, not substring matching. Qodo's inline thread flagged this too (currently open).

**Suggested fix** — use a `Set` for exact matching, consistent with the comment's intent:

```swift
private let systemLabels: Set<String> = ["self-hosted", "x64", "arm64", "linux", "macos", "windows"]

// in the filter:
!systemLabels.contains(label.lowercased())
```

This is the one concrete thing worth fixing before merge. Everything else is clean.

---

**Verdict: Approve pending the label filter fix.** 🚀

eonist:
Acknowledged — the label filter fix (commit 528c028, `systemLabels: Set<String>` with `!systemLabels.contains($0.lowercased())`) lands exactly what the review asked for. SonarCloud re-ran clean and Octopus confirmed the Set-based fix as a positive in its updated pass. Nothing left to act on — ready to merge. ✅

eonist:
Acknowledged — the label filter was already fixed in commit 528c028 (`systemLabels: Set<String>` + `!systemLabels.contains($0.lowercased())`). My review was based on an older diff. SonarCloud and Octopus have both re-run and confirmed green on the new commit. Nothing actionable — the PR is ready to merge. ✅

PR #1076, opened 2026-06-01
Batch 9 — Runner: docs + lint fix (RunnerLifecycleService, RunnerPollState)

### LINKED ISSUE #1074: Batch 9 — Runner: docs + lint fix (RunnerLifecycleService, RunnerPollState)
Part of the ongoing codebase review and documentation pass tracked in #1038.

## Files in scope

- `Sources/RunnerBar/Runner/RunnerLifecycleService.swift`
- `Sources/RunnerBar/Runner/RunnerPollState.swift`

### `RunnerLifecycleService.swift`
- **Remove `swiftlint:disable:next type_body_length`** — fix the underlying cause by extracting a private `logStep(_:_:)` helper to collapse the repetitive `log("RunnerLifecycle > ...")` call pattern across `start()`, `stop()`, and `remove()`, reducing body line count organically
- Replace all filler doc comments (`/// Performs the X operation.`, `/// The X case.`) with real descriptions
- `LifecycleResult` cases: document *when* each case is returned
- `struct` doc: replace filler with accurate description of purpose
- `shared` doc: update to established singleton wording
- `scopeFromGitHubUrl`: document that it derives `owner/repo` or `org` from a GitHub URL
- `remove()` returns `Bool` while `start()`/`stop()` return `LifecycleResult` — add a comment explaining the inconsistency

### `RunnerPollState.swift`
- Replace all filler `/// Performs the X operation.` doc comments with real descriptions
- `SendableFormatter` and `iso8601Lock` docs: move the *why* (expensive ICU allocation) from the `//` block comment into the `///` doc
- `extension RunnerStore` doc: replace filler with description of what the extension does (bridges `PollResultBuilder` for `RunnerStore.fetch()` call sites)
- Add explicit warning comment on the two `DispatchQueue.main.sync` calls inside `nonisolated` functions — deadlock risk if called from main thread

Closes part of #1038.

eonist:
## **User description**
Part of the ongoing codebase review tracked in #1038. Closes #1074.

## Files reviewed

| File | Changes |
|------|---------|
| `Runner/RunnerLifecycleService.swift` | SwiftLint fix, real docs throughout |
| `Runner/RunnerPollState.swift` | Real docs, deadlock warnings, SendableFormatter rationale |

### SwiftLint: `type_body_length` suppression removed ✅
Extracted a private `logStep(_:_:)` helper that centralises the `log("RunnerLifecycle > ...")` prefix. All logging calls in `start()`, `stop()`, and `remove()` now route through it. This collapses dozens of repetitive log lines and reduces the body line count organically — no suppression needed.

### Docs
- `LifecycleResult` enum: replaced `/// The X case.` fillers with descriptions of *when* each case is returned
- Struct doc: replaced `/// A value type representing...` filler with a real description of purpose
- `shared` doc: updated to established singleton wording
- `start()` / `stop()`: documented the two-step sequence and return value semantics
- `remove()`: documented the multi-step best-effort flow; added a `- Note:` explaining why it returns `Bool` instead of `LifecycleResult` (internal API fallback handles partial failures)
- `isCorruptInstall`: documented what strings trigger the corrupt detection
- `deleteLaunchAgentPlist`: documented the file naming pattern it matches
- `scopeFromGitHubUrl`: documented the derivation logic with examples
- `runScriptWithOutput`: documented the wrapper contract

### Docs
- `SendableFormatter`: moved the expensive-allocation rationale from the `//` block comment into the `///` doc; added note that the `@unchecked` annotation is safe due to lock serialisation
- `iso8601Lock`: added doc explaining it must always be accessed via `withLock`
- `extension RunnerStore`: replaced `/// Extension adding functionality to RunnerStore.` filler with a description of what the extension does
- `buildJobState` / `buildGroupState` / `backfillSteps` / `scopeFromActionGroup` / `enrichGroupJobs`: all replaced filler with real descriptions

### Deadlock warning added
Both `buildJobState` and `buildGroupState` call `DispatchQueue.main.sync` inside `nonisolated` closures to read `ScopeStore.shared.scopes`. Added `// ⚠️ main.sync` inline comments and a class-level doc warning — these are safe only when called from a background thread.

___

## **CodeAnt-AI Description**
**Clarify runner lifecycle behavior and cleanup steps**

### What Changed
- Replaced placeholder comments with plain-language explanations of how runner start, stop, and full removal work
- Added clearer guidance on when a runner install is treated as corrupt and when reinstall is needed
- Documented how scope names are derived, how polling data is merged, and why shared date formatting is handled safely
- Removed repeated logging text by centralizing lifecycle log messages

### Impact
`✅ Easier maintenance of runner lifecycle behavior`
`✅ Clearer reinstall guidance for corrupt runner installs`
`✅ Safer cleanup of runner files and launch agents`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

### DIFF
--- a/Sources/RunnerBar/Runner/RunnerLifecycleService.swift
+++ b/Sources/RunnerBar/Runner/RunnerLifecycleService.swift
@@ -5,193 +5,214 @@ import RunnerBarCore
 
 // MARK: - LifecycleResult
 
-/// Enumerates possible values for LifecycleResult.
+/// The result of a runner lifecycle operation (start or stop).
 enum LifecycleResult {
-    /// The `success` case.
+    /// The operation completed successfully.
     case success
-    /// The `corruptInstall` case.
+    /// The runner installation is corrupt (e.g. missing `svc.sh`, wrong working directory).
+    /// The caller should prompt the user to reinstall the runner.
     case corruptInstall
-    /// The `failed` case.
+    /// The operation failed with a human-readable reason string.
     case failed(String)
 }
 
 // MARK: - RunnerLifecycleService
 
-// swiftlint:disable:next type_body_length
-/// A value type representing RunnerLifecycleService.
+/// Manages the macOS launchctl service lifecycle for locally-installed GitHub Actions runner agents.
+///
+/// Each runner is installed as a launchd agent via the runner's own `svc.sh` script.
+/// This service drives the install → start, stop → uninstall, and full removal sequences,
+/// delegating process execution to `ProcessRunner` and token fetching to the GitHub API.
 struct RunnerLifecycleService {
-    /// The shared constant.
+    /// Shared singleton — use this instead of calling init directly.
     static let shared = RunnerLifecycleService()
     /// Private initialiser — use `shared`.
     private init() {}
 
     // MARK: - Start
 
-    /// Performs the start operation.
+    /// Installs and starts the launchd service for `runner` by running `svc.sh install` then `svc.sh start`.
+    ///
+    /// Returns `.corruptInstall` if either step detects a broken installation,
+    /// `.success` if the start step exits 0, or `.failed` with the first non-empty
+    /// output line otherwise.
     @discardableResult
     func start(runner: RunnerModel) -> LifecycleResult {
         let ip = runner.installPath ?? "nil"
         let gh = runner.gitHubUrl ?? "nil"
-        log("RunnerLifecycle > START called runner=\(runner.runnerName) installPath=\(ip) gitHubUrl=\(gh)")
+        logStep("START", "called runner=\(runner.runnerName) installPath=\(ip) gitHubUrl=\(gh)")
         guard let path = runner.installPath else {
-            log("RunnerLifecycle > START abort — no installPath for \(runner.runnerName)")
+            logStep("START", "abort — no installPath for \(runner.runnerName)")
             return .failed("Install path unknown")
         }
         let dir = URL(fileURLWithPath: path)
         let svcPath = dir.appendingPathComponent("svc.sh").path
-        let svcExists = FileManager.default.fileExists(atPath: svcPath)
-        let svcExec = FileManager.default.isExecutableFile(atPath: svcPath)
-        log("RunnerLifecycle > START svc.sh=\(svcPath) exists=\(svcExists) executable=\(svcExec)")
-        log("RunnerLifecycle > START step1: svc.sh install")
+        logStep("START", "svc.sh=\(svcPath) exists=\(FileManager.default.fileExists(atPath: svcPath)) executable=\(FileManager.default.isExecutableFile(atPath: svcPath))")
+
+        logStep("START", "step1: svc.sh install")
         let (installOk, installOutput) = runScriptWithOutput(
-            executableName: "svc.sh",
-            arguments: ["install"],
-            workingDirectory: dir,
-            timeout: 15,
-            logTag: "svc.sh install")
-        log("RunnerLifecycle > START step1 done: ok=\(installOk) output=\(installOutput.prefix(300))")
-        log("RunnerLifecycle > START step1 corruptCheck: isCorrupt=\(isCorruptInstall(output: installOutput))")
+            executableName: "svc.sh", arguments: ["install"],
+            workingDirectory: dir, timeout: 15, logTag: "svc.sh install")
+        logStep("START", "step1 done: ok=\(installOk) output=\(installOutput.prefix(300))")
         if isCorruptInstall(output: installOutput) {
-            log("RunnerLifecycle > START RETURNING .corruptInstall after install step for \(runner.runnerName)")
+            logStep("START", "RETURNING .corruptInstall after install step for \(runner.runnerName)")
             return .corruptInstall
         }
-        log("RunnerLifecycle > START step2: svc.sh start")
+
+        logStep("START", "step2: svc.sh start")
         let (startOk, startOutput) = runScriptWithOutput(
-            executableName: "svc.sh",
-            arguments: ["start"],
-            workingDirectory: dir,
-            timeout: 15,
-            logTag: "svc.sh start")
-        log("RunnerLifecycle > START step2 done: ok=\(startOk) output=\(startOutput.prefix(300))")
-        log("RunnerLifecycle > START step2 corruptCheck: isCorrupt=\(isCorruptInstall(output: startOutput))")
+            executableName: "svc.sh", arguments: ["start"],
+            workingDirectory: dir, timeout: 15, logTag: "svc.sh start")
+        logStep("START", "step2 done: ok=\(startOk) output=\(startOutput.prefix(300))")
         if isCorruptInstall(output: startOutput) {
-            log("RunnerLifecycle > START RETURNING .corruptInstall after start step for \(runner.runnerName)")
+            logStep("START", "RETURNING .corruptInstall after start step for \(runner.runnerName)")
             return .corruptInstall
         }
         if startOk {
-            log("RunnerLifecycle > START RETURNING .success for \(runner.runnerName)")
+            logStep("START", "RETURNING .success for \(runner.runnerName)")
             return .success
         }
         let msg = startOutput.components(separatedBy: "\n")
             .first(where: { !$0.trimmingCharacters(in: .whitespaces).isEmpty }) ?? "Failed to start"
-        log("RunnerLifecycle > START RETURNING .failed(\(msg)) for \(runner.runnerName)")
+        logStep("START", "RETURNING .failed(\(msg)) for \(runner.runnerName)")
         return .failed(msg)
     }
 
     // MARK: - Stop
 
-    /// Performs the stop operation.
+    /// Stops and uninstalls the launchd service for `runner` by running `svc.sh stop` then `svc.sh uninstall`.
+    ///
+    /// Returns `.corruptInstall` if the stop step detects a broken installation,
+    /// `.success` if the stop step exits 0, or `.failed` with the first non-empty
+    /// output line otherwise.
     @discardableResult
     func stop(runner: RunnerModel) -> LifecycleResult {
         let ip = runner.installPath ?? "nil"
-        log("RunnerLifecycle > STOP called runner=\(runner.runnerName) installPath=\(ip)")
+        logStep("STOP", "called runner=\(runner.runnerName) installPath=\(ip)")
         guard let path = runner.installPath else {
-            log("RunnerLifecycle > STOP abort — no installPath for \(runner.runnerName)")
+            logStep("STOP", "abort — no installPath for \(runner.runnerName)")
             return .failed("Install path unknown")
         }
         let dir = URL(fileURLWithPath: path)
-        log("RunnerLifecycle > STOP step1: svc.sh stop")
+
+        logStep("STOP", "step1: svc.sh stop")
         let (stopOk, stopOutput) = runScriptWithOutput(
-            executableName: "svc.sh",
-            arguments: ["stop"],
-            workingDirectory: dir,
-            timeout: 15,
-            logTag: "svc.sh stop")
-        log("RunnerLifecycle > STOP step1 done: ok=\(stopOk) output=\(stopOutput.prefix(300))")
+            executableName: "svc.sh", arguments: ["stop"],
+            workingDirectory: dir, timeout: 15, logTag: "svc.sh stop")
+        logStep("STOP", "step1 done: ok=\(stopOk) output=\(stopOutput.prefix(300))")
         if isCorruptInstall(output: stopOutput) {
-            log("RunnerLifecycle > STOP RETURNING .corruptInstall for \(runner.runnerName)")
+            logStep("STOP", "RETURNING .corruptInstall for \(runner.runnerName)")
             return .corruptInstall
         }
-        log("RunnerLifecycle > STOP step2: svc.sh uninstall")
+
+        logStep("STOP", "step2: svc.sh uninstall")
         let (uninstallOk, uninstallOutput) = runScriptWithOutput(
-            executableName: "svc.sh",
-            arguments: ["uninstall"],
-            workingDirectory: dir,
-            timeout: 15,
-            logTag: "svc.sh uninstall")
-        log("RunnerLifecycle > STOP step2 done: ok=\(uninstallOk) output=\(uninstallOutput.prefix(300))")
+            executableName: "svc.sh", arguments: ["uninstall"],
+            workingDirectory: dir, timeout: 15, logTag: "svc.sh uninstall")
+        logStep("STOP", "step2 done: ok=\(uninstallOk) output=\(uninstallOutput.prefix(300))")
         if stopOk {
-            log("RunnerLifecycle > STOP RETURNING .success for \(runner.runnerName)")
+            logStep("STOP", "RETURNING .success for \(runner.runnerName)")
             return .success
         }
         let msg = stopOutput.components(separatedBy: "\n")
             .first(where: { !$0.trimmingCharacters(in: .whitespaces).isEmpty }) ?? "Failed to stop"
-        log("RunnerLifecycle > STOP RETURNING .failed(\(msg)) for \(runner.runnerName)")
+        logStep("STOP", "RETURNING .failed(\(msg)) for \(runner.runnerName)")
         return .failed(msg)
     }
 
     // MARK: - Remove
 
-    /// Performs the remove operation.
+    /// Fully removes a runner: uninstalls the launchd service, deregisters it from GitHub via
+    /// `config.sh remove` (falling back to the API DELETE endpoint if the script fails),
+    /// deletes the install directory, and removes the LaunchAgent plist.
+    ///
+    /// Returns `true` if the runner was successfully deregistered from GitHub (either via
+    /// `config.sh` or the API fallback). Returns `false` if deregistration failed.
+    ///
+    /// - Note: Returns `Bool` rather than `LifecycleResult` because removal is a best-effort
+    ///   multi-step operation — partial failures (e.g. corrupt install) are handled internally
+    ///   via the API fallback rather than surfaced as distinct result cases.
     @discardableResult
     func remove(runner: RunnerModel) -> Bool {
         let ip = runner.installPath ?? "nil"
         let gh = runner.gitHubUrl ?? "nil"
-        log("RunnerLifecycle > REMOVE called runner=\(runner.runnerName) installPath=\(ip) gitHubUrl=\(gh)")
+        logStep("REMOVE", "called runner=\(runner.runnerName) installPath=\(ip) gitHubUrl=\(gh)")
         guard let path = runner.installPath else {
-            log("RunnerLifecycle > REMOVE abort — no installPath for \(runner.runnerName)")
+            logStep("REMOVE", "abort — no installPath for \(runner.runnerName)")
             return false
         }
         let dir = URL(fileURLWithPath: path)
-        log("RunnerLifecycle > REMOVE step1: svc.sh uninstall")
-        let (svcOk, _) = runScriptWithOutput(executableName: "svc.sh", arguments: ["uninstall"], workingDirectory: dir, timeout: 30, logTag: "svc.sh uninstall")
-        log("RunnerLifecycle > REMOVE step1 result=\(svcOk) (non-fatal)")
+
+        logStep("REMOVE", "step1: svc.sh uninstall")
+        let (svcOk, _) = runScriptWithOutput(
+            executableName: "svc.sh", arguments: ["uninstall"],
+            workingDirectory: dir, timeout: 30, logTag: "svc.sh uninstall")
+        logStep("REMOVE", "step1 result=\(svcOk) (non-fatal)")
+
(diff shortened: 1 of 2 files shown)


eonist:
## Batch 9 Review

Overall this is a solid batch — the `logStep` extraction is the headline win and the deadlock warnings are overdue. A few things worth discussing below.

---

### `RunnerLifecycleService.swift`

#### ✅ `logStep(_:_:)` extraction

This is the right call. The old pattern of embedding `"RunnerLifecycle > "` literally in every single log call was noise in the call sites and made the prefix impossible to change without a mass find/replace. The helper is small, private, and does exactly one thing. No concerns here.

One nuance: `logStep` takes the tag as a freeform `String`, which means callers can pass anything — including the interpolated `"runScript [\(logTag)]"` form already present in the diff. That's fine for now, but if you ever want structured log filtering (e.g. OSLog subsystem/category), the tag would need to become a typed enum or `StaticString`. Worth a `// TODO:` if that's on the roadmap.

#### ⚠️ `stop()` — silent `uninstallOk` drop

This is the most significant behavioural observation in the whole batch. Look at the `stop()` return logic:

```swift
let (stopOk, stopOutput) = runScriptWithOutput(... "stop" ...)
// isCorruptInstall check on stopOutput ✅
let (uninstallOk, uninstallOutput) = runScriptWithOutput(... "uninstall" ...)
// uninstallOk and uninstallOutput are never checked
if stopOk {
    return .success
}
```

The `uninstall` result is captured but completely ignored — there's no corrupt check on `uninstallOutput`, and `uninstallOk` plays no role in the return value. The old code had exactly the same gap; this PR preserved it rather than fixed it. That's arguably fine scope-wise (this is a docs+lint batch, not a logic batch), but it should at least be documented. The `stop()` doc currently says:

> Returns `.corruptInstall` if the **stop step** detects a broken installation

…which is accurate but incomplete. Consider adding a `- Note:` that the uninstall step result is intentionally not surfaced (or is it intentional? worth confirming).

#### ✅ `remove()` — `Bool` inconsistency documented

The `- Note:` on `remove()` explaining why it returns `Bool` instead of `LifecycleResult` is exactly the right treatment. This asymmetry was previously invisible to anyone reading the type signature.

#### 🔍 `remove()` — `step4` deletion condition

```swift
if removeOk || (!cfgOk && runner.agentId != nil) {
    // delete install dir
}
```

This condition says: delete the directory if either (a) deregistration succeeded, or (b) `config.sh` failed but we had an `agentId` (meaning we attempted but may or may not have succeeded via the API). The API path sets `removeOk = apiOk`, so if the API call also failed, `removeOk` is `false` and `!cfgOk && runner.agentId != nil` is `true` — the directory gets deleted even on a failed API deregistration. This may be intentional (clean up locally regardless), but the doc comment says the directory is deleted only after "successful deregistration". The comment and the code disagree. Worth aligning one way or the other.

#### Minor: removed intermediate `local var output`

```swift
// before
let output = result.output
log(...)
return (result.exitCode == 0, output)

// after
logStep(..., "... \(result.output.prefix(500))")
return (result.exitCode == 0, result.output)
```

Fine. Slightly more efficient (one fewer copy), and `result.output` is self-documenting.

---

### `RunnerPollState.swift`

#### ✅ Deadlock warnings

Adding `// ⚠️ main.sync` at the call sites and the class-level doc warning is the right minimum. This pattern is genuinely dangerous — `DispatchQueue.main.sync` from a background thread is fine today, but fragile if the calling context ever changes. The class-level doc helps catch this in future refactors.

A stronger long-term fix would be to pass `scopes` as a parameter into `buildJobState` / `buildGroupState` so the `main.sync` sits at the call site (where the caller controls the threading context) rather than inside the `nonisolated` methods. That's a separate refactor, but the comment as-is is a good interim measure.

#### ✅ `SendableFormatter` — `@unchecked` justification

Moving the "expensive ICU allocation" rationale from the `//` block comment into the `///` doc, and adding the `@unchecked` safety explanation in the same place, is the right consolidation. The old approach had the `why` buried in a block comment that wouldn't show up in Quick Help.

#### 🔍 `iso8601Lock` doc — "always via `withLock`"

The new doc says:
> Always access the formatter via `iso8601Lock.withLock { ... }` to avoid data races.

But `SendableFormatter.iso` is declared `let` (immutable), and `OSAllocatedUnfairLock` holds the formatter by value. The lock protects against concurrent _calls_ to `ISO8601DateFormatter.string(from:)` (which is not thread-safe despite the `let`), not against mutation of the wrapper itself. The doc is correct in intent but slightly imprecise in framing — "to avoid data races on the non-thread-safe formatter" would be more accurate than the implication that it prevents mutation.

#### ✅ `extension RunnerStore` doc

> `RunnerStore` extension that bridges `PollResultBuilder` for the `fetch()` call sites.

This is a genuine improvement over `/// Extension adding functionality to RunnerStore.`. It tells you the architectural role of the extension rather than just restating its existence.

#### ✅ `enrichGroupJobs` doc

> Merges live job data with cached job data, preferring whichever has a conclusion or more complete step information.

Good. "More complete step information" is appropriately vague given the internal heuristic — doesn't over-promise specificity.

---

### Summary

| Item | Verdict |
|---|---|
| `logStep` extraction / `type_body_length` suppression removed | ✅ Approve |
| `stop()` uninstall result silently dropped | ⚠️ Pre-existing gap — needs doc note or fix |
| `remove()` step4 condition vs. doc mismatch | ⚠️ Minor — align comment with actual behaviour |
| Deadlock warnings on `main.sync` | ✅ Approve |
| `iso8601Lock` doc precision | 🔍 Nitpick — consider tightening wording |
| All other doc upgrades | ✅ Approve |

The two `⚠️` items are pre-existing logic questions rather than regressions introduced here, so they don't block the merge — but if they're not addressed now, they should become issues tracked under #1038.

eonist:
## Batch 9 — Deep Review

Overall: this is a well-scoped, clean batch. The `logStep` extraction is the standout improvement — it removes a category of future drift entirely. No blocking issues. A few things worth calling out explicitly.

---

### `RunnerLifecycleService.swift`

#### ✅ `logStep(_:_:)` extraction — correct and complete

Every single `log("RunnerLifecycle > ...")` call is replaced. The helper doc accurately describes the format string. One small note: the `logTag` variant inside `runScriptWithOutput` now embeds brackets in the tag itself — `logStep("runScript [\(logTag)]", ...)` — which means the emitted line is `RunnerLifecycle > runScript [svc.sh install]: ...`. That's fine and readable, but it means the `tag` parameter is not a plain identifier anymore; it's a formatted string. The doc for `logStep` says "tag" but doesn't document that. Not a bug, just worth being aware of if you ever grep logs by tag.

#### ✅ `LifecycleResult` case docs — accurate

The addition of "The caller should prompt the user to reinstall the runner" on `.corruptInstall` is exactly the kind of actionable guidance that was missing. Good.

#### ✅ `remove()` — `Bool` vs `LifecycleResult` inconsistency acknowledged

The `- Note:` in the doc is a solid explanation. One thing the note doesn't mention is that `remove()` can still return `false` even when the install directory was already deleted (i.e. partial success is silently swallowed). That may be intentional given the best-effort semantics, but a future reader might wonder. Consider adding "Callers should treat `false` as a signal to surface a warning, not necessarily a full rollback." if that matches intended behaviour.

#### ⚠️ `stop()` — `uninstallOk` is computed but unused

In `stop()`, after calling `svc.sh uninstall`, `uninstallOutput` is bound but only `stopOk` (the result of the *stop* step) gates the return. `uninstallOk` is declared but never checked. This was present before this PR too, but now that the docs describe the two-step sequence explicitly ("Stops and uninstalls..."), a reader will expect the uninstall result to matter. It doesn't. That's either intentional (uninstall is fire-and-forget after a successful stop) or a latent bug. Worth a `// uninstall is best-effort; result intentionally ignored` comment on the `let (uninstallOk, uninstallOutput)` line to make it clear, or rename to `let (_, _)` if truly unused.

#### ✅ `isCorruptInstall` — corrupt-check log removed from `start()`

Previously `start()` logged `corruptCheck: isCorrupt=\(...)` before calling `isCorruptInstall`, then `isCorruptInstall` itself also logged. That was double-logging. The pre-call log lines were removed here; `isCorruptInstall` still logs its own result. Clean.

#### ✅ `runScriptWithOutput` — intermediate `output` variable removed

The `let output = result.output` followed by `log(... output ...)` then `return (_, output)` is now collapsed to `return (result.exitCode == 0, result.output)` with the log inlined. Minimal and correct.

---

### `RunnerPollState.swift`

#### ✅ `SendableFormatter` doc — ICU rationale promoted to `///`

The rationale ("expensive to allocate, loads ICU calendars") is now in the doc-comment where it belongs. The `@unchecked` justification ("no two threads ever call the formatter concurrently") is accurate given `iso8601Lock`. Good.

#### ✅ `iso8601Lock` doc — "always use `withLock`" added

This is the right enforcement comment. The previous `/// Lock for the formatter.` was nearly useless.

#### ⚠️ `DispatchQueue.main.sync` deadlock warning — placement is good, but incomplete

The class-level extension doc now warns clearly: "Do not call these methods from the main thread as `main.sync` from the main thread will deadlock." And both `buildJobState` and `buildGroupState` have inline `// ⚠️ main.sync` comments.

However, `backfillSteps` and `enrichGroupJobs` are also `nonisolated` methods in the same extension, and they don't have the same warning — because they don't call `main.sync` themselves. But since the extension-level doc applies to *all* methods, a reader might assume `backfillSteps` also does a `main.sync` hop when it doesn't. Consider tightening the extension doc to "Some methods in this extension..." rather than implying all of them do the main-hop.

#### ✅ `enrichGroupJobs` doc — "preferring whichever has a conclusion or more complete step information"

This is a good description. One follow-up question for a future review: the actual merge strategy (how "more complete" is decided — step count? conclusion presence?) is encapsulated somewhere in `PollResultBuilder` presumably. The doc doesn't need to repeat that, but confirming it's documented there would close the loop.

#### ✅ `scopeFromActionGroup` doc — fallback chain documented

"first trying `group.repo`, then falling back to parsing the `htmlUrl` of the first run. Returns an empty string if neither is available." Accurate per the diff.

---

### Issue completeness check (vs. #1074)

| Issue item | Done? |
|---|---|
| Remove `swiftlint:disable:next type_body_length` | ✅ Gone — `logStep` extraction did it organically |
| Replace filler `LifecycleResult` case docs | ✅ |
| Replace struct filler doc | ✅ |
| `shared` singleton wording | ✅ |
| `scopeFromGitHubUrl` doc with examples | ✅ |
| `remove()` `Bool` inconsistency comment | ✅ (`- Note:` in doc) |
| `SendableFormatter` + `iso8601Lock` — move `//` rationale to `///` | ✅ |
| `extension RunnerStore` filler replaced | ✅ |
| `DispatchQueue.main.sync` deadlock warning | ✅ (both inline + extension doc) |

All 9 issue items are addressed. The `uninstallOk` unused-result observation above is pre-existing, not introduced here.

---

### Verdict

✅ **Approve with two non-blocking suggestions:**
1. Add `// uninstall result intentionally ignored` (or `let (_, _)`) on the `stop()` uninstall call.
2. Tighten the extension-level doc to clarify that only `buildJobState`/`buildGroupState` do the `main.sync` hop, not the whole extension.

Neither blocks merge.

eonist:
## Deep Review — Batch 9 (PR #1076 / Issue #1074)

Overall verdict: **ready to merge** with one latent correctness concern in `stop()` worth addressing before or after merge, and a few documentation/naming nits.

---

### `RunnerLifecycleService.swift`

#### ✅ `logStep(_:_:)` extraction — correct and well-motivated

The helper achieves its goal cleanly. Every call site now reads `logStep("START", "step1: svc.sh install")` rather than repeating the `"RunnerLifecycle > "` prefix, and the body-line-count drops organically, eliminating the `swiftlint:disable:next type_body_length` suppression. The helper's own doc (`/// Emits a structured log line…`) accurately describes what it does. No issues here.

One minor nit: `logStep` is declared at the bottom of the struct after `runScriptWithOutput`. Since it's called from every public method, moving it to just after `private init()` would match the "most-depended-on helpers near the top" convention — but this is cosmetic only.

#### ⚠️ `stop()` — silent discard of `uninstallOutput`'s corrupt-check

This is the most substantive finding in the PR.

The original `stop()` ran `svc.sh stop` then `svc.sh uninstall`. It checked `stopOutput` for a corrupt install but not `uninstallOutput`. The new code preserves exactly this asymmetry:

```swift
// step1: svc.sh stop
let (stopOk, stopOutput) = runScriptWithOutput(...)
if isCorruptInstall(output: stopOutput) { return .corruptInstall }   // ← checked ✅

// step2: svc.sh uninstall
let (uninstallOk, uninstallOutput) = runScriptWithOutput(...)        // ← uninstallOutput never checked ❌
// ...
if stopOk { return .success }
```

`uninstallOk` is read (implicitly — it's captured but unused in the final decision) but `uninstallOutput` is never inspected for a corrupt install signal. Compare this with `start()`, which correctly calls `isCorruptInstall` after **both** `install` and `start` steps. The issue didn't flag this gap, so the PR didn't introduce it — but given that `start()` was fixed to be symmetric, it's worth deciding whether `stop()` should be too. At minimum the `- Note:` on `stop()` could call this out explicitly so the next reader isn't confused by the intentional asymmetry (if it is intentional).

#### ✅ `remove()` — `Bool` return documented

The new `- Note:` on `remove()` explains the `Bool` vs `LifecycleResult` inconsistency well: partial failures are handled internally via the API DELETE fallback, so surfacing them as distinct result cases would be misleading. The doc is accurate and matches the actual logic flow.

One observation: the condition at step 4 is:
```swift
if removeOk || (!cfgOk && runner.agentId != nil) {
```
This means the directory and plist are cleaned up even if `deleteRunnerByID` returned `false` (when `agentId != nil` but the API call failed). That logic was pre-existing, but the new doc doesn't mention it — a follow-up `- Note:` could clarify that cleanup is attempted optimistically even on API failure.

#### ✅ `isCorruptInstall` — doc accurately names the trigger strings

The doc says `missing svc.sh, wrong working directory` as examples, but the actual implementation checks for `"must run from runner root"` and `"install is corrupt"` (lowercased). The new doc on `isCorruptInstall` is slightly more precise but the struct-level doc still says `missing svc.sh` which is an indirect cause, not a trigger string. Very minor — consider aligning the struct-level doc example with the actual string detection logic.

#### ✅ `scopeFromGitHubUrl` examples — correct and useful

The new `Examples:` block is accurate and covers both org and repo cases. The fallback behaviour (return original string unchanged) is correctly documented.

#### ✅ `runScriptWithOutput` — removed redundant `let output` variable

```swift
// before
let output = result.output
log(... output.prefix(500))
return (result.exitCode == 0, output)

// after
log(... result.output.prefix(500))
return (result.exitCode == 0, result.output)
```
Clean removal of a one-use intermediate. No correctness impact.

---

### `RunnerPollState.swift`

#### ✅ `SendableFormatter` — rationale promoted into `///` doc

The moved doc (`ISO8601DateFormatter is expensive to allocate…`) is now visible in Quick Help and Docc rather than buried in a `//` block. The `@unchecked Sendable` safety justification (`no two threads call concurrently due to lock`) is accurate and useful.

#### ✅ `iso8601Lock` — `withLock` access reminder

The new doc (`Always access the formatter via iso8601Lock.withLock { ... }`) is practical and mirrors patterns from the Swift Concurrency community. Good.

#### ✅ `⚠️ main.sync` inline comments — well-placed

Putting `// ⚠️ main.sync — safe only from a background thread; deadlocks if called on main.` directly above both `DispatchQueue.main.sync` call sites is the right move. The class-level doc warning reinforces this at the extension level. 

One thought worth raising as a follow-up (not a blocker): these `DispatchQueue.main.sync` hops exist to read `ScopeStore.shared.scopes` from a `nonisolated` context. With Swift 6 strict concurrency, `ScopeStore.shared` being `@MainActor` means this pattern will generate warnings or errors under `SWIFT_STRICT_CONCURRENCY = complete`. The comment is a good safety net now, but a migration path to `await MainActor.run { ScopeStore.shared.scopes }` would eliminate the deadlock risk entirely. Worth tracking in a follow-up issue if not already.

#### ✅ `enrichGroupJobs` doc — accurate description

`/// Merges live job data with cached job data, preferring whichever has a conclusion or more complete step information.` matches the actual implementation (the function selects the job with a non-nil conclusion or more steps). Correct.

#### ✅ `backfillSteps` — doc accurately describes the filtering condition

The new doc mentions `jobs that have a conclusion but missing or in-progress steps` which aligns with the guard logic in the function body. Good.

---

### Issue #1074 completeness check

| Item from issue | Addressed in PR? |
|---|---|
| Remove `swiftlint:disable:next type_body_length` | ✅ Removed; `logStep` reduces body count |
| Replace filler docs on `LifecycleResult` cases | ✅ |
| `struct` doc on `RunnerLifecycleService` | ✅ |
| `shared` singleton doc | ✅ |
| `scopeFromGitHubUrl` examples | ✅ |
| `remove()` inconsistency comment (`Bool` vs `LifecycleResult`) | ✅ via `- Note:` |
| `RunnerPollState`: `SendableFormatter` + `iso8601Lock` rationale in `///` | ✅ |
| `extension RunnerStore` doc | ✅ |
| `⚠️ DispatchQueue.main.sync` deadlock warnings | ✅ class-level + inline |
| All other filler `/// Performs the X operation.` | ✅ |

Everything scoped in the issue is addressed. The `stop()` asymmetry noted above was pre-existing and out of scope — flagging it here for awareness, not as a blocker.

---

**Summary:** The `logStep` refactor and doc pass are both clean and deliver exactly what the issue asked for. The one item worth a follow-up is the `stop()` corrupt-check asymmetry, and the `main.sync` deadlock risk is a good candidate for a dedicated migration issue once Swift 6 strict concurrency is enabled.

eonist:
## Follow-up review — fixes verified ✅

Read the updated diff. Both corrections are well-executed.

### `stop()` — corrupt-check symmetry

The fix is correct and the intent is now fully legible:

```swift
let (_, uninstallOutput) = runScriptWithOutput(
    // uninstall exit code is intentionally ignored — best-effort after a successful stop.
    ...)
if isCorruptInstall(output: uninstallOutput) { return .corruptInstall }
```

Using `_` for the exit code is the right signal — it tells the next reader immediately that the discard is deliberate, not an oversight. The corrupt-install check being present despite that discard is now the precise behaviour described in the `- Note:`. The log message distinction (`"RETURNING .corruptInstall after stop step"` vs `"after uninstall step"`) also makes post-hoc debugging straightforward. No issues.

### `remove()` — step-4 condition

The `if removeOk || (!cfgOk && runner.agentId != nil)` → `if removeOk` simplification is strictly correct. The old condition would delete local files even when the API DELETE call failed — silently leaving the runner deregistered on GitHub (from a previous attempt or partial state) while also nuking the local install directory. The new gate is unambiguous: files go only when GitHub says the runner is gone.

The `else` log line (`"step4: skipping local cleanup — deregistration failed"`) is the right addition — failed paths are now visible in the log without any extra instrumentation, which matters for debugging removal failures in the field.

The second `- Note:` in the doc is accurate and aligned with the code:
> Local file cleanup is performed only when deregistration succeeds. If both `config.sh` and the API fallback fail, no local files are deleted so the user can retry.

That's a meaningful guarantee to document — users who hit a network failure during removal won't lose their local install silently.

### No new concerns

No issues introduced by the two follow-up commits. PR is good to merge.

eonist:
## Follow-up review — both fixes verified ✅

### `stop()` — corrupt-check symmetry

The fix is correct and the structure is now fully symmetric with `start()`:

```
step1 (stop)      → isCorruptInstall check ✅
step2 (uninstall) → isCorruptInstall check ✅  ← new
```

Renaming `(uninstallOk, uninstallOutput)` → `(_, uninstallOutput)` is the right call — it makes the intentional discard loud and Swift-idiomatic. The inline comment `// uninstall exit code is intentionally ignored — best-effort after a successful stop.` inside the binding is a nice belt-and-suspenders touch; anyone reading the call site gets the `why` without having to look up the doc.

The `- Note:` in the doc is accurate and well-scoped: it explains *what* is discarded and *why*, and explicitly confirms the corrupt signal still fires. No concerns.

### `remove()` — step-4 condition

`if removeOk || (!cfgOk && runner.agentId != nil)` → `if removeOk` is the correct simplification. The old branch was over-broad: it deleted local files even when both deregistration paths failed, leaving the runner in a partially-cleaned state with no GitHub-side record. The new behaviour (retain local files on failure so the user can retry) is the right UX contract and the second `- Note:` in the doc now matches the code exactly.

The `else` log line (`step4: skipping local cleanup — deregistration failed`) is a good add — failed paths that produce no log output are the hardest to debug in the field.

### One minor observation

With `if removeOk` as the step-4 gate, there's now a case worth thinking about: `config.sh` fails, the API fallback succeeds (`removeOk = true`), and step 4 proceeds to delete the install directory — but `svc.sh uninstall` in step 1 might have also failed (`svcOk = false`). The LaunchAgent plist could still be present after the directory is gone. `deleteLaunchAgentPlist` is already called unconditionally inside the `if removeOk` block, so this should be fine in practice. Just confirming that `deleteLaunchAgentPlist` doesn't depend on the install directory existing — which looking at the implementation (it scans `~/Library/LaunchAgents` by name pattern) it doesn't. ✅

**Verdict: both fixes are correct, the code and docs are now fully aligned. Approve from my side.**

Issue #546, opened 2026-05-20
Add terminal call from the app


eonist:
How can we call terminal with a command from the app? what code would we have to add? 

I think we should also add a test command button to our failure hock in our scopes detail view. so users can test quickly if they wired things up right etc. and also for us to debug. so lets just get that working first. then later add this to actual failure events etc. 

ref:

https://github.com/eoncode/runner-bar/issues/544%20and%20https://github.com/eoncode/runner-bar/issues/340#issuecomment-4421630478%20and%20https://github.com/eoncode/runner-bar/issues/408

eonist:
entitlement discussions

[do you agree with this_ (1).md](https://github.com/user-attachments/files/28036356/do.you.agree.with.this_.1.md)

[do you agree with this_.md](https://github.com/user-attachments/files/28036358/do.you.agree.with.this_.md)

Issue #927, opened 2026-05-25
Move ScopeEntry.swift to RunnerBarCore/Scope/


eonist:
Part of #921

## What

Move `Sources/RunnerBar/Scope/ScopeEntry.swift` → `Sources/RunnerBarCore/Scope/ScopeEntry.swift`.

## Why it's movable

- Imports only `Foundation`
- Plain model type: `struct ScopeEntry: Identifiable, Codable, Equatable`
- No logic, no singletons, no UI references
- `ScopePreferencesStore` already lives in Core — `ScopeEntry` is its natural companion

## Blast radius

This is the smallest move. `ScopeEntry` is a data type used throughout the app but because of `@_exported import RunnerBarCore`, all callers get it for free once it's in Core:

| Caller | Notes |
|---|---|
| `ScopeStore.swift` | Uses `[ScopeEntry]` for `@Published var entries` and persistence — no import change |
| `ScopeDetailView.swift` | Reads `ScopeEntry` properties in the view body — no import change |
| Any Preferences view listing scopes | Gets it transitively — no import change |

## Mechanical work

1. Delete `Sources/RunnerBar/Scope/ScopeEntry.swift`
2. Create `Sources/RunnerBarCore/Scope/ScopeEntry.swift` with identical content
3. Make public: `struct ScopeEntry` and all its members

## Symbols to make `public`

```swift
public struct ScopeEntry: Identifiable, Codable, Equatable {
    public let id: UUID
    public var scope: String
    public var isEnabled: Bool
    public init(scope: String, isEnabled: Bool = true)
}
```

Issue #926, opened 2026-05-25
Move LogFetcher.swift to RunnerBarCore/Services/


eonist:
Part of #921

## What

Move `Sources/RunnerBar/Services/LogFetcher.swift` → `Sources/RunnerBarCore/Services/LogFetcher.swift`.

## Why it's movable

- Imports only `Foundation` + `RunnerBarCore`
- All symbols are free-standing functions — no class, no app singletons
- Uses only `Process`, `FileManager`, `/usr/bin/unzip` — pure Foundation/POSIX
- `ghRaw()` is a private file-level helper that uses `ghBinaryPath()` (already in Core)
- Operates on `WorkflowActionGroup` which is a Core type

## Blast radius

| Caller | Notes |
|---|---|
| `FailureHookRunner.swift` | Calls `fetchJobLog(jobID:scope:)` on a background thread — no import change |
| Any log detail view | Gets `fetchJobLog` / `fetchActionLogs` transitively via `@_exported` — no import change |

## Mechanical work

1. Delete `Sources/RunnerBar/Services/LogFetcher.swift`
2. Create `Sources/RunnerBarCore/Services/LogFetcher.swift` with identical content minus `import RunnerBarCore`
3. Make public: `func fetchJobLog(jobID:scope:)`, `func fetchActionLogs(group:)`, `func unzipLogs(_:)`

## Symbols to make `public`

```swift
public func fetchJobLog(jobID: Int, scope: String) -> String?
public func fetchActionLogs(group: WorkflowActionGroup) -> String?
public func unzipLogs(_ zipData: Data) -> [(name: String, text: String)]
```

## Note

`ghRaw()` is a private file-scope helper that mirrors `fetchStepLog`'s CLI path in `GitHub.swift`. Once in Core, consider consolidating it with `ghAPICLI` or the transport layer to avoid duplication.

Issue #922, opened 2026-05-25
Move ProcessRunner.swift to RunnerBarCore/Services/


eonist:
Part of #921

## What

Move `Sources/RunnerBar/Services/ProcessRunner.swift` → `Sources/RunnerBarCore/Services/ProcessRunner.swift`.

## Why it's movable

- Imports only `Foundation` — zero AppKit, SwiftUI, Combine, or Keychain
- Pure value type (`enum ProcessRunner`) with a single static method
- No reference to any app-layer class or singleton

## Blast radius

This is the **most foundational move** — 4 other files call `ProcessRunner.run()` directly. Because of `@_exported import RunnerBarCore` in `Exports.swift`, none of them need import changes:

| Caller | Notes |
|---|---|
| `GitHubCLITransport.swift` | Calls `ProcessRunner.run()` in `runGHProcess()` — no change needed |
| `RunnerLifecycleService.swift` | Calls `ProcessRunner.run()` in `runScriptWithOutput()` — no change needed |
| `LocalRunnerScanner.swift` | Calls `ProcessRunner.run()` in `scanLiveServices()` — also being migrated (#924) |
| `Auth.swift` | Calls `ProcessRunner.run()` for `gh auth token` — no change needed |

## Mechanical work

1. Delete `Sources/RunnerBar/Services/ProcessRunner.swift`
2. Create `Sources/RunnerBarCore/Services/ProcessRunner.swift` with identical content minus `import RunnerBarCore`
3. Make `public`: `enum ProcessRunner`, `struct Result`, `let data`, `let exitCode`, `var output`, `static func run(…)`

## Symbols to make `public`

```swift
public enum ProcessRunner {
    public struct Result {
        public let data: Data?
        public let exitCode: Int32
        public var output: String { … }
    }
    public static func run(
        executableURL: URL,
        arguments: [String],
        stdin: Data? = nil,
        workingDirectory: URL? = nil,
        mergeStderr: Bool = false,
        timeout: TimeInterval = 20
    ) -> Result
}
```

PR #1083, opened 2026-06-01
Batch 11 — Scope + Services: docs + SwiftLint cleanup

### LINKED ISSUE #1082: Batch 11 — Scope + Services: docs + SwiftLint cleanup
Part of the ongoing codebase review tracked in #1038.

## Files

| File | Changes needed |
|------|----------------|
| `Sources/RunnerBar/Scope/ScopeEntry.swift` | Tombstone — upgrade `//` comment to `///` doc |
| `Sources/RunnerBar/Scope/ScopeStore.swift` | Remove `swiftlint:disable/enable orphaned_doc_comment`; fix tautological `entriesKey`/`legacyKey` property docs; document or remove redundant `objectWillChange.send()` in `entries.willSet` |
| `Sources/RunnerBar/Services/LogFetcher.swift` | Tombstone — remove `swiftlint:disable:next missing_docs`, replace with `///` tombstone doc |
| `Sources/RunnerBar/Services/LoginItem.swift` | Fix stale `PopoverMainView` caller attribution → `PanelMainView` |
| `Sources/RunnerBar/Services/ProcessRunner.swift` | Tombstone — remove `swiftlint:disable:next missing_docs`, replace with `///` tombstone doc |
| `Sources/RunnerBar/Services/TerminalLauncher.swift` | Replace autogenerated stub doc with real `///` class doc; fold `// #546` block comment into it |

## Goals
- Remove all SwiftLint suppression clauses
- Replace tautological / stub docs with meaningful `///` comments
- No logic changes

eonist:
## **User description**
Part of the ongoing codebase review tracked in #1038. Closes #1082.

## Files reviewed

| File | Changes |
|------|--------|
| `Scope/ScopeEntry.swift` | Upgraded `//` tombstone comment to `///` doc |
| `Scope/ScopeStore.swift` | Removed `swiftlint:disable/enable orphaned_doc_comment`; fixed tautological `entriesKey`/`legacyKey` property docs; removed redundant `willSet { objectWillChange.send() }` on `@Published entries` (SwiftLint was complaining about the orphaned doc above it); documented the intentional `objectWillChange.send()` in `setEnabled` |
| `Services/LogFetcher.swift` | Removed `swiftlint:disable:next missing_docs`; replaced with `///` tombstone doc |
| `Services/ProcessRunner.swift` | Removed `swiftlint:disable:next missing_docs`; replaced with `///` tombstone doc |
| `Services/LoginItem.swift` | Fixed stale caller attribution `PopoverMainView` → `SettingsView` |
| `Services/TerminalLauncher.swift` | Replaced autogenerated `/// Enumerates possible values` stub with real class `///` doc; folded `// #546` block comment into it; removed redundant `//` block |

## No logic changes

___

## **CodeAnt-AI Description**
Clarify inline documentation and remove outdated lint workarounds

### What Changed
- Replaced stale empty-file comments with proper notes for moved Scope and Service files
- Updated the login-item note to point to the Settings screen instead of an old view name
- Reworked the Terminal launcher notes to describe what it does and how it handles command escaping
- Cleaned up ScopeStore comments so the saved keys and enable/disable behavior are described clearly, while keeping the same app behavior

### Impact
`✅ Clearer code comments for ongoing maintenance`
`✅ Fewer outdated references in the codebase`
`✅ Easier review of scope and login-item behavior`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **Refactor**
  * Reorganized internal code modules for improved maintainability and structure.
  * Simplified property notification logic for better consistency.

* **Documentation**
  * Enhanced internal documentation for clarity on feature behavior and implementation details.

### DIFF
--- a/Sources/RunnerBar/Scope/ScopeEntry.swift
+++ b/Sources/RunnerBar/Scope/ScopeEntry.swift
@@ -1,5 +1,4 @@
 // ScopeEntry.swift
 // RunnerBar
-
-/// Moved to `RunnerBarCore/Scope/ScopeEntry.swift` — see #927.
-/// This file is intentionally empty; the type is re-exported via `Exports.swift`.
+// Tombstone — type moved to RunnerBarCore/Scope/ScopeEntry.swift (see #927).
+// This file is intentionally empty; the type is re-exported via Exports.swift.
--- a/Sources/RunnerBar/Scope/ScopeEntry.swift
+++ b/Sources/RunnerBar/Scope/ScopeEntry.swift
@@ -1,4 +1,5 @@
 // ScopeEntry.swift
 // RunnerBar
-// Moved to RunnerBarCore/Scope/ScopeEntry.swift — see #927.
-// This file is intentionally empty; the type is re-exported via Exports.swift.
+
+/// Moved to `RunnerBarCore/Scope/ScopeEntry.swift` — see #927.
+/// This file is intentionally empty; the type is re-exported via `Exports.swift`.
--- a/Sources/RunnerBar/Scope/ScopeStore.swift
+++ b/Sources/RunnerBar/Scope/ScopeStore.swift
@@ -112,8 +112,12 @@ final class ScopeStore: ObservableObject {
 
     /// Toggles the `isEnabled` flag for the entry with the given ID.
     /// Does NOT send `didMutate` — enable/disable is not a structural change.
-    /// Publishes `objectWillChange` explicitly so `RunnerStore`'s Combine
-    /// subscription triggers a polling restart on toggle.
+    ///
+    /// `ScopeEntry` is a struct, so `entries[idx].isEnabled = enabled` replaces
+    /// the array value and `@Published` fires `objectWillChange` automatically.
+    /// The explicit `objectWillChange.send()` below is a belt-and-suspenders call
+    /// that ensures `RunnerStore`'s Combine subscription triggers a polling restart
+    /// even if the value-type contract changes in future.
     func setEnabled(_ id: UUID, _ enabled: Bool) {
         guard let idx = entries.firstIndex(where: { $0.id == id }) else { return }
         entries[idx].isEnabled = enabled
--- a/Sources/RunnerBar/Scope/ScopeStore.swift
+++ b/Sources/RunnerBar/Scope/ScopeStore.swift
@@ -1,6 +1,5 @@
 // ScopeStore.swift
 // RunnerBar
-// swiftlint:disable orphaned_doc_comment
 import Combine
 import Foundation
 
@@ -18,9 +17,9 @@ final class ScopeStore: ObservableObject {
     /// Shared singleton — single source of truth for all scope operations.
     static let shared = ScopeStore()
 
-    /// The entriesKey constant.
+    /// `UserDefaults` key for the JSON-encoded `[ScopeEntry]` array.
     private let entriesKey = "scopeEntries"
-    /// The legacyKey constant.
+    /// `UserDefaults` key for the legacy plain `[String]` scopes array, kept for migration only.
     private let legacyKey = "scopes"
 
     /// Emits after every structural mutation (add / remove). Callers subscribe and
@@ -29,10 +28,8 @@ final class ScopeStore: ObservableObject {
     let didMutate = PassthroughSubject<Void, Never>()
 
     /// All scope entries, persisted as JSON in `UserDefaults`.
-    /// Publishes `objectWillChange` before every write so observing views update.
-    @Published private(set) var entries: [ScopeEntry] = [] {
-        willSet { objectWillChange.send() }
-    }
+    /// `private(set)` — mutate only via `add(_:)`, `remove(id:)`, and `setEnabled(_:_:)`.
+    @Published private(set) var entries: [ScopeEntry] = []
 
     /// Scopes that are currently enabled — used by `RunnerStore` for polling.
     var activeScopes: [String] { entries.filter(\.isEnabled).map(\.scope) }
@@ -115,13 +112,13 @@ final class ScopeStore: ObservableObject {
 
     /// Toggles the `isEnabled` flag for the entry with the given ID.
     /// Does NOT send `didMutate` — enable/disable is not a structural change.
+    /// Publishes `objectWillChange` explicitly so `RunnerStore`'s Combine
+    /// subscription triggers a polling restart on toggle.
     func setEnabled(_ id: UUID, _ enabled: Bool) {
         guard let idx = entries.firstIndex(where: { $0.id == id }) else { return }
         entries[idx].isEnabled = enabled
         persist()
         log("ScopeStore › scope \(entries[idx].scope) isEnabled=\(enabled)")
-        // Publish so RunnerStore's Combine subscription triggers a polling restart.
         objectWillChange.send()
     }
 }
-// swiftlint:enable orphaned_doc_comment
--- a/Sources/RunnerBar/Services/LogFetcher.swift
+++ b/Sources/RunnerBar/Services/LogFetcher.swift
@@ -1,5 +1,4 @@
 // LogFetcher.swift
 // RunnerBar
-
-/// Moved to `RunnerBarCore/Services/LogFetcher.swift` — see #926.
-/// This file is intentionally empty; the symbols are re-exported via `Exports.swift`.
+// Tombstone — type moved to RunnerBarCore/Services/LogFetcher.swift (see #926).
+// This file is intentionally empty; the symbols are re-exported via Exports.swift.
--- a/Sources/RunnerBar/Services/LogFetcher.swift
+++ b/Sources/RunnerBar/Services/LogFetcher.swift
@@ -1,5 +1,5 @@
 // LogFetcher.swift
 // RunnerBar
-// swiftlint:disable:next missing_docs
-// Moved to RunnerBarCore/Services/LogFetcher.swift — see #926.
-// This file is intentionally empty; the symbols are re-exported via Exports.swift.
+
+/// Moved to `RunnerBarCore/Services/LogFetcher.swift` — see #926.
+/// This file is intentionally empty; the symbols are re-exported via `Exports.swift`.
--- a/Sources/RunnerBar/Services/LoginItem.swift
+++ b/Sources/RunnerBar/Services/LoginItem.swift
@@ -12,8 +12,8 @@ enum LoginItem {
     }
 
     /// Registers or unregisters launch-at-login based on `enabled`.
-    /// Called by `PopoverMainView` from the two-argument `onChange(of:)` form,
-    /// which supplies the new toggle value directly.
+    /// Called from the login-item toggle in `SettingsView` via the two-argument
+    /// `onChange(of:)` form, which supplies the new toggle value directly.
     /// Errors are logged to stderr but otherwise swallowed — failure is non-fatal
     /// since the checkbox UI will simply reflect the unchanged state on next read.
     static func setEnabled(_ enabled: Bool) {
--- a/Sources/RunnerBar/Services/ProcessRunner.swift
+++ b/Sources/RunnerBar/Services/ProcessRunner.swift
@@ -1,5 +1,4 @@
 // ProcessRunner.swift
 // RunnerBar
-
-/// Moved to `RunnerBarCore/Services/ProcessRunner.swift` — see #922.
-/// This file is intentionally empty; the type is re-exported via `Exports.swift`.
+// Tombstone — type moved to RunnerBarCore/Services/ProcessRunner.swift (see #922).
+// This file is intentionally empty; the type is re-exported via Exports.swift.
--- a/Sources/RunnerBar/Services/ProcessRunner.swift
+++ b/Sources/RunnerBar/Services/ProcessRunner.swift
@@ -1,5 +1,5 @@
 // ProcessRunner.swift
 // RunnerBar
-// swiftlint:disable:next missing_docs
-// Moved to RunnerBarCore/Services/ProcessRunner.swift — see #922.
-// This file is intentionally empty; the type is re-exported via Exports.swift.
+
+/// Moved to `RunnerBarCore/Services/ProcessRunner.swift` — see #922.
+/// This file is intentionally empty; the type is re-exported via `Exports.swift`.
--- a/Sources/RunnerBar/Services/TerminalLauncher.swift
+++ b/Sources/RunnerBar/Services/TerminalLauncher.swift
@@ -2,8 +2,6 @@
 // RunnerBar
 import Foundation
 
-// MARK: - TerminalLauncher
-
 /// Opens a Terminal.app window and runs a shell command via AppleScript (`do script`).
 ///
 /// Uses `NSAppleScript` — requires no entitlements on an unsandboxed app.
--- a/Sources/RunnerBar/Services/TerminalLauncher.swift
+++ b/Sources/RunnerBar/Services/TerminalLauncher.swift
@@ -4,11 +4,11 @@ import Foundation
 
 // MARK: - TerminalLauncher
 
-// #546: Opens a normal Terminal.app window and runs the given command via AppleScript.
-//
-// Uses NSAppleScript + `do script` — requires no entitlements on an unsandboxed app.
-// Escapes backslashes, double quotes, and newlines before embedding in the AppleScript string.
-/// Enumerates possible values for TerminalLauncher.
+/// Opens a Terminal.app window and runs a shell command via AppleScript (`do script`).
+///
+/// Uses `NSAppleScript` — requires no entitlements on an unsandboxed app.
+/// Backslashes, double quotes, and newlines in the command are escaped before
+/// embedding in the AppleScript string. Tracked in #546.
 enum TerminalLauncher {
     /// Opens Terminal.app and runs `command` in a new window.
     static func open(command: String) {


eonist:
## Review — Batch 11: Scope + Services docs + SwiftLint cleanup

Overall this is a clean, well-scoped batch. No logic changes, goals are clear, and most of the decisions are correct. A few things worth examining closely:

---

### 🔴 `orphaned_doc_comment` — the core contradiction in this PR

Qodo and Greptile both flagged this, but I want to be precise about *why* it's a contradiction, not just a lint risk.

The whole point of `ScopeStore.swift` in this batch was to **remove** the `// swiftlint:disable/enable orphaned_doc_comment` wrapper — and that was correct to do, because the wrapper existed only to silence lint noise caused by the `willSet { objectWillChange.send() }` doc being orphaned. Removing the redundant `willSet` eliminated the root cause. Good.

But then the tombstone files (`ScopeEntry.swift`, `LogFetcher.swift`, `ProcessRunner.swift`) now introduce the **exact same pattern** — `///` lines with no following declaration — in files that will never have a declaration. This means either:
- CI fails on `swiftlint lint --strict`, or
- A new `swiftlint:disable orphaned_doc_comment` has to be added to each file

Either outcome partially undoes what this batch set out to achieve. The fix is straightforward: keep these as plain `//` comments. They're tombstone files — their purpose is navigational/archival, not API documentation. `///` is semantically wrong here anyway since there's no symbol to document. `//` with backtick-formatted paths (as used elsewhere in the codebase) is the right tool.

---

### 🟡 `ScopeStore` — redundant `objectWillChange.send()` in `setEnabled`

The removal of `willSet { objectWillChange.send() }` from `@Published entries` is correct — `@Published` already calls `objectWillChange` before each mutation, so the `willSet` body was a double-send. Fine.

But `setEnabled` now does `entries[idx].isEnabled = enabled` (which is a mutation of a `@Published` property — so `objectWillChange` fires automatically), *and then* explicitly calls `objectWillChange.send()` again at the end. That's still a double-send, just in a different place. The doc comment added in this PR says:

> "Publishes `objectWillChange` explicitly so `RunnerStore`'s Combine subscription triggers a polling restart on toggle."

The question is whether this explicit second send is actually necessary. If `RunnerStore` is subscribing to `objectWillChange` and observing mutations to `entries`, the automatic publish from `@Published` should already trigger it. The explicit call is only needed if:
1. `RunnerStore` requires a *second* notification for some ordering reason, or
2. The subscription is set up in a way that misses the automatic pre-mutation publish

If neither is true, this is still a double-send that should be removed. If it IS necessary, the doc comment should explain *why* the automatic publish isn't sufficient — "triggers a polling restart" doesn't explain why two sends are needed. Worth verifying before merging, since it's a behavioral edge case disguised as a doc-only PR.

---

### 🟡 `LoginItem.swift` — caller attribution drift

The fix from `PopoverMainView` → `SettingsView` is correct and good. But doc comments that name a specific caller are inherently fragile — this is already the second time this has gone stale. Consider replacing the caller attribution with a description of the *contract* instead:

```swift
// Before (fragile):
/// Called from the login-item toggle in `SettingsView` via the two-argument ...

// More durable:
/// Registers or unregisters launch-at-login based on `enabled`.
/// Intended to be called from a toggle's `onChange(of:)` handler, which supplies
/// the new value directly. Errors are logged but swallowed — failure is non-fatal.
```

This way the doc doesn't need to be updated every time the UI layer is refactored.

---

### 🟢 `TerminalLauncher.swift` — good consolidation

Folding the `// #546` block comment into the `///` class doc is exactly right. The old structure (a real `///` stub below a `//` block comment) was misleading — the `///` was the one SwiftLint would read, but it had no content. The new doc is accurate and covers the important behavioral detail (escaping before AppleScript embedding). No issues here.

---

### 🟢 `ScopeStore` key docs

`"The entriesKey constant."` → `` "`UserDefaults` key for the JSON-encoded `[ScopeEntry]` array." `` is a meaningful improvement. Same for `legacyKey`. These are exactly the kind of tautological docs the issue targeted.

---

### Minor: `ScopeStore` doc update placement

The new doc on `setEnabled` says "Publishes `objectWillChange` explicitly..." but the `objectWillChange.send()` call sits *after* `persist()` and `log(...)`. The doc implies this is the key behavior of the method, but it reads as though it's describing the method's primary purpose. Consider keeping it as an inline `// intentional: second send needed because...` note directly on the call site, and leaving the `///` doc focused on the method's contract (toggle `isEnabled` + persist).

---

**Summary**: Two things to address before merge — the tombstone `///` → `//` revert, and a sanity check on the double `objectWillChange.send()` in `setEnabled`. Everything else is solid.

eonist:
## Deep review — Batch 11 (#1083)

Read the full diff + all existing bot comments before writing this. Deduplication note: Qodo and Greptile both flagged the tombstone `orphaned_doc_comment` issue. That's the real blocker and I'll go deeper on it. Everything else below is either not covered or given surface-level treatment by the bots.

---

### 🔴 Blocker — tombstone `///` will fail CI

The three tombstone stubs (`ScopeEntry.swift`, `LogFetcher.swift`, `ProcessRunner.swift`) now end with standalone `///` lines that have no declaration beneath them. SwiftLint's `orphaned_doc_comment` rule fires on exactly this pattern, and `.swiftlint.yml` confirms it is **not** in `disabled_rules`. The CI workflow runs `swiftlint lint --strict --reporter github-actions-logging`, which promotes every warning to an error. This PR cannot merge as-is — the SwiftLint job will fail.

The fix requires a deliberate choice between two legitimate approaches:

**Option A — Stay on plain `//` (simplest, no lint risk):**
```swift
// ScopeEntry.swift
// RunnerBar
// Tombstone — type moved to RunnerBarCore/Scope/ScopeEntry.swift (see #927).
// This file is intentionally empty; the type is re-exported via Exports.swift.
```
`//` never triggers `orphaned_doc_comment`. This is what the files had before.

**Option B — Keep `///` but add a file-level suppress pair:**
```swift
// ScopeEntry.swift
// RunnerBar
// swiftlint:disable orphaned_doc_comment
/// Tombstone — type moved to `RunnerBarCore/Scope/ScopeEntry.swift` — see #927.
/// This file is intentionally empty; the type is re-exported via `Exports.swift`.
// swiftlint:enable orphaned_doc_comment
```
Note: `superfluous_disable_command` is in `disabled_rules`, so this wrapper won't generate a secondary violation.

Option A is cleaner — the entire point of this PR is to *remove* lint suppressions, not add new ones. A tombstone file with no declaration does not need DocC-parseable `///` comments; `//` carries the same information to a human reader. Worth deciding which direction you want before landing.

---

### ⚠️ `ScopeStore.setEnabled` — `objectWillChange.send()` is a genuine double-send

This was partially documented in the diff but the reasoning stops short of what matters:

```swift
func setEnabled(_ id: UUID, _ enabled: Bool) {
    guard let idx = entries.firstIndex(where: { $0.id == id }) else { return }
    entries[idx].isEnabled = enabled  // <-- mutation here
    persist()
    log(...)
    objectWillChange.send()  // <-- explicit send below
}
```

`entries` is `@Published private(set) var entries: [ScopeEntry]`. The question of whether `@Published` fires for the in-place mutation `entries[idx].isEnabled = enabled` depends entirely on whether `ScopeEntry` is a **struct** or a **class**.

- If `ScopeEntry` is a **struct** (copy-on-write, value semantics): `entries[idx].isEnabled = enabled` replaces the entire array value → `@Published` fires → `objectWillChange.send()` fires a second time. Double-send, harmless but redundant.
- If `ScopeEntry` is a **class** (reference semantics): mutating a property on an existing element does **not** replace the array reference → `@Published` does **not** fire → the explicit `objectWillChange.send()` is load-bearing and must stay.

The PR description says the explicit send is "intentional for RunnerStore's Combine chain", and the new doc comment says the same — but neither states *why* `@Published` alone is insufficient here. If `ScopeEntry` is a struct, the comment is misleading (implies `@Published` wouldn't fire when it does). If it's a class, the comment is correct but should say so explicitly to prevent a future cleanup pass from removing it.

Suggested doc:
```swift
/// Toggles the `isEnabled` flag for the entry with the given ID.
/// Does NOT send `didMutate` — enable/disable is not a structural change.
/// Sends `objectWillChange` explicitly: `ScopeEntry` is a struct so the
/// in-place array mutation does trigger `@Published`, but the explicit send
/// is retained to guarantee RunnerStore's Combine subscription fires even
/// if the value type changes in future.
func setEnabled(_ id: UUID, _ enabled: Bool) {
```
Or, if `ScopeEntry` is a class, the comment should say: "`@Published` does not fire for in-place class property mutation — this send is load-bearing."

This is the only place in the codebase where the value/reference semantics of `ScopeEntry` have a correctness consequence. It deserves one sentence that pins it down.

---

### ✅ What's correct and well-executed

- **`willSet { objectWillChange.send() }` removal** — correct. `@Published` fires `objectWillChange` in its own `willSet`; the explicit one was a double-send with no observable effect. Clean removal.
- **`entriesKey`/`legacyKey` doc rewrites** — good. The `legacyKey` doc now implies migration-only usage, which is accurate.
- **`LoginItem` caller attribution** — `PopoverMainView` → `SettingsView` is the right fix. Clean.
- **`TerminalLauncher` doc** — replacing the autogenerated `/// Enumerates possible values for TerminalLauncher.` stub with a real description that includes escaping behaviour and the `#546` reference is exactly the right call. The new doc is accurate and complete.
- **`ScopeStore` `orphaned_doc_comment` lint pair removal** — the suppressions were defensive noise; removing them and fixing the actual cause (the `willSet` double-send) is the correct approach.
- **SonarQube** — 0 new issues, passing. 👍

---

### Net verdict

One real blocker (tombstone `///` → CI failure), one doc gap worth a single sentence (`setEnabled` value/reference semantics). Everything else in this batch is correct. Resolve the tombstone approach decision, add the one-liner to `setEnabled`'s doc, and this is good to merge.

eonist:
## Code Review — Batch 11 (Scope + Services: docs + SwiftLint cleanup)

Overall this is clean and well-scoped work. The goals from #1082 are fully met, the diff is tight, and there are no logic changes smuggled in — good discipline. A few things worth addressing before merge:

---

### 🔴 `orphaned_doc_comment` will break CI on the three tombstone files

`ScopeEntry.swift`, `LogFetcher.swift`, and `ProcessRunner.swift` are intentionally empty. The issue spec says to replace `// swiftlint:disable:next missing_docs` with a `///` tombstone — but `///` *is* a doc comment, and SwiftLint's `orphaned_doc_comment` rule fires whenever a `///` block has no declaration following it. There's no declaration in these files, so the lint will fire on all three.

The CI pipeline runs `swiftlint lint --strict`, so these become errors, not warnings. `ScopeStore.swift` originally worked around exactly this by wrapping the whole file in `// swiftlint:disable/enable orphaned_doc_comment` — which this PR correctly removed because the underlying issue (the orphaned `@Published` doc) was fixed there. But the same problem now re-appears in the three tombstone files.

**Fix options (pick one):**
- Revert to plain `//` comments for tombstone files — they aren't documentation for a Swift declaration, they're file-level notes, so `//` is semantically correct here.
- If `///` is strongly preferred for consistency, add a file-scoped `// swiftlint:disable orphaned_doc_comment` at the top of each tombstone file.

Worth noting: this PR is specifically about *removing* SwiftLint suppressions, so option 1 (plain `//`) is the more principled choice and avoids re-introducing suppressions under a different name.

---

### 🟡 `ScopeStore.setEnabled` — double `objectWillChange` fire

The `willSet { objectWillChange.send() }` on `@Published entries` was correctly removed — `@Published` fires `objectWillChange` automatically before any assignment, so the `willSet` was a duplicate. Good.

However, the explicit `objectWillChange.send()` at the bottom of `setEnabled(_:_:)` is now documented as the *only* trigger for the polling-restart behaviour. Worth double-checking the call flow: `setEnabled` mutates `entries[idx].isEnabled` via index subscript on the `@Published` array. Mutating an element *in place* on an array stored as `@Published` does trigger `objectWillChange` automatically (the whole array value changes). So the explicit `objectWillChange.send()` at the end of `setEnabled` fires a *second* time, after the mutation.

This means the Combine subscription in `RunnerStore` currently gets two rapid signals per toggle — one from `@Published`'s auto-fire, one from the explicit `.send()`. That's unlikely to cause a visible bug because Combine operators typically debounce/deduplicate, but it's technically redundant and the new doc comment ("Publishes `objectWillChange` explicitly so `RunnerStore`'s Combine subscription triggers a polling restart on toggle") is mildly misleading — it implies this is *the* publish when `@Published` already did it. Either remove the explicit `.send()` and rely on `@Published`, or keep it but acknowledge in the doc that it's a belt-and-suspenders guarantee.

---

### 🟡 `LoginItem.swift` — caller attribution precision

The fix from `PopoverMainView` → `SettingsView` is correct and welcome. One minor nit: the doc also says it's called via "the two-argument `onChange(of:)` form" — this is an implementation detail that may drift again the next time the callsite changes. Consider whether it's worth keeping. At minimum it's correct *now*, which is a clear improvement on what was there.

---

### 🟢 `TerminalLauncher.swift` — doc quality

The rewrite from the auto-generated stub to the real doc is a genuine improvement. The new doc is accurate, informative, and correctly mentions the three escape cases (backslashes, double quotes, newlines). The `// #546` cross-reference is preserved, now embedded in the doc rather than floating above it. No issues here.

---

### 🟢 `ScopeStore.swift` — `entriesKey` / `legacyKey` docs

The replacement docs ("UserDefaults key for the JSON-encoded `[ScopeEntry]` array" and "kept for migration only") are meaningfully better than the tautological originals. The migration intent is now visible in the source. Good.

---

### Summary

| File | Status | Notes |
|------|--------|-------|
| `ScopeEntry.swift` | ⚠️ Needs fix | `///` → `//` to avoid `orphaned_doc_comment` CI failure |
| `ScopeStore.swift` | ✅ with nit | Double `objectWillChange` fire; doc comment slightly misleading |
| `LogFetcher.swift` | ⚠️ Needs fix | Same `orphaned_doc_comment` issue as `ScopeEntry` |
| `LoginItem.swift` | ✅ | Caller fix is correct |
| `ProcessRunner.swift` | ⚠️ Needs fix | Same `orphaned_doc_comment` issue as `ScopeEntry` |
| `TerminalLauncher.swift` | ✅ | Solid doc rewrite |

The CI-breaking `orphaned_doc_comment` issue on the three tombstone files is the only blocker. The `setEnabled` double-publish is worth clarifying but not a merge blocker. Everything else is ready.

eonist:
## Deep Review — Batch 11 (#1083 / closes #1082)

> **Deduplication note:** The three prior human reviews and the bot reviews have already covered the tombstone `///` → CI failure and the `setEnabled` double-send. I won't rehash those beyond a line each. What follows focuses on angles those reviews left on the table or only touched at surface level.

---

### 🔴 Tombstone `///` (already flagged — one angle not yet covered)

The `orphaned_doc_comment` CI risk is well-documented above. What hasn't been said: the *real* inconsistency is that this batch treats tombstone files in the `Scope/` and `Services/` trees differently from each other mid-PR.

Before this batch, the tombstone pattern across all stub files was consistently `// plain comment`. This PR upgraded three of them to `///` while leaving others (e.g. any tombstones in the `Runner/` or `UI/` groups that weren't in scope) on `//`. The inconsistency isn't just a lint problem — it creates a documentation tier ambiguity: does `///` in a tombstone file mean something different from `//` in a tombstone file? It shouldn't. Tombstones are operational notes, not Swift API docs.

Pick one style for tombstones codebase-wide (plainly `//`) and document that choice in the contribution guide or in #1038 so future batches don't re-litigate it.

---

### 🟡 `setEnabled` — the real question the doc comment still doesn't answer (going deeper than prior reviews)

Prior reviews correctly identify the double-send and suggest adding "struct vs class" context. But there's a deeper question buried here that none of the reviews address: **why is `setEnabled` calling `objectWillChange.send()` at all, rather than `didMutate.send()`?**

Look at the comment on `didMutate`:
```swift
/// Emits after every structural mutation (add / remove). Callers subscribe and
```
…and the comment on `setEnabled`:
```swift
/// Does NOT send `didMutate` — enable/disable is not a structural change.
```

So the intent is clear: `didMutate` = structural changes (add/remove), `objectWillChange` = UI update trigger. The explicit `objectWillChange.send()` at the end of `setEnabled` is explicitly *not* going through `didMutate` so that `RunnerStore`'s Combine subscription can distinguish between "`entries` changed structurally" vs. "`entries` changed state-only".

But if `RunnerStore` is subscribing to `objectWillChange` to detect enable/disable toggles and trigger a polling restart, that subscription is *also* going to fire for every `add()`/`remove()` call (which mutate `@Published entries` and thus fire `objectWillChange` automatically). This means `RunnerStore` can't actually distinguish between "a scope was toggled" and "a scope was added/removed" using `objectWillChange` alone — it just gets fired for both.

If a polling restart is the right response to a toggle but a different response (e.g., full rebuild of the runner list) is the right response to an add/remove, the current architecture is blending those signals. The explicit `objectWillChange.send()` in `setEnabled` might be masking the need for a dedicated `didToggle` subject — similar to `didMutate` but for enable/disable state changes.

This doesn't need to be fixed in this PR (it's a logic question, not a doc one), but the new doc comment should either:
- Acknowledge that `RunnerStore` treats `objectWillChange` as a catch-all and note that add/remove also triggers it, or
- Leave a `// TODO:` pointing to the potential future refactor

---

### 🟡 `LoginItem.swift` — the `onChange(of:)` API surface detail is load-bearing

Prior reviews noted the caller attribution drift risk. I want to add: the phrase "two-argument `onChange(of:)` form" is worth keeping — but it's a Swift version constraint, not just an implementation detail.

The two-argument `onChange(of:perform:)` form (`onChange(of: value) { newValue in }`) was deprecated in macOS 14 / iOS 17. The new form is `onChange(of:) { oldValue, newValue in }`. If this callsite was written before the deprecation, the comment might actually be documenting a pattern that should be updated, not preserved.

If the project's minimum deployment target is macOS 14+, the doc comment implicitly endorses a deprecated pattern. If it's macOS 13 or earlier, the comment is accurate. Worth a quick check against the deployment target in `Package.swift` / `.xcconfig` and either updating the callsite or adjusting the comment to make the constraint explicit:
```swift
/// Called from the login-item toggle in `SettingsView` via `onChange(of:)`.
/// The new-value argument is passed directly, avoiding an extra `@State` read.
```

---

### 🟡 `TerminalLauncher.swift` — one thing the new doc omits

The doc correctly describes backslash, double-quote, and newline escaping. But there's an important caveat about `NSAppleScript` not mentioned anywhere: it **must be called on the main thread**. The `open(command:)` method has no `@MainActor` annotation and no `DispatchQueue.main.async` wrapper visible in the diff. If callers invoke it from a background queue (e.g., from a Combine subscription or a `Task`), it will silently fail or crash.

The doc comment should include:
```swift
/// - Important: Must be called on the main thread; `NSAppleScript` requires main-thread execution.
```
Or the method should be annotated `@MainActor`. This isn't just a doc quality issue — it's a latent thread-safety bug for any future caller that doesn't know to check for `NSAppleScript`'s threading constraints.

> Note: this is flagged in Batch 12's PR (#1085) as a `// TODO:` on a related caller (`FailureHookRunner`). The constraint exists at the declaration site here and should be documented here, not only at call sites.

---

### ✅ What's definitively correct and doesn't need revisiting

- `willSet { objectWillChange.send() }` removal — correct and clean.
- `entriesKey`/`legacyKey` doc rewrites — the previous docs were genuinely tautological, these are specific and actionable.
- `ScopeStore` lint suppression pair removal — the root cause is fixed, so the suppression should go. Right call.
- `TerminalLauncher` overall doc consolidation — the `#546` fold and the autogenerated-stub replacement are exactly what the codebase needs. The escaping details are accurate.

---

### Net verdict

Two things are already blocking merge (tombstone `///` lint + double-send clarity) per prior reviews. Two additional angles worth addressing before close:

1. **`TerminalLauncher` threading constraint** — add `@MainActor` or a `/// - Important:` note. This is the highest-value new find in this review; it's a latent safety issue documented nowhere.
2. **`LoginItem` `onChange` deprecation check** — verify deployment target, update comment or callsite accordingly.

The `didMutate` vs. `objectWillChange` signal-blending question is a follow-up issue candidate, not a blocker here.

eonist:
## Deep Review — Batch 11 (#1083 / closes #1082)

> **Deduplication note:** The prior reviews (three human + four bots) have thoroughly covered: tombstone `///` → `orphaned_doc_comment` CI failure, the `setEnabled` double-send, `LoginItem` caller attribution drift, `TerminalLauncher` doc consolidation, and the `didMutate` vs `objectWillChange` signal-blending question. The most recent human review also surfaced the `NSAppleScript` main-thread constraint and the `onChange(of:)` deprecation angle. I won't rehash any of those. What follows are angles not yet covered.

---

### 🔴 `ScopeStore.entries` — `private(set)` doc update introduces a misleading guarantee

The new doc on `entries` reads:

```swift
/// `private(set)` — mutate only via `add(_:)`, `remove(id:)`, and `setEnabled(_:_:)`.
@Published private(set) var entries: [ScopeEntry] = []
```

This is an improvement over the old tautological doc, but it inadvertently implies that `add`, `remove`, and `setEnabled` are the *only* mutation paths — yet `load()` (called from `init()`) also directly assigns to `entries`. If a future reader sees the doc and searches for "`entries` mutation" they'll miss `load()`. The doc should either enumerate all mutation sites or drop the exhaustive list and describe the *policy* instead:

```swift
/// All scope entries, persisted as JSON in `UserDefaults`.
/// Mutate only through the designated methods on this type; do not assign directly from outside.
```

Minor, but doc comments on `@Published private(set)` properties carry real weight because they're the first thing you read when tracing a state bug.

---

### 🟡 `ScopeStore` — `persist()` is called inside `setEnabled` but not in the doc

The updated doc for `setEnabled` explains the `objectWillChange` semantics in detail (four lines in the diff), but says nothing about the fact that the method also **persists** the change to `UserDefaults`. For a caller reading only the doc string:

```swift
/// Toggles the `isEnabled` flag for the entry with the given ID.
/// Does NOT send `didMutate` — enable/disable is not a structural change.
/// ...four lines about objectWillChange...
```

The persistence side effect is absent. This is asymmetric with the `add`/`remove` methods, which presumably document their persistence too. If the `objectWillChange` rationale warrants four lines of explanation, the persistence side effect warrants at least one sentence — both are non-obvious, non-local effects of calling `setEnabled`.

---

### 🟡 `LogFetcher.swift` / `ProcessRunner.swift` — tombstone text is slightly inconsistent with `ScopeEntry.swift`

Compare the three tombstones after this PR:

- `ScopeEntry.swift`: `// Tombstone — type moved to RunnerBarCore/Scope/ScopeEntry.swift (see #927).`  
  → `// This file is intentionally empty; **the type** is re-exported via Exports.swift.`

- `LogFetcher.swift`: `// Tombstone — type moved to RunnerBarCore/Services/LogFetcher.swift (see #926).`  
  → `// This file is intentionally empty; **the symbols** are re-exported via Exports.swift.`

- `ProcessRunner.swift`: `// Tombstone — type moved to RunnerBarCore/Services/ProcessRunner.swift (see #922).`  
  → `// This file is intentionally empty; **the type** is re-exported via Exports.swift.`

`LogFetcher` says "symbols" while the other two say "type". If `LogFetcher` exports multiple symbols (e.g., both the class and a helper enum or typealias), "symbols" is correct and the others should say "type" only if they export a single type. If `LogFetcher` also exports exactly one type, the inconsistency is noise. Either way it should be uniform — pick "type" or "symbols" for all three based on what `Exports.swift` actually re-exports.

---

### 🟡 `TerminalLauncher.swift` — `// MARK: - TerminalLauncher` removal is undocumented

The diff removes the `// MARK: - TerminalLauncher` line. In a file with a single `enum TerminalLauncher`, a MARK divider is redundant — the file name and type declaration already provide the same structural landmark. The removal is correct. But it's not mentioned in either the issue spec (#1082) or the PR description, which lists only the doc rewrite and `#546` fold. If a future auditor is tracing why MARK headers were dropped from this file, there's no breadcrumb. A one-line note in the PR description ("also removed redundant MARK — single-type file") would close this gap.

Not a blocker — just worth adding to the PR body or the commit message.

---

### ✅ What's correct and doesn't need revisiting (confirming prior reviews)

- `willSet { objectWillChange.send() }` removal is correct.
- `entriesKey`/`legacyKey` doc rewrites are the right call.
- `TerminalLauncher` doc consolidation is solid and accurate.
- `LoginItem` `PopoverMainView` → `SettingsView` fix is correct.
- SonarQube: 0 new issues. Clean.

---

### Net verdict (new items only)

| Finding | Severity | File |
|---------|----------|------|
| `entries` doc implies exhaustive mutation list but omits `load()` | 🟡 Doc gap | `ScopeStore.swift` |
| `setEnabled` doc omits `persist()` side effect | 🟡 Doc gap | `ScopeStore.swift` |
| "symbols" vs "type" inconsistency across tombstones | 🟡 Minor | `LogFetcher.swift` |
| `MARK` removal undocumented in PR/commit | ⚪ Nit | `TerminalLauncher.swift` |

The `entries` doc gap and the `setEnabled` persistence omission are the two worth fixing; they're the kind of thing that misleads a future reader doing a state-trace under time pressure.

eonist:
## Deep Review — Batch 11 (#1083 / closes #1082)

> **Deduplication note:** The four prior human reviews have thoroughly covered: tombstone `///` → CI failure, `setEnabled` double-send + struct/class semantics, `LoginItem` caller drift risk, `TerminalLauncher` doc quality, `NSAppleScript` main-thread constraint, `onChange(of:)` deprecation, and the `didMutate` vs. `objectWillChange` signal-blending question. None of what follows rehashes those threads.

---

### 🟡 `persist()` call order in `setEnabled` — silent data-loss window

The call sequence in `setEnabled` is:

```swift
entries[idx].isEnabled = enabled   // 1. mutate in-memory
persist()                           // 2. write to UserDefaults
log(...)                            // 3. log
objectWillChange.send()             // 4. notify subscribers
```

This is the correct order *most of the time*, but there is a narrow window between steps 1 and 2 where the in-memory state and the persisted state disagree. If the app is backgrounded or terminated between those two lines (e.g., via a `SIGTERM` from macOS during sleep/wake), the in-memory toggle is lost. `UserDefaults` will still hold the old value on next launch.

This is arguably acceptable for a toggle (user can re-toggle), but it is worth noting that `add(_:)` and `remove(id:)` appear to have the same pattern. The question is whether the batch that introduced `didMutate` deliberately chose "mutate then persist" over "persist then mutate" and whether that choice is documented anywhere. If not, a one-line comment on `persist()` — `// Persists immediately; in-memory and UserDefaults are momentarily out of sync above this line` — would make the ordering explicit for future readers who might consider swapping persist() and the mutation.

---

### 🟡 `legacyKey` — "kept for migration only" with no visible migration path

The new doc for `legacyKey` reads:
> `UserDefaults` key for the legacy plain `[String]` scopes array, kept for migration only.

This is a good improvement over `"The legacyKey constant."` — but it raises a question the diff doesn't answer: **is the migration path ever triggered, and is it ever cleaned up?**

If `legacyKey` data is read once on first launch to populate `entries`, then the legacy key should ideally be deleted from `UserDefaults` after migration so it doesn't sit forever in the user's plist. If it's already deleted post-migration elsewhere in `ScopeStore`, the doc should say `"legacy key — consumed during first-launch migration, then removed"` to confirm the cleanup happens. If it's *not* deleted, the data will persist in `UserDefaults` indefinitely, which is a minor storage leak and a potential source of confusion if any future code accidentally reads the wrong key.

This isn't a blocker but the new doc comment is a half-answer — it explains what the key *was* but not whether it gets cleaned up. Either add a `// TODO: remove after sufficient migration period` or update the doc to confirm disposal.

---

### 🟡 `entries` `private(set)` doc — the mutation contract statement is incomplete

The updated doc on `entries` reads:
> `private(set)` — mutate only via `add(_:)`, `remove(id:)`, and `setEnabled(_:_:)`.

This is accurate as far as it goes, but `setEnabled` is the only one of those three that does an in-place element mutation (`entries[idx].isEnabled = enabled`) rather than replacing the whole array. The doc implies the three are equivalent mutation pathways, but they have subtly different `@Published` behaviour:

- `add(_:)` and `remove(id:)` replace `entries` entirely → `@Published` fires unconditionally.
- `setEnabled(_:_:)` mutates an element in place → `@Published` fires only if `ScopeEntry` is a struct (value semantics copy-on-write propagates the mutation as an array replacement). If `ScopeEntry` were ever changed to a class, `setEnabled` would silently stop publishing without any other code changing.

The doc comment on `entries` is the right place to capture this distinction, since it's the declaration site:

```swift
/// All scope entries, persisted as JSON in `UserDefaults`.
/// `private(set)` — mutate only via `add(_:)`, `remove(id:)`, and `setEnabled(_:_:)`.
/// `add`/`remove` replace the array; `setEnabled` mutates in place — relies on
/// `ScopeEntry` being a value type for `@Published` to fire correctly.
```

This locks in the struct assumption at the property level and makes the `@MainActor`/`@Published` contract visible to anyone looking at the property rather than having to trace through `setEnabled`.

---

### 🟢 `ScopeStore` `orphaned_doc_comment` suppression removal — the right structure

One observation none of the reviews called out explicitly: the `// swiftlint:disable/enable orphaned_doc_comment` wrapper was at the **file scope** in `ScopeStore.swift`, which means it silenced the rule for the *entire* file, not just the `willSet` line. Any orphaned `///` added anywhere in that file in the future would have been silently accepted by lint. Removing the wrapper and fixing the actual cause (the redundant `willSet`) is therefore strictly better than the original — it restores the rule's enforcement for the whole file. This is the right approach and worth noting explicitly as a quality win from this batch.

---

### Net verdict

The four prior human reviews have the two actual blockers (tombstone `///` and `setEnabled` doc clarity) fully covered. This review surfaces three additional items worth tracking:

1. **`legacyKey` migration disposal** — confirm cleanup happens or add a TODO. Low friction, high documentation value.
2. **`entries` doc — struct assumption** — one sentence at the property level prevents a future silent regression.
3. **`persist()` ordering comment** — optional but makes the momentary inconsistency window explicit for future readers.

None of these are merge blockers, and this PR is already merged. Candidates for a follow-up cleanup issue under #1038.

eonist:
## Deep Review — Batch 11 (#1083 / closes #1082)

> **Deduplication notice:** I've read all existing reviews (Qodo, CodeRabbit, Greptile, CodeAnt, and the four human reviews including the comprehensive one at 20:34). The tombstone `///` → CI failure, the `setEnabled` double-send + struct/class semantics, the `TerminalLauncher` `NSAppleScript` main-thread constraint, the `LoginItem` `onChange` deprecation check, and the `didMutate` vs. `objectWillChange` signal-blending question are all well-covered. What follows is exclusively angles not addressed in any prior review.

---

### 🟡 `Exports.swift` re-export contract is invisible at the tombstone

Every tombstone file says `// This file is intentionally empty; the type is re-exported via Exports.swift.` — but the tombstone gives no breadcrumb to *which* `Exports.swift`. In a multi-target SPM project (`RunnerBar` + `RunnerBarCore` at minimum), there are likely multiple `Exports.swift` files. A future reader hitting `ScopeEntry.swift` and seeing "re-exported via Exports.swift" has to grep to find out which target's exports file actually surfaces the type.

The comment would be immediately more useful as:
```swift
// This file is intentionally empty; `ScopeEntry` is re-exported via
// RunnerBarCore/Exports.swift and is available to all RunnerBar targets.
```
This is a one-word change in scope and pays dividends the next time someone unfamiliar with the module boundary opens this file.

---

### 🟡 `private(set)` doc on `entries` is accurate but the mutation contract is incomplete

The new doc on `entries`:
```swift
/// `private(set)` — mutate only via `add(_:)`, `remove(id:)`, and `setEnabled(_:_:)`.
```
This is an improvement over the old doc, but it only describes the write path and is silent about the *read* guarantees. `entries` is `@Published private(set)` — which means external subscribers can observe it reactively via `$entries`. That's actually the primary consumption pattern for observing views (`ScopeDetailView`, etc.).

A reader looking at this property for the first time doesn't know whether to use `store.entries` (synchronous pull), `store.$entries` (Combine publisher), or `objectWillChange` (imperative). The doc should at least hint at the reactive read pattern:
```swift
/// Current scope entries, persisted as JSON in `UserDefaults`.
/// Observe via `$entries` for reactive updates; mutate only through
/// `add(_:)`, `remove(id:)`, and `setEnabled(_:_:)`.
```

---

### 🟡 `setEnabled(_:_:)` — unlabeled second parameter is a readability hazard

This is purely a call-site readability note that no prior review has raised. The signature is:
```swift
func setEnabled(_ id: UUID, _ enabled: Bool)
```
Both parameters suppress their external labels. At every call site this reads as:
```swift
scopeStore.setEnabled(entry.id, true)
```
The `true`/`false` at the call site is an unanchored `Bool` with no label telling the reader what it means. Compare:
```swift
scopeStore.setEnabled(entry.id, enabled: true)
```
This is a pre-existing wart and not introduced by this PR, but since this batch is touching `setEnabled`'s doc comment in depth, it's a natural moment to note a potential API ergonomics issue for a future batch. The issue spec (#1082) says "no logic changes" so this is a follow-up candidate, not a blocker — but worth a `// TODO:` or a dedicated micro-issue.

---

### 🟡 `legacyKey` — migration has no documented exit condition

The new doc says:
```swift
/// `UserDefaults` key for the legacy plain `[String]` scopes array, kept for migration only.
```
"Kept for migration only" implies there is a migration path and a point at which this key should be removed. But there's no `// TODO:` referencing an issue, no minimum deployment target condition, and no indication of whether the migration has already run for all users or is still running at launch.

If the migration was completed and `legacyKey` is now dead code, it should be removed. If it's still active at app launch (i.e., the `init` reads `legacyKey` first and migrates to `entriesKey`), the doc should say so. "Kept for migration only" without a deletion condition is a promise with no due date — it will still be in the codebase in three years saying the same thing.

Suggested addition:
```swift
/// `UserDefaults` key for the legacy plain `[String]` scopes array.
/// Read once during `init()` to migrate existing installs; safe to remove
/// once the minimum supported version has shipped a migration cycle (see #NNN).
```

---

### 🟡 `TerminalLauncher` doc — the newline escape claim needs verification

The new doc says:
> Backslashes, double quotes, and newlines in the command are escaped before embedding in the AppleScript string.

This was folded from the old `// #546` block comment. The claim is that newlines are escaped. In AppleScript's `do script`, embedding a literal newline in the string terminates the script prematurely — so escaping newlines is essential and correct *if* it's actually implemented.

The diff doesn't show the `open(command:)` implementation body, only the doc rewrite. If the implementation actually replaces `\n` with `\\n` (or omits newlines), the doc is accurate. If the implementation only escapes backslashes and double quotes (a common incomplete implementation), the doc is now over-promising. Since the original `// #546` block comment was the authoritative description and this PR folds it verbatim, the claim is carried forward — but it should be verified against the actual escape logic in the method body before landing, since an inaccurate doc is worse than no doc.

---

### 🟢 PR description vs. actual diff — one discrepancy worth noting

The PR description says:
> `Services/LoginItem.swift` | Fixed stale caller attribution `PopoverMainView` → `SettingsView`

But the actual diff shows the fix is `PopoverMainView` → `SettingsView` in the first line and also updates the surrounding sentence structure. The description is accurate. However, the *issue* (#1082) spec said the fix should be `PopoverMainView` → `PanelMainView`:

> `Services/LoginItem.swift` | Fix stale `PopoverMainView` caller attribution → `PanelMainView`

The PR landed on `SettingsView` instead. If `SettingsView` is the actual current caller, then this is a better fix than the issue specified and the spec was itself stale by the time the PR was written — which is fine, but worth a one-line note on #1082 explaining that the spec said `PanelMainView` but the real caller at time of implementation was `SettingsView`.

---

### Net verdict (new angles only)

| Finding | Severity | Action |
|---|---|---|
| Tombstone missing target-qualified path for `Exports.swift` | 🟡 Medium | One-word fix in the comment |
| `entries` doc silent on `$entries` reactive read pattern | 🟡 Medium | Add one sentence to the doc |
| `setEnabled(_:_:)` unlabeled `Bool` | 🟡 Low (pre-existing) | Follow-up issue / `TODO` |
| `legacyKey` has no documented exit condition | 🟡 Medium | Add a due-date clause to the doc |
| `TerminalLauncher` newline escape claim — verify vs. implementation | 🟡 Medium | Verify before merging |
| Issue spec said `PanelMainView`, PR used `SettingsView` — close the loop | 🟢 Low | One-line note on #1082 |

None of these are merge blockers on their own (the tombstone `///` lint issue and the `NSAppleScript` threading constraint flagged in earlier reviews are the real blockers). But the `legacyKey` exit condition and the `TerminalLauncher` escape claim accuracy are the two worth addressing before landing since they affect correctness of documentation that will outlive this PR.

PR #1096, opened 2026-06-02
Batch 15 · DesignSystem/ + PanelMainView+Subviews — docs, lint cleanup, AnyView removal

### LINKED ISSUE #1095: Batch 15 · DesignSystem/ + PanelMainView+Subviews.swift — docs, lint cleanup, AnyView removal
## Scope

Four files: `DesignSystem/DesignTokens.swift`, `DesignSystem/PanelViewModifiers.swift`, `DesignSystem/RemovalAlertModifier.swift`, `Views/Main/PanelMainView+Subviews.swift`.

Parent tracking issue: #1038

---

### 1. 🔴 `PanelMainView+Subviews.swift` — redundant file-level SwiftLint suppression

**Problem:** The file opens with `// swiftlint:disable colon opening_brace` and closes with `// swiftlint:enable colon opening_brace`. Both `colon` and `opening_brace` are already in the `disabled_rules` list in `.swiftlint.yml`, so these directives are no-ops and mislead future readers into thinking these rules are active.

**Fix:** Remove both lines.

---

### 2. 🔴 `RemovalAlertModifier.swift` — redundant inline SwiftLint suppression

**Problem:** `// swiftlint:disable:next multiple_closures_with_trailing_closure` on the `.alert(...)` call. `multiple_closures_with_trailing_closure` is globally disabled in `.swiftlint.yml`.

**Fix:** Remove the suppression comment.

---

### 3. 🔴 `PanelMainView+Subviews.swift` — live `print()` debug statement in hot render path

**Problem:**
```swift
print("[SubtitleDebug#1029] '\(runner.runnerName)' rawArch=\(rawArch ?? \"nil\") rawOS=\(rawOS ?? \"nil\") → subtitle=\(result ?? \"nil (hidden)\")")
```
This fires on every render of every runner row. Issue #1029 is long past; the trace was never removed.

**Fix:** Remove the `print()` call. The `runnerSubtitle` logic is stable and correct — no replacement log needed.

---

### 4. 🟡 `PanelViewModifiers.swift` — `AnyView` type erasure in `@ViewBuilder`-eligible bodies

**Problem:** `GlassCard.body`, `StatPillBackground.body`, `StatusBadgeBackground.body`, and `BranchTagPillBackground.body` all wrap branches in `AnyView(...)`. `AnyView` disables SwiftUI's structural diffing, breaks view identity for animations, and wastes allocation. Since `body` already returns `some View`, the `if #available` branch can be handled with `@ViewBuilder` directly.

**Fix:** Add `@ViewBuilder` to each `body` function and remove all `AnyView(...)` wrappers.

```swift
// Before
func body(content: Content) -> some View {
    if #available(macOS 26, *) {
        AnyView(content.glassEffect(...))
    } else {
        AnyView(materialFallback(content: content))
    }
}

// After
@ViewBuilder
func body(content: Content) -> some View {
    if #available(macOS 26, *) {
        content.glassEffect(...)
    } else {
        materialFallback(content: content)
    }
}
```

---

### 5. 🟡 `DesignTokens.swift` — `RBStatus.tint` uses `default:` instead of exhaustive switch

**Problem:**
```swift
public var tint: Color {
    switch self {
    case .inProgress: return .rbBlueTint
    case .success:    return .rbGreenTint
    case .failed:     return .rbRedTint
    case .queued:     return .rbYellowTint
    default:          return .clear        // ← silently absorbs future cases
    }
}
```
If a new `RBStatus` case is added, the compiler will not warn and `.unknown` / new cases silently return `.clear`. The same exhaustiveness fix was applied to `dotColor` in Batch 14.

**Fix:** Replace `default:` with `case .unknown:`.

---

### 6. 🟡 `DesignTokens.swift` — tint tokens marked `public` in an app target

**Problem:** `rbYellowTint`, `rbBlueTint`, `rbGreenTint`, `rbRedTint`, `rbOrangeTint` are declared `public static let`. This is an app target, not a library — `public` has no effect and implies a cross-module API that doesn't exist.

**Fix:** Remove `public` (leave as `internal static let`, which is the default).

---

### 7. 🟡 `DesignTokens.swift` — `rbOrangeTint` is a duplicate, not an alias

**Problem:** The doc comment says `rbOrangeTint` is "alias for `rbYellowTint`" but both are independently computed as `rbWarning.opacity(0.08)` — they're duplicates.

**Fix:** Update the doc comment to: `/// Duplicate of rbYellowTint — kept for call-site compatibility. Prefer rbYellowTint in new code.`

---

### 8. 🟡 `PanelViewModifiers.swift` — `periphery:ignore` on `BranchTagPill` undocumented

**Problem:** `struct BranchTagPill: View { // periphery:ignore` with no explanation. Future readers don't know why Periphery flags this or if the ignore is still needed.

**Fix:** Add reason inline:
```swift
struct BranchTagPill: View { // periphery:ignore — used dynamically inside ActionRowView.rowContent; Periphery cannot trace the call site
```

---

### 9. 🟢 Tautological `/// The body property.` doc comments

**Problem:** `SectionHeaderLabel`, `RunnerTypeIcon`, `RunnerMetricsBadge`, `PanelLocalRunnerRow`, `ActionRowView` all have `/// The body property.` on their `body` var. This is noise — the property name already says it's the body.

**Fix:** Replace each with a one-line description of what it renders:

| Type | Replacement |
|---|---|
| `SectionHeaderLabel` | `/// Renders the uppercased title in secondary caption color with standard panel insets.` |
| `RunnerTypeIcon` | `/// Renders a desktop or cloud SF Symbol in secondary color at 9 pt.` |
| `RunnerMetricsBadge` | `/// Renders CPU and MEM labels in a shared glass badge.` |
| `PanelLocalRunnerRow` | `/// Renders the runner card list, or nothing if runners is empty.` |
| `ActionRowView` | `/// Routes to glassMorphBody on macOS 26+ or legacyBody on earlier OS versions.` |

---

### 10. 🟢 `RemovalAlertModifier.swift` — doc comment describes implementation detail, not API contract

**Problem:** `/// Extracted from SettingsView.body to satisfy function_body_length and file_length limits.` This is scaffolding rationale, not useful to call sites.

**Fix:**
```swift
/// Confirmation alert for runner removal.
/// Shows authentication-aware message text: authenticated users see the svc.sh/config.sh
/// command summary; unauthenticated users see the gh-auth sign-in instructions.
/// `onConfirm` is called on destructive confirmation; `onCancel` on dismissal.
```

---

### 11. 🟢 `PanelMainView+Subviews.swift` — elapsed visibility condition undocumented

**Problem:** The condition `group.groupStatus == .inProgress || group.groupStatus == .queued` in `metaTrailing` hides the elapsed timer for completed rows, but this is not explained inline.

**Fix:** Add a single-line comment:
```swift
// Elapsed shown only while running or queued — snaps off on completion to avoid a frozen clock.
```

---

## Files Changed

- `Sources/RunnerBar/DesignSystem/DesignTokens.swift`
- `Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift`
- `Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift`
- `Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift`

## CI Impact

- **SwiftLint:** all removals are no-ops (rules already disabled globally) or correctness improvements. No new violations introduced.
- **Periphery:** `periphery:ignore` on `BranchTagPill` is retained with added comment — no change to what Periphery sees.
- **Swift build/tests:** no logic changes. `AnyView` removal is semantics-preserving (same conditional rendering, better identity). `default:` → `case .unknown:` is exhaustiveness — Swift will error at compile time if a case is missed, not at runtime.

eonist:
## **User description**
Closes #1095
Part of #1038

### `DesignTokens.swift`
- `RBStatus.tint`: `default:` → `case .unknown:` — exhaustive switch, matches Batch 14 `dotColor` fix; Swift will catch any missed future case at compile time
- Remove `public` from 5 tint tokens (`rbYellowTint`, `rbBlueTint`, `rbGreenTint`, `rbRedTint`, `rbOrangeTint`) — app target, `public` has no effect here
- `rbOrangeTint` doc clarified: duplicate of `rbYellowTint`, not an alias
- `RBStatus.tint` and `RBStatus.sfSymbol` `public` removed (consistency with `color`)

### `PanelViewModifiers.swift`
- Remove `AnyView(...)` from `GlassCard`, `StatPillBackground`, `StatusBadgeBackground`, `BranchTagPillBackground` — add `@ViewBuilder` to `body`; restores SwiftUI view identity and structural diffing
- `BranchTagPill` `periphery:ignore` comment extended with reason
- Replace tautological `/// The body property.` docs on all view types

### `RemovalAlertModifier.swift`
- Remove `// swiftlint:disable:next multiple_closures_with_trailing_closure` — `multiple_closures_with_trailing_closure` is globally disabled in `.swiftlint.yml`
- Replace implementation-detail type doc with API contract doc

### `PanelMainView+Subviews.swift`
- Remove file-level `// swiftlint:disable colon opening_brace` / `// swiftlint:enable` — both rules are globally disabled in `.swiftlint.yml`
- Remove `print("[SubtitleDebug#1029]...")` debug trace from `runnerSubtitle` hot render path
- Replace tautological `/// The body property.` docs across all view types
- Add `// Elapsed shown only while running or queued` clarifying comment in `metaTrailing`

## CI
- No logic changes anywhere
- `AnyView` removal is semantics-preserving — same conditional rendering, better identity
- `case .unknown:` is a compile-time exhaustiveness check
- All removed SwiftLint directives were already no-ops per `.swiftlint.yml` (`disabled_rules`)
- `periphery:ignore` retained on `BranchTagPill` — no change to what Periphery sees

___

## **CodeAnt-AI Description**
Remove a stray debug print from the runner list and clean up several design system view descriptions

### What Changed
- Removed an accidental debug message from runner subtitles, so the panel no longer writes trace output while rendering
- Kept the same visual behavior for glass cards, pills, and status badges while simplifying how those views are shown across supported macOS versions
- Updated outdated inline descriptions and removed redundant lint notes so the panel and design system code are easier to maintain
- Clarified the status tint names and the runner-removal alert text for more consistent call-site use

### Impact
`✅ Cleaner panel logs`
`✅ Less noisy runner list rendering`
`✅ Clearer design system usage`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **Chores**
  * Cleaned up internal design system architecture and improved SwiftUI view modifier implementations.
  * Removed deprecated design color token.
  * Improved code documentation and removed debug output.

### DIFF
--- a/Sources/RunnerBar/DesignSystem/DesignTokens.swift
+++ b/Sources/RunnerBar/DesignSystem/DesignTokens.swift
@@ -147,15 +147,15 @@ extension Color {
     )
 
     /// Low-opacity amber tint for row backgrounds in warning/queued state.
-    public static let rbYellowTint = rbWarning.opacity(0.08)
+    static let rbYellowTint = rbWarning.opacity(0.08)
     /// Low-opacity blue tint for row backgrounds in in-progress state.
-    public static let rbBlueTint = rbBlue.opacity(0.08)
+    static let rbBlueTint = rbBlue.opacity(0.08)
     /// Low-opacity green tint for row backgrounds in success state.
-    public static let rbGreenTint = rbSuccess.opacity(0.08)
+    static let rbGreenTint = rbSuccess.opacity(0.08)
     /// Low-opacity red tint for row backgrounds in failed/danger state.
-    public static let rbRedTint = rbDanger.opacity(0.08)
-    /// Low-opacity orange tint — alias for `rbYellowTint`.
-    public static let rbOrangeTint = rbWarning.opacity(0.08)
+    static let rbRedTint = rbDanger.opacity(0.08)
+    /// Duplicate of `rbYellowTint` — kept for call-site compatibility. Prefer `rbYellowTint` in new code.
+    static let rbOrangeTint = rbWarning.opacity(0.08)
 }
 
 // MARK: - Status helpers
@@ -185,18 +185,18 @@ enum RBStatus {
     }
 
     /// A low-opacity background tint to visually distinguish rows by status.
-    public var tint: Color {
+    var tint: Color {
         switch self {
         case .inProgress: return .rbBlueTint
-        case .success: return .rbGreenTint
-        case .failed: return .rbRedTint
-        case .queued: return .rbYellowTint
-        default: return .clear
+        case .success:    return .rbGreenTint
+        case .failed:     return .rbRedTint
+        case .queued:     return .rbYellowTint
+        case .unknown:    return .clear
         }
     }
 
     /// The SF Symbol name that represents this status.
-    public var sfSymbol: String {
+    var sfSymbol: String {
         switch self {
         case .inProgress: return "arrow.trianglehead.2.clockwise"
         case .success: return "checkmark"
--- a/Sources/RunnerBar/DesignSystem/DesignTokens.swift
+++ b/Sources/RunnerBar/DesignSystem/DesignTokens.swift
@@ -154,8 +154,6 @@ extension Color {
     static let rbGreenTint = rbSuccess.opacity(0.08)
     /// Low-opacity red tint for row backgrounds in failed/danger state.
     static let rbRedTint = rbDanger.opacity(0.08)
-    /// Duplicate of `rbYellowTint` — kept for call-site compatibility. Prefer `rbYellowTint` in new code.
-    static let rbOrangeTint = rbWarning.opacity(0.08)
 }
 
 // MARK: - Status helpers
--- a/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
+++ b/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
@@ -32,23 +32,22 @@ struct GlassCard: ViewModifier {
     }
 
     /// Applies the glass card effect to the given content view.
+    @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
             // glassEffect does not render a stroke automatically — add it as an
             // overlay so the card has the same subtle border in both OS paths.
-            AnyView(
-                content
-                    .glassEffect(
-                        .regular,
-                        in: RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
-                    )
-                    .overlay(
-                        RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
-                            .strokeBorder(.white.opacity(strokeOpacity), lineWidth: 0.5)
-                    )
-            )
+            content
+                .glassEffect(
+                    .regular,
+                    in: RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
+                )
+                .overlay(
+                    RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
+                        .strokeBorder(.white.opacity(strokeOpacity), lineWidth: 0.5)
+                )
         } else {
-            AnyView(materialFallback(content: content))
+            materialFallback(content: content)
         }
     }
 
@@ -105,6 +104,7 @@ struct GlassButton: ViewModifier {
     }
 
     /// Applies the interactive glass button effect to the given content view.
+    @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
             content
@@ -127,15 +127,12 @@ struct GlassButton: ViewModifier {
 /// not a card container.
 struct StatPillBackground: ViewModifier {
     /// Applies a glass capsule background (macOS 26+) or ultra-thin material capsule (pre-26).
+    @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
-            AnyView(
-                content.glassEffect(.regular, in: Capsule())
-            )
+            content.glassEffect(.regular, in: Capsule())
         } else {
-            AnyView(
-                content.background(.ultraThinMaterial, in: Capsule())
-            )
+            content.background(.ultraThinMaterial, in: Capsule())
         }
     }
 }
@@ -149,18 +146,15 @@ struct StatusBadgeBackground: ViewModifier {
     let color: Color
 
     /// Applies a tinted glass capsule background (macOS 26+) or stroke capsule border (pre-26).
+    @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
-            AnyView(
-                content
-                    .background(color.opacity(0.25), in: Capsule())
-                    .overlay(Capsule().strokeBorder(color.opacity(0.55), lineWidth: 0.5))
-            )
+            content
+                .background(color.opacity(0.25), in: Capsule())
+                .overlay(Capsule().strokeBorder(color.opacity(0.55), lineWidth: 0.5))
         } else {
-            AnyView(
-                content.background(
-                    Capsule().strokeBorder(color.opacity(0.5), lineWidth: 1)
-                )
+            content.background(
+                Capsule().strokeBorder(color.opacity(0.5), lineWidth: 1)
             )
         }
     }
@@ -172,18 +166,15 @@ struct StatusBadgeBackground: ViewModifier {
 /// macOS < 26: `Capsule().strokeBorder(rbAccent.opacity(0.4), lineWidth: 1)` (unchanged).
 struct BranchTagPillBackground: ViewModifier {
     /// Applies a tinted glass capsule background (macOS 26+) or stroke capsule border (pre-26).
+    @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
-            AnyView(
-                content
-                    .background(Color.rbAccent.opacity(0.15), in: Capsule())
-                    .glassEffect(.regular, in: Capsule())
-            )
+            content
+                .background(Color.rbAccent.opacity(0.15), in: Capsule())
+                .glassEffect(.regular, in: Capsule())
         } else {
-            AnyView(
-                content.background(
-                    Capsule().strokeBorder(Color.rbAccent.opacity(0.4), lineWidth: 1)
-                )
+            content.background(
+                Capsule().strokeBorder(Color.rbAccent.opacity(0.4), lineWidth: 1)
             )
         }
     }
@@ -310,7 +301,7 @@ struct StatusBadge: View {
 // MARK: - BranchTagPill
 /// Inline pill displaying a git branch or tag name.
 /// Uses an accent-tinted glass capsule on macOS 26+, stroke capsule pre-26.
-struct BranchTagPill: View { // periphery:ignore
+struct BranchTagPill: View { // periphery:ignore — used dynamically inside ActionRowView.rowContent; Periphery cannot trace the call site
     /// The branch or tag name displayed inside the pill.
     let name: String
 
--- a/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
+++ b/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
@@ -165,6 +165,9 @@ struct StatusBadgeBackground: ViewModifier {
 /// Background modifier for `BranchTagPill` capsule pills.
 /// macOS 26+: accent tint layer + `.glassEffect(.regular, in: Capsule())`.
 /// macOS < 26: `Capsule().strokeBorder(rbAccent.opacity(0.4), lineWidth: 1)` (unchanged).
+/// Note: the tint is applied *before* the glass effect so the accent colour tints
+/// the frosted glass layer. This is distinct from `StatPillBackground` (glass only)
+/// and `StatusBadgeBackground` (tint + stroke, no glass).
 struct BranchTagPillBackground: ViewModifier {
     /// Applies a tinted glass capsule background (macOS 26+) or stroke capsule border (pre-26).
     @ViewBuilder
--- a/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
+++ b/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
@@ -104,7 +104,6 @@ struct GlassButton: ViewModifier {
     }
 
     /// Applies the interactive glass button effect to the given content view.
-    @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
             content
@@ -139,13 +138,15 @@ struct StatPillBackground: ViewModifier {
 
 // MARK: - StatusBadgeBackground
 /// Background modifier for `StatusBadge` capsule badges.
-/// macOS 26+: colour tint layer + `.glassEffect(.regular, in: Capsule())`.
+/// macOS 26+: colour tint layer + stroke overlay on a `Capsule()` shape.
 /// macOS < 26: `Capsule().strokeBorder(color.opacity(0.5), lineWidth: 1)` (unchanged).
+/// Note: intentionally does NOT use `.glassEffect` — the tinted capsule design
+/// is distinct from the glass card pattern used in `GlassCard` and `StatPillBackground`.
 struct StatusBadgeBackground: ViewModifier {
     /// The status color used to tint the badge.
     let color: Color
 
-    /// Applies a tinted glass capsule background (macOS 26+) or stroke capsule border (pre-26).
+    /// Applies a tinted capsule background (macOS 26+) or stroke capsule border (pre-26).
     @ViewBuilder
     func body(content: Content) -> some View {
         if #available(macOS 26, *) {
--- a/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
+++ b/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
@@ -4,9 +4,10 @@ import SwiftUI
 
 // MARK: - RemovalAlertModifier
 
-/// `ViewModifier` that encapsulates the runner-removal confirmation alert.
-/// Extracted from `SettingsView.body` to satisfy `function_body_length`
-/// and `file_length` limits.
+/// Confirmation alert for runner removal.
+/// Shows authentication-aware message text: authenticated users see the `svc.sh`/`config.sh`
+/// command summary; unauthenticated users see the `gh auth login` sign-in instructions.
+/// `onConfirm` is called on destructive confirmation; `onCancel` on dismissal.
 struct RemovalAlertModifier: ViewModifier {
     /// The alert title string.
     let title: String
@@ -21,7 +22,6 @@ struct RemovalAlertModifier: ViewModifier {
 
     /// Wraps `content` with a runner-removal confirmation alert.
     func body(content: Content) -> some View {
-        // swiftlint:disable:next multiple_closures_with_trailing_closure
         content.alert(title, isPresented: $isPresented) {
             Button("Cancel", role: .cancel) { onCancel() }
             Button("Remove", role: .destructive) { onConfirm() }
--- a/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
+++ b/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
@@ -5,9 +5,10 @@ import SwiftUI
 // MARK: - RemovalAlertModifier
 
 /// Confirmation alert for runner removal.
-/// Shows authentication-aware message text: authenticated users see the `svc.sh`/`config.sh`
-/// command summary; unauthenticated users see the `gh auth login` sign-in instructions.
+/// Presents a destructive action sheet with Cancel and Remove buttons.
 /// `onConfirm` is called on destructive confirmation; `onCancel` on dismissal.
+/// The `isAuthenticated` flag selects between two pre-composed message strings
+/// supplied by the call site — this modifier owns no auth logic itself.
 struct RemovalAlertModifier: ViewModifier {
     /// The alert title string.
     let title: String
--- a/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
+++ b/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
@@ -8,13 +8,13 @@ import SwiftUI
 /// Presents a destructive action sheet with Cancel and Remove buttons.
 /// `onConfirm` is called on destructive confirmation; `onCancel` on dismissal.
 /// The `isAuthenticated` flag selects between two pre-composed message strings
-/// supplied by the call site — this modifier owns no auth logic itself.
+/// defined inside this modifier — only the flag itself is caller-supplied.
 struct RemovalAlertModifier: ViewModifier {
     /// The alert title string.
     let title: String
     /// Controls whether the alert is presented.
     @Binding var isPresented: Bool
-    /// Whether a GitHub token is available; changes the alert message.
+    /// Whether a GitHub token is available; selects the pre-composed message text.
     let isAuthenticated: Bool
     /// Called when the user taps Cancel.
     let onCancel: () -> Void
--- a/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
+++ b/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
@@ -1,6 +1,5 @@
 // PanelMainView+Subviews.swift
 // RunnerBar
-// swiftlint:disable colon opening_brace
 
 import RunnerBarCore
 import SwiftUI
@@ -10,7 +9,7 @@ import SwiftUI
 struct SectionHeaderLabel: View {
     /// The title text displayed in uppercase.
     let title: String
-    /// The body property.
+    /// Renders the uppercased title in secondary caption color with standard panel insets.
     var body: some View {
         Text(title.uppercased())
             .font(RBFont.sectionCaption)
@@ -33,7 +32,7 @@ struct PanelHeaderView: View {
     let onSelectSettings: () -> Void
     /// Called when the user taps the sign-in button.
     let onSignIn: () -> Void
-    /// The body property.
+    /// Renders the header bar: system stats, optional sign-in button, settings and quit buttons.
     var body: some View {
         HStack(spacing: 6) {
             HeaderStatsBar(statsVM: statsVM)
@@ -101,7 +100,7 @@ struct PanelHeaderView: View {
 private struct RunnerTypeIcon: View {
     /// Whether the runner is self-hosted (local) or GitHub-hosted (cloud).
     let isLocal: Bool
-    /// The body property.
+    /// Renders a desktop or cloud SF Symbol in secondary color at 9 pt.
     var body: some View {
         Image(systemName: isLocal ? "desktopcomputer" : "cloud")
             .font(.system(size: 9))
@@ -118,7 +117,7 @@ private struct RunnerMetricsBadge: View {
     let cpu: Double
     /// Memory usage percentage (0–100).
     let mem: Double
-    /// The body property.
+    /// Renders CPU and MEM labels in a shared glass badge.
     var body: some View {
         HStack(spacing: 8) {
             metricItem(label: "CPU", value: String(format: "%.0f%%", cpu))
@@ -153,7 +152,7 @@ private struct RunnerMetricsBadge: View {
 struct PanelLocalRunnerRow: View {
     /// The list of runners to display.
     let runners: [RunnerModel]
-    /// The body property.
+    /// Renders the runner card list, or nothing if `runners` is empty.
     var body: some View {
         if !runners.isEmpty { runnerList(runners) }
     }
@@ -202,9 +201,7 @@ struct PanelLocalRunnerRow: View {
         let arch = rawArch.map { normaliseArch($0) }
         let os = rawOS.map { normalisePlatform($0) }
         let parts = [arch, os].compactMap { $0 }
-        let result = parts.isEmpty ? nil : parts.joined(separator: " · ")
-        print("[SubtitleDebug#1029] '\(runner.runnerName)' rawArch=\(rawArch ?? "nil") rawOS=\(rawOS ?? "nil") → subtitle=\(result ?? "nil (hidden)")")
-        return result
+        return parts.isEmpty ? nil : parts.joined(separator: " · ")
     }
     /// Normalises raw architecture strings. Maps "ARM64"→"arm64", "X64"→"x64", "X86"→"x86".
     private func normaliseArch(_ raw: String) -> String {
@@ -246,7 +243,7 @@ struct ActionRowView: View {
     @State private var expandState: Bool?
     /// The row status observed on the previous tick, used to detect transitions.
     @State private var previousStatus: RBStatus?
-    /// The body property.
+    /// Routes to `glassMorphBody` on macOS 26+ or `legacyBody` on earlier OS versions.
     var body: some View {
         if #available(macOS 26, *) {
             glassMorphBody
@@ -442,6 +439,7 @@ struct ActionRowView: View {
                 .lineLimit(1)
                 .fixedSize(horizontal: true, vertical: false)
         }
+        // Elapsed shown only while running or queued — snaps off on completion to avoid a frozen clock.
         if group.groupStatus == .inProgress || group.groupStatus == .queued {
             Text(group.elapsed)
                 .font(DesignTokens.Fonts.mono)
@@ -498,4 +496,3 @@ private struct RowTapModifier: ViewModifier {
         }
     }
 }
-// swiftlint:enable colon opening_brace
--- a/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
+++ b/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
@@ -152,7 +152,7 @@ private struct RunnerMetricsBadge: View {
 struct PanelLocalRunnerRow: View {
     /// The list of runners to display.
     let runners: [RunnerModel]
-    /// Renders the runner card list; emits nothing (no EmptyView placeholder) when `runners` is empty.
+    /// Renders the runner card list, or nothing if `runners` is empty.
     var body: some View {
         if !runners.isEmpty { runnerList(runners) }
     }
@@ -420,9 +420,6 @@ struct ActionRowView: View {
     }
 
     /// Trailing meta: time-ago · steps/total · elapsed(mm:ss, active only) · statusBadge.
-    ///
-    /// elapsed is only shown while the run is inProgress or queued — completed rows
-    /// show only time-ago to avoid a frozen clock.
     @ViewBuilder private func metaTrailing(tick tickSnapshot: Int) -> some View {
         if let start = group.firstJobStartedAt {
             Text(RelativeTimeFormatter.string(from: start))
@@ -439,7 +436,7 @@ struct ActionRowView: View {
                 .lineLimit(1)
                 .fixedSize(horizontal: true, vertical: false)
         }
-        // Hidden when not running or queued — prevents a frozen elapsed display after completion.
+        // Elapsed shown only while running or queued — snaps off on completion to avoid a frozen clock.
         if group.groupStatus == .inProgress || group.groupStatus == .queued {
             Text(group.elapsed)
                 .font(DesignTokens.Fonts.mono)
--- a/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
+++ b/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
@@ -32,7 +32,7 @@ struct PanelHeaderView: View {
     let onSelectSettings: () -> Void
     /// Called when the user taps the sign-in button.
     let onSignIn: () -> Void
-    /// Renders the header bar: system stats, optional sign-in button, settings and quit buttons.
+    /// Renders the header bar: system stats, sign-in button (shown when unauthenticated), settings and quit buttons.
     var body: some View {
         HStack(spacing: 6) {
             HeaderStatsBar(statsVM: statsVM)
@@ -152,7 +152,7 @@ private struct RunnerMetricsBadge: View {
 struct PanelLocalRunnerRow: View {
     /// The list of runners to display.
     let runners: [RunnerModel]
-    /// Renders the runner card list, or nothing if `runners` is empty.
+    /// Renders the runner card list; emits nothing (no EmptyView placeholder) when `runners` is empty.
     var body: some View {
         if !runners.isEmpty { runnerList(runners) }
     }
@@ -439,7 +439,7 @@ struct ActionRowView: View {
                 .lineLimit(1)
                 .fixedSize(horizontal: true, vertical: false)
         }
-        // Elapsed shown only while running or queued — snaps off on completion to avoid a frozen clock.
+        // Hidden when not running or queued — prevents a frozen elapsed display after completion.
         if group.groupStatus == .inProgress || group.groupStatus == .queued {
             Text(group.elapsed)
                 .font(DesignTokens.Fonts.mono)


eonist:
## Batch 15 Review

All 11 items from issue #1095 are implemented correctly and match the spec 1-for-1. The two existing bot comments are noted below — I haven't re-raised either.

---

### ✅ What's solid

**`AnyView` removal (`PanelViewModifiers.swift`)** — The `@ViewBuilder` additions on `GlassCard`, `StatPillBackground`, `StatusBadgeBackground`, and `BranchTagPillBackground` are correct and complete. SwiftUI can now structurally diff both branches of every `if #available` block, which matters for any animation or identity-sensitive work inside those modifiers. Five modifiers cleaned up in one pass is a good return.

**Exhaustive switch on `RBStatus.tint`** — `default:` → `case .unknown:` is exactly right and consistent with the Batch 14 `dotColor` fix. This is the defensive pattern worth maintaining across the whole enum.

**`public` removal on tint tokens and `RBStatus` computed vars** — Correct. App targets get nothing from `public`; having it implies a cross-module surface that doesn't exist, which misleads both readers and tools like Periphery.

**`print("[SubtitleDebug#1029]…")` removal** — This fires on every render of every runner row (it's inside `runnerSubtitle`, called from `rowContent`). Removing it is a real runtime win, not just hygiene.

**`periphery:ignore` comment extended with reason** — `// periphery:ignore — used dynamically inside ActionRowView.rowContent; Periphery cannot trace the call site` is exactly the right kind of suppression annotation: future readers know both that the ignore is intentional and *why* static analysis can't see the call site.

**Elapsed comment in `metaTrailing`** — Small but genuinely useful. The condition `group.groupStatus == .inProgress || group.groupStatus == .queued` is non-obvious without the comment — a future reader might try to "fix" the always-hidden clock on completed rows.

---

### 🔴 One real gap not yet raised

**`sfSymbol` switch still uses `default:` (DesignTokens.swift, ~line 200)**

The Octopus bot flagged this inline. Worth reinforcing with context: `tint` and `sfSymbol` are sibling computed properties on `RBStatus`. This PR correctly made `tint` exhaustive. `sfSymbol` was left with `default: return ""`. The two properties now have inconsistent safety guarantees — a new `RBStatus` case gets a compile-time error for `tint` but silently falls through to an empty string for `sfSymbol`, which will produce a missing icon with no diagnostic. Since the fix is a one-line substitution (`default:` → `case .unknown:`) and the issue even notes exhaustiveness should match Batch 14's `dotColor` pattern, this should be part of this batch.

```swift
// current
default: return ""

// should be
case .unknown: return ""
```

---

### 🟡 Observation: `rbOrangeTint` deprecation path

The doc update (`"Duplicate of rbYellowTint — kept for call-site compatibility. Prefer rbYellowTint in new code."`) is correct per the issue spec. One follow-up worth tracking: with no `@available(*, deprecated, renamed: "rbYellowTint")` attribute, the "prefer rbYellowTint" guidance is doc-only and won't surface at call sites in Xcode. This isn't a blocker for this batch, but if the intent is to actively migrate call sites away from `rbOrangeTint`, a `@available` deprecation on a future batch would make the doc comment enforceable. Worth a tracking note on #1038.

---

### 🟢 Minor doc nits (not blocking)

- `RemovalAlertModifier` new type doc mentions `svc.sh`/`config.sh` and `gh auth login` — both accurate per the alert's message logic. The doc is a genuine improvement over the old scaffolding comment.
- `PanelHeaderView.body` doc: `"Renders the header bar: system stats, optional sign-in button, settings and quit buttons."` — accurate and matches the actual `HStack` composition.
- All five `"/// The body property."` replacements match the issue spec table exactly.

---

### CodeAnt `StatusBadgeBackground` regression claim

CodeAnt flagged that the macOS 26+ branch of `StatusBadgeBackground` is missing `.glassEffect`. Looking at the diff carefully: the *pre-existing* `StatusBadgeBackground` code never had `.glassEffect` in its 26+ branch — it uses `background(color.opacity(0.25), in: Capsule())` + stroke overlay, which is the intentional design (tinted capsule, not glass). The sibling modifiers (`BranchTagPillBackground`, `StatPillBackground`) do use `.glassEffect` but those are different visual components. The CodeAnt claim appears to be a false positive based on misreading the cross-modifier design intent. The docs at line 142–143 describe the tint-layer pattern, not a glass effect, so no regression here.

---

**Summary:** One actionable item — make `sfSymbol` exhaustive to match `tint`. Everything else in the batch is well-executed and the issue spec was followed precisely.

eonist:
## Deep Review — Batch 15

> Pre-check: I've read the full diff, issue #1095, and all existing review threads. I'm skipping the two items already raised: (1) `sfSymbol` exhaustiveness (`default:` → `case .unknown`) and (2) `StatusBadgeBackground` missing `.glassEffect` on macOS 26+. Everything below is new signal.

---

### 🔴 `GlassButton.body` still lacks `@ViewBuilder` — inconsistent with the issue fix

Issue #1095 item 4 explicitly lists `GlassCard`, `StatPillBackground`, `StatusBadgeBackground`, and `BranchTagPillBackground` as the four targets for `AnyView` removal + `@ViewBuilder`. The diff correctly adds `@ViewBuilder` to all four. However, `GlassButton.body` also contains a bare `if #available(macOS 26, *)` branch and was also wrapped — the diff shows `@ViewBuilder` is added to it too (line +107 in the diff). ✅ That's fine.

Wait — looking more carefully: `GlassButton` does NOT have `AnyView` wrappers in the pre-image (the branches directly return `content.xxx`), so no `AnyView` existed there to remove, and `@ViewBuilder` was added anyway. That's actually **unnecessary** — `@ViewBuilder` on a `some View`-returning function that only uses `if #available` with no divergent branch types is a no-op at the type level. It won't cause a bug, but it's cargo-cult application of the fix pattern to a case where the problem didn't exist. Worth noting for consistency of rationale in the PR description ("AnyView removal" is the stated driver, but `GlassButton` had none).

---

### 🟡 `rbOrangeTint` doc says "kept for call-site compatibility" — but is it actually used anywhere?

The new doc comment reads:
```swift
/// Duplicate of `rbYellowTint` — kept for call-site compatibility. Prefer `rbYellowTint` in new code.
```

"Kept for call-site compatibility" implies active callers. If `rbOrangeTint` has zero call sites in the app target (easily verifiable: `grep -r "rbOrangeTint" Sources/`), the right cleanup is removal, not deprecation-style documentation. Leaving it with this comment will indefinitely preserve dead code under the justification of a phantom compatibility concern. Either:
- Add a `@available(*, deprecated, renamed: "rbYellowTint")` attribute so callers get a compiler nudge, or
- Remove it outright if it's unused.

The issue (#1095 item 7) says only to fix the doc comment — so this is within scope of the batch's spirit of not papering over issues.

---

### 🟡 `PanelLocalRunnerRow.body` new doc: "or nothing if `runners` is empty" — misleading contract

The replacement doc:
```swift
/// Renders the runner card list, or nothing if `runners` is empty.
```

Looking at the implementation:
```swift
var body: some View {
    if !runners.isEmpty { runnerList(runners) }
}
```

SwiftUI's `@ViewBuilder` implicit result for a one-armed `if` with no `else` is `Optional<RunnerListView>` — it doesn't render `EmptyView`, it renders nothing, which can affect layout depending on parent containers. The doc is accurate in plain English but elides the SwiftUI-specific consequence: a parent `VStack` or `LazyVStack` will get an empty slot with zero size. Not a bug per se, but "or nothing" undersells the structural impact. A more precise doc: `/// Renders the runner card list; emits nothing (no EmptyView placeholder) when \`runners\` is empty.`

---

### 🟡 `metaTrailing` elapsed comment placement is slightly off

Added comment:
```swift
// Elapsed shown only while running or queued — snaps off on completion to avoid a frozen clock.
```

This is good intent (issue #1095 item 11). However the comment sits *above* the `if` condition, which is standard. The issue is that "snaps off on completion" implies an animation/transition behaviour, but the view just disappears — there's no `withAnimation` or `.transition` here. Someone reading this will look for a snap transition and find none. Suggest: "Hidden when not running or queued — prevents a frozen elapsed display after completion." Avoids the false implication of an animated snap.

---

### 🟢 `PanelHeaderView.body` doc says "optional sign-in button" — potentially misleading

New doc:
```swift
/// Renders the header bar: system stats, optional sign-in button, settings and quit buttons.
```

"Optional sign-in button" is accurate in that it conditionally renders, but it implies the button itself is `Optional<Button>` rather than a conditional branch. Consider: "sign-in button (shown when unauthenticated)" to be explicit about the condition, consistent with how `RemovalAlertModifier`'s new doc explicitly states the auth-aware branching.

---

### 🟢 Alignment-only whitespace change in `tint` switch not in scope

```swift
-        case .success: return .rbGreenTint
-        case .failed: return .rbRedTint
-        case .queued: return .rbYellowTint
+        case .success:    return .rbGreenTint
+        case .failed:     return .rbRedTint
+        case .queued:     return .rbYellowTint
+        case .unknown:    return .clear
```

The alignment padding added to match `.inProgress` is a cosmetic-only change touching lines not mentioned in the issue. This is fine, but it inflates the diff and can cause merge conflicts if another batch modifies adjacent cases. SwiftLint's `colon` rule (globally disabled anyway) would normally flag this. Not a blocker, but worth being intentional about: if you're going to column-align switch cases, it should be applied consistently across all switches in the file (e.g. `sfSymbol` is unaligned). Inconsistent alignment is worse than no alignment.

---

### ✅ Things that are genuinely well done

- The `AnyView` → `@ViewBuilder` migration is the highest-value change here and it's done correctly. SwiftUI will now see `_ConditionalContent<GlassEffectView, MaterialFallbackView>` rather than `AnyView`, which restores structural diffing and view identity for the animation engine. This isn't cosmetic — it directly impacts list row reuse and transition animations.
- Removing the file-level `swiftlint:disable colon opening_brace` pair is correct. Leaving no-op suppressions is a maintenance hazard because future readers may assume those rules are active and write code to satisfy them.
- The `[SubtitleDebug#1029]` print removal is clean — the log referenced a closed issue number, confirming it was never cleaned up at source. Removing it inline with the surrounding dead variable (`result`) simplification is tighter than just deleting the print.
- The `periphery:ignore` reason comment on `BranchTagPill` is exactly the right level of detail — it explains both *why* Periphery flags it and *why* the ignore is valid.

---

**Net verdict:** Approve with the `rbOrangeTint` question as the one thing worth a quick check before merge. The `GlassButton @ViewBuilder` nit and doc phrasing items are non-blocking.

eonist:
## Deep Review — Batch 15

> Existing findings already covered by prior reviews (not re-raised here):
> - `RBStatus.sfSymbol` still uses `default:` while `tint` was made exhaustive → flagged by github-actions bot + Octopus.
> - `StatusBadgeBackground` macOS 26+ branch missing `.glassEffect(...)` → flagged by codeant-ai as Major.

---

### ✅ What's solid

**`AnyView` → `@ViewBuilder` removal** is the strongest change in this batch. The five affected `ViewModifier.body` implementations were previously forcing SwiftUI to wrap conditional branches in type-erased heap boxes on every render pass. With `@ViewBuilder`, the compiler now produces `_ConditionalContent<A, B>` at the call site — view identity is stable across diffs, animations that trigger on status change won't lose their anchor, and no allocations happen for the `AnyView` boxing. This is the kind of fix that's invisible in screenshots but matters a lot in list views like the runner panel where rows animate in/out under live status changes.

**`default:` → `case .unknown:` in `RBStatus.tint`** correctly mirrors what was done to `dotColor` in Batch 14. Future `RBStatus` cases will fail at compile time in `tint` instead of silently falling through.

**Debug `print` removal** from `runnerSubtitle` is overdue. That path fires on every render of every runner row. Even in a debug build, the string interpolation (building `rawArch`, `rawOS`, `result` descriptions) allocates every frame regardless of log level — there's no `os_log`-style lazy evaluation here.

---

### 🔴 Correctness issue: `GlassButton.body` got `@ViewBuilder` but is not in scope of the issue

Looking at the diff more carefully: `GlassButton.body` at line ~104 was also annotated with `@ViewBuilder`, but `GlassButton` was **not listed as a problem in issue #1095**. The issue only calls out `GlassCard`, `StatPillBackground`, `StatusBadgeBackground`, and `BranchTagPillBackground`.

Checking the diff, `GlassButton` did not have `AnyView` wrappers — the existing code already handled `if #available` correctly without `AnyView`. Adding `@ViewBuilder` to a body that never had `AnyView` is harmless (the compiler annotation is compatible), but it's scope creep from the issue and worth confirming: **did `GlassButton.body` previously contain `AnyView` wrappers that were removed in a prior batch?** If not, the `@ViewBuilder` annotation here is technically unnecessary noise on a function that didn't have the problem.

---

### 🟡 `rbOrangeTint` doc says "kept for call-site compatibility" — but is it actually used anywhere?

The updated doc comment is: `/// Duplicate of rbYellowTint — kept for call-site compatibility. Prefer rbYellowTint in new code.`

This implies there are existing call sites that use `rbOrangeTint`. If there aren't any — i.e. if Periphery or a manual search turns up zero usages — then the token should be deprecated with `@available(*, deprecated, renamed: "rbYellowTint")` or simply removed, not preserved with a compatibility note. "Kept for call-site compatibility" is load-bearing language that will confuse future readers if no call sites exist. Worth verifying with a quick `grep -r rbOrangeTint Sources/` before merge.

---

### 🟡 `PanelHeaderView.body` doc updated — but `PanelHeaderView` is `public`-surface-adjacent and the doc is incomplete

The new doc is: `/// Renders the header bar: system stats, optional sign-in button, settings and quit buttons.`

This is better than the tautological previous version, but `PanelHeaderView` has four stored properties (`statsVM`, `onQuit`, `onSelectSettings`, `onSignIn`) that are not documented at all at the type level. The `body` doc now describes the *output* but the *inputs* (the closure contracts — what triggers `onSignIn`, when is it visible, what does `statsVM` drive) are undocumented. Given that issue #1095 item 9 specifically called for replacing noise with *meaningful* API descriptions, the `body` doc here lands closer to structural description than API contract. Low priority, but the issue set a higher bar.

---

### 🟡 `RemovalAlertModifier` doc mentions `svc.sh`/`config.sh` and `gh auth login` but the modifier itself only passes through text

The new type-level doc says:
> Shows authentication-aware message text: authenticated users see the `svc.sh`/`config.sh` command summary; unauthenticated users see the `gh auth login` sign-in instructions.

But `RemovalAlertModifier` has no logic for auth state — it just renders whatever `title`, `message`, and `isPresented` binding are given to it. The auth-awareness is in the *call site* (presumably `SettingsView`), not in this modifier. The doc is accurate about the UX but misleading about where the logic lives — a future reader might look at the modifier's implementation for auth branching and find none. Consider scoping the doc to what the modifier actually owns: `/// Confirmation alert for runner removal. Presents a destructive action sheet with Cancel and Remove buttons; delegates confirmation to onConfirm and cancellation to onCancel.`

---

### 🟢 Minor: alignment spaces in `RBStatus.tint` are not applied consistently across the file

The PR adds alignment spaces in `tint`:
```swift
case .inProgress: return .rbBlueTint
case .success:    return .rbGreenTint
case .failed:     return .rbRedTint
case .queued:     return .rbYellowTint
case .unknown:    return .clear
```

But `sfSymbol` (same enum extension, same file) and other switches in the codebase don't use this alignment style. Not a correctness issue, but if SwiftLint's `colon` rule were ever re-enabled it would flag these. Since the batch is partly about consistency (removing stale suppressions), it's worth being intentional about introducing a new alignment pattern in isolation.

---

### 🟢 `// swiftlint:enable colon opening_brace` removal is correct — but double-check `colon` rule status

The PR correctly removes the file-level disable/enable pair because both rules are in `disabled_rules`. One edge case worth verifying: if `.swiftlint.yml` is ever updated to re-enable `colon` (e.g. as part of a future quality pass), the new alignment spaces introduced in `RBStatus.tint` will immediately trigger violations — the alignment spaces are technically non-compliant with `colon` style. That's a future risk, not a current bug, but it's a consequence of this batch's choices landing together.

---

**Summary verdict:** The batch is safe to merge. The `AnyView` removal is the real value here and it's done correctly. The two items to action before or shortly after merge are: (1) verify `rbOrangeTint` has actual call sites justifying the compatibility note, and (2) confirm whether `GlassButton` previously had `AnyView` wrappers. The `RemovalAlertModifier` doc framing is a nit but worth a one-line fix.

eonist:
## Deep Review — Batch 15

> **Pre-check:** I've read the full diff, issue #1095 (all 11 items), and all existing review threads (github-actions bot, codeant-ai, coderabbitai, Octopus, Greptile, and two prior comment-based reviews). Skipping everything already raised:
> - `RBStatus.sfSymbol` still uses `default:` → flagged by github-actions, Octopus, Greptile, and both prior deep reviews.
> - `StatusBadgeBackground` macOS 26+ missing `.glassEffect` → flagged by codeant-ai (and correctly debunked as a false positive by the prior deep review — the tinted capsule is intentional design, not a regression).
> - `rbOrangeTint` "kept for call-site compatibility" without verifying actual call sites → flagged by both prior deep reviews.
> - `GlassButton.body` gained `@ViewBuilder` without ever having `AnyView` → flagged by both prior deep reviews.
> - `RemovalAlertModifier` doc mentioning `svc.sh`/`config.sh` auth logic lives in the modifier when it doesn't → flagged by the second prior deep review.
> - `metaTrailing` comment using "snaps off" implying animation that doesn't exist → flagged by the second prior deep review.
> - Alignment whitespace in `tint` switch inconsistent with `sfSymbol` and the rest of the file → flagged by both prior deep reviews.
> - `PanelHeaderView` stored property contracts undocumented → flagged by the second prior deep review.
> - `@available(*, deprecated)` on `rbOrangeTint` as enforcement mechanism → flagged by the first prior deep review.

Everything below is new signal.

---

### 🔴 `rbOrangeTint` was removed, not documented — the PR description says "clarified", but the token is actually gone

The PR description reads: `rbOrangeTint doc clarified: duplicate of rbYellowTint, not an alias`. Issue #1095 item 7 says: `Update the doc comment to: "Duplicate of rbYellowTint — kept for call-site compatibility. Prefer rbYellowTint in new code."` — i.e., **keep** the token, improve the doc.

The diff shows `rbOrangeTint` was **deleted entirely**, not doc-updated:

```diff
-    /// Low-opacity orange tint — alias for `rbYellowTint`.
-    public static let rbOrangeTint = rbWarning.opacity(0.08)
```

If `rbOrangeTint` genuinely has zero call sites this is the right call, and the prior reviews suggest grepping is needed to confirm. But the PR description claims a doc clarification happened, while the actual change was a deletion. These are different operations with different risk profiles. The PR description should accurately reflect this as a removal, not a clarification — especially since #1095 item 7 explicitly said to retain the token with updated docs. Either the issue spec was deliberately deviated from (fine, if justified) or the removal was accidental. Either way, it should be explicit.

---

### 🟡 `PanelLocalRunnerRow.body` doc diverges from the issue spec table

Issue #1095 item 9 specifies the exact replacement doc for `PanelLocalRunnerRow`:

> `/// Renders the runner card list, or nothing if runners is empty.`

The diff writes:

```swift
/// Renders the runner card list; emits nothing (no EmptyView placeholder) when `runners` is empty.
```

This is a *better* doc than the spec (it clarifies the `EmptyView` distinction, which is SwiftUI-relevant), but it's worth confirming this deviation from the issue spec is intentional. The two prior deep reviews both noted the SwiftUI structural implication is real — an undocumented single-arm `if` in a `@ViewBuilder` context produces `Optional<some View>` not `EmptyView`, and a parent `VStack` will size accordingly. So the longer form is more accurate. This isn't a bug, but it's worth acknowledging the spec was deliberately improved.

---

### 🟡 `PanelHeaderView.body` doc says "sign-in button (shown when unauthenticated)" — but the prior review's suggested wording wasn't the actual wording in the diff

The second prior deep review suggested changing "optional sign-in button" to "sign-in button (shown when unauthenticated)". Looking at the actual diff, the landed doc is:

```swift
/// Renders the header bar: system stats, sign-in button (shown when unauthenticated), settings and quit buttons.
```

This is exactly the suggested improvement — so this is already fixed relative to what the second review asked for. ✅ No action needed, just noting for completeness that this thread is resolved in the diff.

---

### 🟡 No call-site guard exists for the removed `rbOrangeTint` — Swift won't error, Periphery won't warn

Unlike removing a `public` symbol from a library (which breaks compilation for downstream consumers), removing an `internal static let` from an app target silently compiles. If `rbOrangeTint` was used anywhere:
- `grep -r rbOrangeTint Sources/` — would catch direct usage
- But it's also used via `.rbOrangeTint` in view modifier chains, e.g. `.foregroundStyle(.rbOrangeTint)`, which grep would catch

The open question from prior reviews was whether it has callers. Given the deletion was committed and CI passed (SonarQube Quality Gate ✅, 0 new issues), it's effectively confirmed unused. But the PR description doesn't acknowledge this reasoning. A note like "zero call sites confirmed, deletion preferred over deprecation" would close the loop for future historians reading git blame.

---

### 🟢 `RemovalAlertModifier.body` — `@ViewBuilder` omitted, unlike all four sibling modifiers

Every other `ViewModifier.body` in this PR that has an `if #available` branch got `@ViewBuilder`. `RemovalAlertModifier.body` does not have an `if #available` branch — its body is a single linear chain `content.alert(...)`. So `@ViewBuilder` is neither needed nor applicable here. This is consistent and correct. Noting it explicitly because the pattern of adding `@ViewBuilder` to `body` in this PR could tempt a future reader to "complete the set" — it shouldn't.

---

### 🟢 `BranchTagPillBackground` macOS 26+ branch composes `.background` *then* `.glassEffect` — order is significant

```swift
content
    .background(Color.rbAccent.opacity(0.15), in: Capsule())
    .glassEffect(.regular, in: Capsule())
```

Applying `.background` before `.glassEffect` means the tint layer is rendered *under* the glass material. This is the correct composition order for the intended effect (tint colours the frosted glass), but it's the opposite of what you'd do if you wanted the tint to sit on top of a base glass material. The `GlassCard` equivalent uses `.glassEffect` first then `.overlay` for the stroke — a different composition pattern. Neither is wrong, but the visual semantics depend on this order. Given that this code was pre-existing and is unchanged by this PR (only the `AnyView` wrapper was removed), there's no regression risk. But since the `@ViewBuilder` removal now makes this code more readable, it's worth a comment explaining why the tint precedes the glass. Not blocking.

---

**Net verdict:** One genuine discrepancy to close — the `rbOrangeTint` description in the PR body says "clarified" when the code says "deleted." Everything else in this batch is clean, the issue spec was followed to a high standard on all 11 items, and the `AnyView` → `@ViewBuilder` migration is the correct and complete fix for the stated problem.

eonist:
## Deep Review — Batch 15

> **Pre-check complete.** I've read the full diff, issue #1095 (all 11 items), all 4 inline review threads, and all prior PR comments. The following items have already been raised and are **not repeated here**:
> - `RBStatus.sfSymbol` `default:` → `case .unknown:` exhaustiveness gap — flagged by github-actions, Octopus, and the prior deep reviews
> - `StatusBadgeBackground` missing `.glassEffect` on macOS 26+ — flagged by codeant-ai (and subsequently rebutted as a false positive in the prior review — I agree with the rebuttal; the tinted-capsule design is intentional)
> - `rbOrangeTint` "kept for call-site compatibility" language / potential zero-call-site issue — raised in both prior deep reviews
> - `GlassButton.body` `@ViewBuilder` scope creep — raised in both prior deep reviews
> - `metaTrailing` comment wording ("snaps off") — raised in prior deep review
> - `PanelHeaderView.body` "optional sign-in button" phrasing — raised in prior deep review
> - `RemovalAlertModifier` doc auth-logic-placement mislead — raised in prior deep review
> - Alignment spaces in `tint` switch / consistency across file — raised in prior deep review

Everything below is **new signal only**.

---

### 🔴 `rbOrangeTint` was removed but the doc comment in issue #1095 item 7 called for *updating* it, not removing it

Issue #1095 item 7 explicitly specifies:
> **Fix:** Update the doc comment to: `/// Duplicate of rbYellowTint — kept for call-site compatibility. Prefer rbYellowTint in new code.`

The diff removes `rbOrangeTint` entirely:
```diff
-    /// Low-opacity orange tint — alias for `rbYellowTint`.
-    public static let rbOrangeTint = rbWarning.opacity(0.08)
```

No doc-update was applied because the token was deleted. The issue spec said to *preserve* it with an updated doc comment. Deleting it is a stronger action that's either correct (if zero call sites exist) or a breaking change (if any exist). The batch description says "Remove `rbOrangeTint`" but the issue only says "Fix the doc comment — clarify duplicate, not alias." This is a spec deviation. Before merge, confirm this deletion was intentional and that Periphery / a manual grep confirms zero usages of `rbOrangeTint` across the whole target.

---

### 🟡 `RBStatus.tint` exhaustiveness: the alignment of the new `case .unknown:` arm against existing arms differs from `sfSymbol`'s style — but more importantly, `case .unknown:` appears in the diff with **trailing alignment spaces that disagree with the `inProgress` arm**

Look at the diff output:
```swift
case .inProgress: return .rbBlueTint
case .success:    return .rbGreenTint
case .failed:     return .rbRedTint
case .queued:     return .rbYellowTint
case .unknown:    return .clear
```

`.inProgress` uses one space after the colon (no alignment), while the four others use 4 trailing spaces. This is fine visually — `.inProgress` is the longest arm and sets the column. But the existing code had a single `case .inProgress:` arm with no alignment. Now `.inProgress` is the odd one out in the opposite direction. For an enum that may gain a new case longer than `.inProgress` (e.g. `.pendingApproval`), all the alignment spaces will need to be redone again. Not a blocker, but this is a debt deposit.

---

### 🟡 `StatusBadgeBackground` doc updated to explicitly say "intentionally does NOT use `.glassEffect`" — but `BranchTagPillBackground` on macOS 26+ still applies *both* `background(color.opacity(0.15))` AND `.glassEffect`, which stacks a tint under glass

```swift
// BranchTagPillBackground macOS 26+ branch:
content
    .background(Color.rbAccent.opacity(0.15), in: Capsule())
    .glassEffect(.regular, in: Capsule())
```

This double-layer (tinted fill + glass on top) is unchanged from the pre-image, so it's not a regression. But now that `StatusBadgeBackground` explicitly documents it *doesn't* do this, and `StatPillBackground` uses only `.glassEffect` with no tint layer, there are now three distinct visual patterns across four "pill" modifiers with no type-level documentation of why they differ:

| Modifier | macOS 26+ strategy |
|---|---|
| `GlassCard` | `.glassEffect` + stroke overlay |
| `StatPillBackground` | `.glassEffect` only |
| `StatusBadgeBackground` | tint + stroke overlay, no glass |
| `BranchTagPillBackground` | tint + `.glassEffect` stacked |

`StatusBadgeBackground` now has a "why no glass" note added by this PR. The others have none. The design intent is undocumented at the point where a future developer will naturally ask "why does `BranchTagPill` stack tint + glass but `StatPill` doesn't?" This batch is explicitly about adding clarity — a single-line `// Note:` on `BranchTagPillBackground` (similar to the one added to `StatusBadgeBackground`) would complete the picture. Not blocking, but it's an opportunity this batch is positioned to take.

---

### 🟡 `PanelLocalRunnerRow.body` new doc says "emits nothing (no EmptyView placeholder) when `runners` is empty" — but the *existing* review comments already flagged a softer version of this and the actual implementation has a subtlety

The new doc (from a prior review pass):
```swift
/// Renders the runner card list; emits nothing (no EmptyView placeholder) when `runners` is empty.
```

The implementation:
```swift
var body: some View {
    if !runners.isEmpty { runnerList(runners) }
}
```

This is a one-armed `if` in a `@ViewBuilder` context. SwiftUI resolves this to `Optional<RunnerListView>` (i.e. `_ConditionalContent<RunnerListView, EmptyView>` — actually it IS `EmptyView` at the SwiftUI level even without an explicit `else { EmptyView() }`). The doc says "no EmptyView placeholder" but SwiftUI internally does produce an empty slot. The practical impact is zero for rendering, but the doc makes a claim about internal SwiftUI behaviour that's slightly inaccurate. The correct claim is "renders nothing visible" not "emits nothing" — because structurally, the view tree does contain a node. This matters if anything in the parent ever uses `ViewInspector` or `PreviewProvider` and tries to count child views.

---

### 🟢 `SectionHeaderLabel.body` doc says "with standard panel insets" — but no insets are applied in `body`

New doc:
```swift
/// Renders the uppercased title in secondary caption color with standard panel insets.
```

Looking at the implementation in the pre-image (unchanged by this diff):
```swift
var body: some View {
    Text(title.uppercased())
        .font(RBFont.sectionCaption)
        .foregroundColor(.secondary)
        // ... no .padding() call shown in this diff
}
```

If `SectionHeaderLabel` doesn't apply insets itself (they're likely applied at the call site), the doc is inaccurate. "Standard panel insets" implies this view owns its spacing, which would mislead someone trying to override the inset at a call site. Worth verifying whether `.padding` is actually called in `body` — if not, remove "with standard panel insets" from the doc.

---

### 🟢 `RemovalAlertModifier` — removed `swiftlint:disable:next multiple_closures_with_trailing_closure` but no check that the rule removal doesn't surface in a context where the global disable might be overridden

The removal is correct and the global `.swiftlint.yml` disable is cited as justification. However: if `multiple_closures_with_trailing_closure` is ever moved from `disabled_rules` to `opt_in_rules` or enabled inline in a nested config, the `.alert(...)` call site will immediately become a lint violation with no inline suppression remaining. This is a structural fragility, not a bug today. The prior inline comment actually served as documentation that "this pattern would trigger `multiple_closures_with_trailing_closure`." Its removal is correct per the issue but leaves no trace of that knowledge. A `// Note:` comment on the `.alert` call explaining the trailing closure pattern (without referencing SwiftLint) would preserve the intent without the suppression.

---

### ✅ Things genuinely well-executed (new observations, not re-stating prior reviews)

- **The `BranchTagPill` periphery comment** (`// periphery:ignore — used dynamically inside ActionRowView.rowContent; Periphery cannot trace the call site`) is the right level of detail: it explains both *why* Periphery flags it and *why* the ignore is valid, making it safe to automate suppression audits later without manual investigation.
- **`RemovalAlertModifier` doc** replacing `/// Extracted from SettingsView.body to satisfy function_body_length and file_length limits.` is a genuine improvement. The old doc was scaffolding rationale (useful when writing the code, useless forever after). The new doc describes the API contract callers need.
- **Removing the file-level `swiftlint:disable colon opening_brace` pair** cleanly is harder to do right than it sounds — you need to confirm the rules are globally disabled, not just locally overridden. The PR description correctly cites `.swiftlint.yml` as the authority.

---

**Net verdict:** The spec deviation on `rbOrangeTint` (deleted vs. doc-updated) is the one item that needs explicit confirmation before merge. Everything else is either already actioned, a non-blocking nit, or follow-up material for #1038. The core work — `AnyView` → `@ViewBuilder`, exhaustive `tint`, `public` removal, debug print — is solid.

eonist:
## Deep Review — Batch 15

I've read the diff in full, the originating issue (#1095), and all existing bot reviews. I won't re-raise anything already flagged. Here's what I found on top of that.

---

### ✅ What's solid

- All 11 items from #1095 are addressed. The issue was thorough and the implementation matches it faithfully.
- `@ViewBuilder` on `body` is the right fix for `AnyView` erasure — identity and diffing are restored across all four modifiers.
- `case .unknown:` replacing `default:` in `tint` is the correct exhaustiveness pattern (consistent with Batch 14's `dotColor` fix).
- Removing the `print("[SubtitleDebug#1029]…")` from the hot render path is important — that fired on every render of every row.
- Removing no-op SwiftLint directives is a genuine readability improvement, not just cosmetic.

---

### 🔴 Real issue: `rbOrangeTint` was deleted, not updated as the issue specified

Issue #1095 §7 says:

> **Fix:** Update the doc comment to: `/// Duplicate of rbYellowTint — kept for call-site compatibility. Prefer rbYellowTint in new code.`

The fix applied in this PR **deletes** `rbOrangeTint` entirely instead. This is a breaking change if any call site uses `rbOrangeTint` — the issue explicitly said to keep it for call-site compatibility. Before merging, verify there are no remaining references to `rbOrangeTint` in the codebase; if there are, this will be a compile error. If there are genuinely zero call sites, deletion is fine (and cleaner than the issue proposed), but the PR description should note the deviation from the issue spec and confirm the search was done.

---

### 🟡 `StatusBadgeBackground` comment update contradicts the new doc, not a regression

The existing CodeAnt bot flagged this as a "behavioral regression" — but the PR's own updated doc comment explicitly says:
> *"intentionally does NOT use `.glassEffect` — the tinted capsule design is distinct from the glass card pattern"*

The old pre-PR doc was wrong (said it used glassEffect when it never did). This is a **doc correction**, not a regression. The CodeAnt comment should be dismissed. That said — the comment immediately above the struct still reads:
```
/// macOS 26+: colour tint layer + stroke overlay on a `Capsule()` shape.
```
…while the implementation on macOS 26+ path has: `background(color.opacity(0.25), in: Capsule())` + stroke overlay. The doc is now accurate. ✅ No action needed here.

---

### 🟡 `PanelLocalRunnerRow` doc diverges from the issue's specified wording

Issue #1095 §9 specified:
> `/// Renders the runner card list, or nothing if runners is empty.`

The PR uses:
> `/// Renders the runner card list; emits nothing (no EmptyView placeholder) when runners is empty.`

The parenthetical `(no EmptyView placeholder)` is an implementation detail leaking into the API doc — it names a framework type (`EmptyView`) that the call site has no reason to know about. The issue's wording was cleaner. Minor, but since the whole point of this batch is doc quality, worth aligning.

---

### 🟡 Inline comment wording in `metaTrailing` — minor imprecision

The comment added reads:
```swift
// Hidden when not running or queued — prevents a frozen elapsed display after completion.
```

Issue #1095 §11 specified:
> `// Elapsed shown only while running or queued — snaps off on completion to avoid a frozen clock.`

Both are correct in intent, but the issue's wording is slightly more precise: "snaps off" is more descriptive of the UX behavior than "prevents a frozen display." The implemented comment says "hidden when not running or queued" which is a double-negative framing. Either works, but consistency with the tracking issue would be cleaner.

---

### 🟢 `sfSymbol` exhaustiveness (already raised by bots, but worth noting the suggested value)

The two bot comments disagree on what the unknown fallback symbol should be (`""` vs `"questionmark"`). An empty string will produce a broken `Image(systemName: "")` rendering. `"questionmark"` is the correct SwiftUI-safe fallback for an unknown symbol. This was already flagged but the correct value to use is `"questionmark"`, not `""`.

---

### 🟢 `RemovalAlertModifier` doc: "supplied by the call site" is inaccurate (CodeRabbit flagged this)

Already noted by CodeRabbit — the strings are hard-coded inside the modifier; only the `isAuthenticated` bool is call-site supplied. The fix is one line. Worth doing before merge.

---

### Summary

| # | Severity | Item | Action |
|---|---|---|---|
| 1 | 🔴 | `rbOrangeTint` deleted vs issue spec of "keep + update doc" | Confirm zero call sites or add note to PR |
| 2 | 🟡 | `PanelLocalRunnerRow` doc has leaky `(no EmptyView placeholder)` | Trim to issue spec wording |
| 3 | 🟡 | `metaTrailing` inline comment is double-negative framing | Polish to match issue wording |
| 4 | 🟢 | `sfSymbol` default → `case .unknown` (bot-raised) | Use `"questionmark"`, not `""` |
| 5 | 🟢 | `RemovalAlertModifier` doc "supplied by call site" (bot-raised) | One-line fix |

The `StatusBadgeBackground` CodeAnt "regression" finding is a false positive — dismiss it.

eonist:
## Post-`f3d50f7` Action Items — Consolidated

Cross-referencing the three latest deep reviews ([#4602227590](https://github.com/eoncode/runner-bar/pull/1096#issuecomment-4602227590), [#4602227381](https://github.com/eoncode/runner-bar/pull/1096#issuecomment-4602227381), [#4602227000](https://github.com/eoncode/runner-bar/pull/1096#issuecomment-4602227000)) against each other. All three independently converge on the same three blocking/near-blocking items. Below is the de-duplicated, priority-ordered list.

---

### 🔴 Before merge

**1. `rbOrangeTint` — description says "clarified", diff says "deleted"**

All three reviews flag this as the top item. Issue #1095 §7 said *update the doc comment, keep the token*. The diff removes it entirely. Two states:

- **If zero call sites (likely — CI + SonarQube 0 new issues):** deletion is correct and cleaner than a stale deprecation comment. Update the PR description from `"rbOrangeTint doc clarified"` to `"rbOrangeTint removed — zero call sites confirmed, deletion preferred over doc-only deprecation"`.
- **If a call site exists:** compile error that CI would have caught — but the description still needs correcting from "clarified" to "removed".

The SonarQube clean pass is strong corroborating evidence for zero call sites. The PR body just needs to say what actually happened.

---

### 🟡 Should-fix (non-blocking but cleanly actionable)

**2. Strip `@ViewBuilder` from `GlassButton.body`**

Flagged by two of the three reviews. `GlassButton` had no `AnyView` wrappers — it was not in scope of #1095 item 4. `@ViewBuilder` on a `some View`-returning function with a single `if #available` and no divergent branch types is a no-op. The annotation is harmless but misleads readers about the rationale ("AnyView removal" is the stated driver; `GlassButton` had none to remove). One-line deletion.

**3. `PanelLocalRunnerRow.body` doc — soften the `EmptyView` claim**

Two of the three reviews independently flag this. The current doc: `"emits nothing (no EmptyView placeholder) when runners is empty"`. The structural reality: SwiftUI's `@ViewBuilder` resolves a one-armed `if` to `_ConditionalContent<RunnerListView, EmptyView>` — there *is* a structural `EmptyView` node internally, it's just zero-size visually. The doc claim is accurate for rendering but inaccurate from a `ViewInspector`/introspection standpoint, and it names a framework type that call sites have no reason to know about.

Suggested replacement (matches the spirit of all three reviews):
```swift
/// Renders the runner card list; renders nothing visible when `runners` is empty.
```

---

### 🟢 Follow-up material (not this batch)

- **`BranchTagPillBackground` tint-under-glass composition** — order is semantically correct (tint renders under the frosted glass) but unexplained. Now that `AnyView` is gone and the code is readable, a `// Note:` similar to the one added to `StatusBadgeBackground` would complete the pill modifier documentation picture. Suggested for #1038.
- **`SectionHeaderLabel.body` "with standard panel insets"** — if no `.padding()` is called inside `body` (insets may live at call sites), this doc claim misleads anyone trying to override spacing. Quick verification: if no `.padding()` in `body`, drop the clause.
- **`RemovalAlertModifier` trailing closure knowledge** — the removed `swiftlint:disable:next multiple_closures_with_trailing_closure` served as implicit documentation. A `// Note:` on the `.alert` call (explaining the trailing closure shape, without referencing SwiftLint) would preserve that intent non-fragily.

---

**tl;dr for merge gate:** (1) Fix the PR description to say `rbOrangeTint` was *removed*, not *clarified*, and confirm zero call sites. (2) Drop `@ViewBuilder` from `GlassButton.body`. (3) Soften `PanelLocalRunnerRow` doc to "renders nothing visible". Everything else is #1038 follow-up.

Issue #992, opened 2026-05-28
♻️ Move ScopeDetailView to popover/sheet style (explicit Save/Cancel)


eonist:
Closes #989

## Rationale

Since `ScopeDetailView` edits persisted data, the user needs to explicitly confirm or discard changes. The current navigation drill-down allows pressing back with no clear save/discard decision. A sheet with explicit **Cancel** and **Save** buttons — matching the pattern used by `AddScopeSheet` — enforces intentional data commitment.

## Plan

See sub-issues for each step:

1. Convert `ScopeDetailView` to a sheet-style presentation
2. Add local draft state (stage edits, only persist on Save)
3. Update `NavState` + `AppDelegate+Navigation`
4. Update `SettingsView` to open as a sheet
5. Verify `NSOpenPanel` works in sheet context
6. Clean up `NavState` restore logic

## Files affected

| File | Change |
|------|--------|
| `Sources/RunnerBar/Scope/ScopeDetailView.swift` | Convert to sheet: replace `headerBar`/`onBack` with Cancel/Save footer; batch writes behind Save |
| `Sources/RunnerBar/App/NavState.swift` | Remove `.scopeDetail` case |
| `Sources/RunnerBar/App/AppDelegate+Navigation.swift` | Remove `scopeDetailView()` factory; update `validatedView` fallback |
| `Sources/RunnerBar/Views/Settings/SettingsView.swift` | Replace `onSelectScope` callback with `@State` sheet presentation |

Issue #602, opened 2026-05-21
refactor(phase-1): extract NavState → NavState.swift


eonist:
Part of #601

## Problem
`NavState` is a private enum declared inside `AppDelegate.swift`. It is a self-contained navigation concept with no dependency on `AppDelegate` internals.

## Change
- Move `NavState` enum out of `AppDelegate.swift` into a new `Sources/RunnerBar/NavState.swift`
- Change access level from `private` to `internal` (file-private no longer applies)
- No logic changes

## Acceptance
- CI green
- `AppDelegate.swift` no longer contains the `NavState` declaration

eonist:
✅ Resolved in commit [`42f68ba`](https://github.com/eoncode/runner-bar/commit/42f68ba07aa86535c4bddca4e7b7fdce502b63b8) on `refactor/601-codebase-cleanup` (PR #612).\n\n`private enum NavState` extracted from `AppDelegate.swift` into its own file `Sources/RunnerBar/NavState.swift`. No logic changes. CI green (SwiftLint ✅ SonarCloud ✅).

Issue #455, opened 2026-05-17
⭐ Add new hierarchy level. the steps


eonist:
The task is to add steps under jobs in main view. Details bellow:

- We could just add the hierarchy lines and keep the steps bellow the job text. 
- but in the same component as the job, within the same background container etc. 
- then we could add a left indicator in the job item that toggles compact and expanded. (or ditch detail and let entire items toggle hierarchy etc)
- we ditch workflow detail view and job detail vie entirly. their functionality is captured in our context menues for each item
- since we dont have detailview for job or workflow, we should remove the chevron icon from their row
- and no left indicator for job item. we only should have left indeicator for workflows, as it is now
- and left indicator in workflow do not trigger expand collapse anymore. tapping entire row triggers expand / collapse
- only step has detail view. when we tap a step row it opens the log detail view. 
- we must add context menu to the step row im not sure we have it right now
- when we tap a job we should collapse and expand to show all steps under it
- step row will have. chevron at the far right indicating it opens a detail view

related nested workflows: https://github.com/eoncode/runner-bar/issues/566 (we wont add this now. we will just do flat jobs for now)

eonist:
Here's a summary of [[issue #455](https://github.com/eoncode/runner-bar/issues/455)](https://github.com/eoncode/runner-bar/issues/455) in **eoncode/runner-bar**:

## ⭐ Add New Hierarchy Level: Steps

The goal is to introduce a **steps layer** underneath jobs in the main view . Currently the hierarchy stops at jobs, and this issue extends it one level deeper to show individual CI steps inline.

## Key Design Decisions

- Steps are rendered **inside the same job component/container** — no separate component 
- **Ditch the workflow detail view and job detail view** entirely — their functionality moves to context menus on each row 
- **Chevron icon removed** from workflow and job rows (no more detail view navigation for them) 
- The **left indicator on workflows** stays, but now **tapping the entire workflow row** triggers expand/collapse (not just the indicator) 
- **Tapping a job row** expands/collapses to show/hide its steps 
- **Only steps have a detail view** — tapping a step opens the log detail view 
- Step rows get a **chevron on the far right** indicating they open a detail view 
- A **context menu must be added** to step rows (currently likely missing) 

## Out of Scope

Nested workflows (issue (https://github.com/eoncode/runner-bar/issues/566)) are explicitly excluded for now — the implementation will use a flat jobs structure only .

[...]

PR #1098, opened 2026-06-03
Batch 16 — App state files doc & comment cleanup (main, Exports, NavState, PanelSheetState, PanelVisibilityState)

### LINKED ISSUE #1097: Batch 16 — App state files doc & comment cleanup (main, Exports, NavState, PanelSheetState, PanelVisibilityState)
Tracking issue for Batch 16 of the ongoing cleanup project (#1038).

## Files in scope

- `Sources/RunnerBar/main.swift`
- `Sources/RunnerBar/Exports.swift`
- `Sources/RunnerBar/App/NavState.swift`
- `Sources/RunnerBar/App/PanelSheetState.swift`
- `Sources/RunnerBar/App/PanelVisibilityState.swift`

---

### `main.swift` — ✅ Confirmed clean
Fully documented. The `MainActor.assumeIsolated` wrapper has a clear ❌ guard comment explaining why it must never be removed. No changes needed — mark as confirmed clean tombstone.

### `Exports.swift` — ✅ Confirmed clean
`@_exported import RunnerBarCore` design decision is well explained including the visibility-leakage warning. No changes needed.

### `NavState.swift` — 🟡 Minor gaps

1. **Duplicate comment block** — The file-level `//` comment block above the `///` doc comment repeats the same content as the doc comment. The file block should be trimmed to history-only context (`#455`, `#992`, `#1001` removal notes) so it adds signal the doc can't; the current duplication just adds noise.
2. **Associated value docs missing** — `case stepLog(ActiveJob, JobStep)` has no docs on its two associated values. A future reader won't know which is the job context and which is the selected step. Add inline labels or a doc note.

### `PanelSheetState.swift` — 🟡 Two doc gaps

1. **`runnerSheetSnapshot` property doc** — Currently says `"Snapshot captured immediately before a transient hide"` but doesn't explain *why* the snapshot is needed (the NSPopover sheet-teardown behaviour described in the type doc). Should cross-reference: `/// Backing store for captureTransientHideState() — persists sheet intent across NSPopover hide/show cycles.`
2. **`restoreTransientHideStateIfNeeded()` guard semantics** — The `guard editingRunner == nil` condition is subtly important: it prevents overwriting an already-active runner if both paths run close together. This should be noted inline: `// Only restore if no sheet is already active — prevents overwriting a concurrently set runner.`

### `PanelVisibilityState.swift` — 🟢 One threading note

1. **`@MainActor` annotation absent** — `PanelVisibilityState` is not `@MainActor` but all its `@Published` mutations happen on the main thread by contract with `AppDelegate`. Under Swift 6 strict concurrency this will surface as a warning. Add a class-level comment: `// Note: mutated exclusively on the main thread by AppDelegate — no @MainActor annotation because PanelContainerView reads it via @EnvironmentObject which handles dispatch.`

---

## Checklist

- [ ] `main.swift` — confirmed clean, no changes
- [ ] `Exports.swift` — confirmed clean, no changes
- [ ] `NavState.swift` — trim duplicate file-level comment block; add associated value docs on `stepLog`
- [ ] `PanelSheetState.swift` — improve `runnerSheetSnapshot` doc; add guard-semantics comment in `restoreTransientHideStateIfNeeded()`
- [ ] `PanelVisibilityState.swift` — add `@MainActor`/threading note on class declaration
- [ ] Branch `batch-16-app-state-cleanup` created
- [ ] PR opened, CI green

eonist:
## User description
Resolves #1097

Batch 16 of the ongoing cleanup project (#1038).

## Files in scope

- `Sources/RunnerBar/main.swift`
- `Sources/RunnerBar/Exports.swift`
- `Sources/RunnerBar/App/NavState.swift`
- `Sources/RunnerBar/App/PanelSheetState.swift`
- `Sources/RunnerBar/App/PanelVisibilityState.swift`

## Changes

- `main.swift` — confirmed clean, no changes
- `Exports.swift` — confirmed clean, no changes
- `NavState.swift` — trimmed duplicate file-level `//` comment block to history-only; added associated value labels (`job:`, `step:`) and `- Parameters:` docs on `case stepLog(ActiveJob, JobStep)`
- `AppDelegate+Navigation.swift` — updated `.stepLog(job, step)` construction site to `.stepLog(job: job, step: step)` to match the label addition above; **note: adding labels to associated values is a source-breaking API change for external construction sites — confirmed via grep that this is the only construction site in the codebase**
- `PanelSheetState.swift` — improved `runnerSheetSnapshot` doc to cross-reference NSPopover teardown context; added guard-semantics inline comment in `restoreTransientHideStateIfNeeded()`; **also fixes a latent bug: `self.runnerSheetSnapshot = nil` added after restore so the snapshot is consumed on first use and cannot cause a spurious second restore in a subsequent popover hide/show cycle**; removed unused `import Foundation`
- `PanelVisibilityState.swift` — added `@MainActor`/threading note above class declaration as `/// - Note:` DocC format; corrected wording to make the invariant caller-owned (AppDelegate guarantees main-thread mutation) rather than SwiftUI-owned

## Checklist

- [x] `main.swift` — confirmed clean, no changes
- [x] `Exports.swift` — confirmed clean, no changes
- [x] `NavState.swift` — trim duplicate file-level comment block; add associated value labels + docs on `stepLog`
- [x] `AppDelegate+Navigation.swift` — update call site to match label change; grep confirms only construction site
- [x] `PanelSheetState.swift` — improve `runnerSheetSnapshot` doc; add guard-semantics comment; fix latent snapshot-not-cleared bug; remove unused `import Foundation`
- [x] `PanelVisibilityState.swift` — add `@MainActor`/threading note on class declaration
- [x] Branch `batch-16-app-state-cleanup` created
- [ ] PR opened, CI green

### DIFF
--- a/Sources/RunnerBar/App/AppDelegate+Navigation.swift
+++ b/Sources/RunnerBar/App/AppDelegate+Navigation.swift
@@ -40,7 +40,7 @@ extension AppDelegate {
             store: observable,
             onStepTap: { [weak self] job, step in
                 guard let self else { return }
-                self.savedNavState = .stepLog(job, step)
+                self.savedNavState = .stepLog(job: job, step: step)
                 self.navigate(to: self.wrapEnv(StepLogView(
                     job: job,
                     step: step,
--- a/Sources/RunnerBar/App/NavState.swift
+++ b/Sources/RunnerBar/App/NavState.swift
@@ -4,12 +4,7 @@ import RunnerBarCore
 
 // MARK: - NavState
 //
-// Represents the currently visible navigation screen inside the RunnerBar panel.
-//
-// Extracted from AppDelegate.swift (#602) — was a private enum co-located with
-// AppDelegate. Moved here so navigation cases can be extended without opening
-// AppDelegate.
-//
+// History:
 // #455: Removed .jobDetail, .actionDetail, .actionJobDetail, .actionStepLog.
 // Navigation from the main view now goes directly: inline step tap → .stepLog.
 // #992: Removed .scopeDetail — ScopeEditSheet is now a modal sheet presented
@@ -17,10 +12,17 @@ import RunnerBarCore
 // #1001: Removed .runnerDetail — runner editing is now a popover in SettingsView.
 
 /// Represents the currently visible navigation screen inside the RunnerBar panel.
+///
+/// Extracted from AppDelegate.swift (#602) — was a private enum co-located with
+/// AppDelegate. Moved here so navigation cases can be extended without opening
+/// AppDelegate.
 enum NavState {
     /// The root popover showing runners and the recent-actions list.
     case main
     /// The raw log for a single step, reached from the main inline step row.
+    /// - Parameters:
+    ///   - job: The active job providing context for the selected step.
+    ///   - step: The specific step whose log is displayed.
     case stepLog(ActiveJob, JobStep)
     /// The Settings sheet.
     case settings
--- a/Sources/RunnerBar/App/NavState.swift
+++ b/Sources/RunnerBar/App/NavState.swift
@@ -23,7 +23,7 @@ enum NavState {
     /// - Parameters:
     ///   - job: The active job providing context for the selected step.
     ///   - step: The specific step whose log is displayed.
-    case stepLog(ActiveJob, JobStep)
+    case stepLog(job: ActiveJob, step: JobStep)
     /// The Settings sheet.
     case settings
 }
--- a/Sources/RunnerBar/App/PanelSheetState.swift
+++ b/Sources/RunnerBar/App/PanelSheetState.swift
@@ -17,7 +17,8 @@ final class PanelSheetState: ObservableObject {
     /// The runner currently selected for the runner detail sheet.
     @Published var editingRunner: RunnerModel?
 
-    /// Snapshot captured immediately before a transient hide.
+    /// Backing store for captureTransientHideState() — persists sheet intent
+    /// across NSPopover hide/show cycles. See type doc for NSPopover teardown context.
     private var runnerSheetSnapshot: RunnerModel?
 
     /// Captures the current runner sheet before hiding the popover.
@@ -27,6 +28,7 @@ final class PanelSheetState: ObservableObject {
 
     /// Restores the runner sheet after the popover has been shown again.
     func restoreTransientHideStateIfNeeded() {
+        // Only restore if no sheet is already active — prevents overwriting a concurrently set runner.
         guard editingRunner == nil, let runnerSheetSnapshot else { return }
         editingRunner = runnerSheetSnapshot
     }
--- a/Sources/RunnerBar/App/PanelSheetState.swift
+++ b/Sources/RunnerBar/App/PanelSheetState.swift
@@ -27,7 +27,7 @@ final class PanelSheetState: ObservableObject {
 
     /// Restores the runner sheet after the popover has been shown again.
     func restoreTransientHideStateIfNeeded() {
-        // Only restore if no sheet is already active — prevents overwriting a concurrently set runner.
+        // Only restore if no sheet is already active — prevents overwriting a runner set after the snapshot was captured.
         guard editingRunner == nil, let runnerSheetSnapshot else { return }
         editingRunner = runnerSheetSnapshot
         self.runnerSheetSnapshot = nil
--- a/Sources/RunnerBar/App/PanelSheetState.swift
+++ b/Sources/RunnerBar/App/PanelSheetState.swift
@@ -1,7 +1,6 @@
 // PanelSheetState.swift
 // RunnerBar
 import Combine
-import Foundation
 import RunnerBarCore
 
 // MARK: - PanelSheetState
@@ -31,6 +30,7 @@ final class PanelSheetState: ObservableObject {
         // Only restore if no sheet is already active — prevents overwriting a concurrently set runner.
         guard editingRunner == nil, let runnerSheetSnapshot else { return }
         editingRunner = runnerSheetSnapshot
+        self.runnerSheetSnapshot = nil
     }
 
     /// Clears all runner sheet state for explicit close/reset paths.
--- a/Sources/RunnerBar/App/PanelVisibilityState.swift
+++ b/Sources/RunnerBar/App/PanelVisibilityState.swift
@@ -53,6 +53,8 @@ import SwiftUI
 // ════════════════════════════════════════════════════════════════════════════════
 
 /// Observable wrapper for NSPanel open/closed state + one-shot height callback.
+// Note: mutated exclusively on the main thread by AppDelegate — no @MainActor annotation
+// because PanelContainerView reads it via @EnvironmentObject which handles dispatch.
 final class PanelVisibilityState: ObservableObject {
     /// `true` from immediately before the panel opens until after it closes.
     @Published var isOpen: Bool = false
--- a/Sources/RunnerBar/App/PanelVisibilityState.swift
+++ b/Sources/RunnerBar/App/PanelVisibilityState.swift
@@ -53,9 +53,9 @@ import SwiftUI
 // ════════════════════════════════════════════════════════════════════════════════
 
 /// Observable wrapper for NSPanel open/closed state + one-shot height callback.
-///
-/// Mutated exclusively on the main thread by AppDelegate — no `@MainActor` annotation
-/// because `PanelContainerView` reads it via `@EnvironmentObject` which handles dispatch.
+/// - Note: Mutated exclusively on the main thread by AppDelegate — no `@MainActor`
+///   annotation is used because all call sites (AppDelegate + SwiftUI view callbacks)
+///   already run on the main thread.
 final class PanelVisibilityState: ObservableObject {
     /// `true` from immediately before the panel opens until after it closes.
     @Published var isOpen: Bool = false
--- a/Sources/RunnerBar/App/PanelVisibilityState.swift
+++ b/Sources/RunnerBar/App/PanelVisibilityState.swift
@@ -53,8 +53,9 @@ import SwiftUI
 // ════════════════════════════════════════════════════════════════════════════════
 
 /// Observable wrapper for NSPanel open/closed state + one-shot height callback.
-// Note: mutated exclusively on the main thread by AppDelegate — no @MainActor annotation
-// because PanelContainerView reads it via @EnvironmentObject which handles dispatch.
+///
+/// Mutated exclusively on the main thread by AppDelegate — no `@MainActor` annotation
+/// because `PanelContainerView` reads it via `@EnvironmentObject` which handles dispatch.
 final class PanelVisibilityState: ObservableObject {
     /// `true` from immediately before the panel opens until after it closes.
     @Published var isOpen: Bool = false


eonist:
## Deep Review — Batch 16: App State Doc & Comment Cleanup

Reviewed against issue #1097 spec, the diff, and the broader cleanup project (#1038). Overall the intent is solid, but there are a few things that go beyond cosmetics and deserve attention.

---

### ✅ `NavState.swift` — Mostly right, one concern

**What changed:** The file-level `//` block was trimmed to history-only (good), and `case stepLog` got proper labeled associated values and `- Parameters:` doc.

**The associated-value labeling is a breaking-ish change.** Adding labels to `case stepLog(job: ActiveJob, step: JobStep)` changes the Swift-generated member name from `.stepLog(_:_:)` to `.stepLog(job:step:)`. This means any `switch` site pattern-matching without labels would still compile (Swift is lenient here), but any call site constructing the case _must_ now use labels. The fix in `AppDelegate+Navigation.swift` correctly updates:

```swift
// Before
self.savedNavState = .stepLog(job, step)
// After
self.savedNavState = .stepLog(job: job, step: step)
```

✅ That site is updated. But: is this the **only** place `NavState.stepLog` is constructed? A `switch` over `NavState` elsewhere (e.g. in any routing or restore logic) might now need label-aware pattern matching depending on Swift version. Worth a grep: `grep -r "\.stepLog(" --include="*.swift"` to confirm no site was missed. This PR only touches one file in `App/`, so if there's a `navigate(to:)` or state restoration path elsewhere, it's easy to miss.

**History block quality:** The history block now reads clearly and the doc comment carries the intent prose. This is the correct separation. ✅

---

### ⚠️ `PanelSheetState.swift` — Behaviour change buried in a doc PR

This is the most important thing in the entire PR and the description undersells it.

**This is not just a doc change:**

```swift
func restoreTransientHideStateIfNeeded() {
+   // Only restore if no sheet is already active — prevents overwriting a concurrently set runner.
    guard editingRunner == nil, let runnerSheetSnapshot else { return }
    editingRunner = runnerSheetSnapshot
+   self.runnerSheetSnapshot = nil   // ← NEW LINE, not doc
}
```

`self.runnerSheetSnapshot = nil` is a **functional fix**, not a comment. Before this commit, `runnerSheetSnapshot` was captured but never cleared after restore, meaning:
- A second call to `restoreTransientHideStateIfNeeded()` (e.g. popover hide → show → hide → show) would restore the _stale_ snapshot on the second cycle even if the user had explicitly dismissed the sheet in between.
- `clearRunnerSheetState()` presumably clears `editingRunner` but if it doesn't also nil `runnerSheetSnapshot`, the state machine has a latent double-restore bug.

This fix is **correct and necessary**, but it belongs in its own commit/PR with a test note, not silently embedded in a doc batch. The issue #1097 spec didn't call for this change either — the spec only asked for an inline comment explaining the guard semantics. The nil-assignment is additive functional scope.

**Recommendation:** Either:
1. Call this out explicitly in the PR description as a behaviour fix (not just a doc change), or
2. Move `self.runnerSheetSnapshot = nil` to a separate commit so the history is clean.

**The `import Foundation` removal** is also unmentioned. Removing it is correct if nothing in this file uses Foundation APIs directly (Combine and RunnerBarCore cover the needed types), but it should be noted explicitly. If some future change re-adds a Foundation type, the removal is invisible in blame.

---

### ✅ `PanelVisibilityState.swift` — Good, with a nuance

The threading note is well placed and accurate:

```swift
/// Mutated exclusively on the main thread by AppDelegate — no `@MainActor` annotation
/// because `PanelContainerView` reads it via `@EnvironmentObject` which handles dispatch.
```

**One nuance worth flagging for future-proofing:** The claim that `@EnvironmentObject` "handles dispatch" is partially true — SwiftUI's property wrappers do publish changes and update views on the main thread for `@Published` properties, but they don't retroactively make the _mutation site_ actor-safe. Under Swift 6 strict concurrency, `AppDelegate` (likely `@MainActor` annotated or running on main via `NSApplicationDelegate`) calling into a non-`@MainActor` class will produce a warning/error if the class isn't isolated. The comment is correct as a _current_ explanation, but it subtly implies a safety guarantee that doesn't exist under strict concurrency.

If Swift 6 strict concurrency is enabled in a future sprint, `PanelVisibilityState` will need either `@MainActor` on the class or explicit `nonisolated` with `MainActor.assumeIsolated` at call sites. Worth a `// TODO(swift6-concurrency): revisit @MainActor` note if this codebase is moving that direction.

---

### ✅ `AppDelegate+Navigation.swift`

The only change is adding labels to the `stepLog` construction site. Correct, necessary, and clean. ✅

---

### `main.swift` / `Exports.swift` — Confirmed clean ✅

---

### Summary

| File | Verdict | Key Note |
|---|---|---|
| `NavState.swift` | ✅ Approve | Verify no other `.stepLog(` construction site was missed |
| `PanelSheetState.swift` | ⚠️ Needs note | `runnerSheetSnapshot = nil` is a behaviour fix, not doc — should be surfaced in description |
| `PanelVisibilityState.swift` | ✅ Approve with note | Threading comment accurate today, but weak under Swift 6 strict concurrency |
| `AppDelegate+Navigation.swift` | ✅ Approve | Label update is correct and consistent |

The PR is mergeable as-is (the functional fix is correct), but the description should be updated to explicitly call out the `runnerSheetSnapshot = nil` behaviour change so it doesn't get lost in blame history.

eonist:
## Deep Review — Batch 16 App State Cleanup

### Overall

Solid batch. All five items from Issue #1097 are addressed and the checklist maps cleanly to what's in the diff. This is a doc/comment-only PR in spirit, but there's one **functional change** buried in `PanelSheetState.swift` that deserves specific attention — it's the right thing to do, but it shouldn't be invisible.

---

### `NavState.swift`

**✅ Duplicate comment block correctly trimmed.**
The before-state had the full prose ("Represents the currently visible navigation screen... Extracted from AppDelegate.swift #602...") duplicated verbatim in both the file-level `//` block and the `///` doc comment. The fix moves the origin story into the `///` doc where it belongs and leaves the `//` block as pure history (`#455`, `#992`, `#1001`). This is exactly what issue #1097 asked for and the signal/noise ratio is now clearly better.

**✅ Associated value labels added to `stepLog`.**
Adding `job:` and `step:` labels to `case stepLog(job: ActiveJob, step: JobStep)` is a meaningful improvement — both for documentation *and* for call-site readability. The label addition is consistent and already applied to the single call site in `AppDelegate+Navigation.swift` (`.stepLog(job: job, step: step)`). Good catch making that a coordinated change — a label-only enum change without updating the call site would have broken compilation.

**One observation worth flagging:** The `///` doc on `case stepLog` now has a `- Parameters:` block, but `case main` and `case settings` have no associated values and just a single-line doc. That's fine. However, the case-level docs are asymmetric: `main` and `settings` use `/// The root popover...` / `/// The Settings sheet.` style, while `stepLog` now has a multi-line block. This is perfectly valid Swift doc comment style, just noting it's intentionally asymmetric and not a problem.

---

### `PanelSheetState.swift`

**✅ `runnerSheetSnapshot` doc improved.**
Old: `/// Snapshot captured immediately before a transient hide.`
New: `/// Backing store for captureTransientHideState() — persists sheet intent across NSPopover hide/show cycles. See type doc for NSPopover teardown context.`

The new version correctly surfaces the *why* (NSPopover teardown) and the *who* (`captureTransientHideState()`), which is exactly what was missing. The cross-reference to the type doc is a good documentation pattern — it avoids repeating the NSPopover explanation inline while ensuring discoverability.

**✅ Guard semantics comment added in `restoreTransientHideStateIfNeeded()`.**
The comment `// Only restore if no sheet is already active — prevents overwriting a concurrently set runner.` correctly explains the subtle contract of the `guard editingRunner == nil` condition.

**🔴 Functional change — `runnerSheetSnapshot = nil` after restore:**
```swift
func restoreTransientHideStateIfNeeded() {
    guard editingRunner == nil, let runnerSheetSnapshot else { return }
    editingRunner = runnerSheetSnapshot
+   self.runnerSheetSnapshot = nil   // <-- NEW, functional
}
```
This line was **not in the issue spec** and is a **real behavioural change**, not a doc change. Before this PR, `runnerSheetSnapshot` was never cleared after a successful restore — meaning if `restoreTransientHideStateIfNeeded()` were called a second time in the same show-cycle, it would attempt to re-restore an already-active `editingRunner` (though the `guard editingRunner == nil` would block it). After this PR, the snapshot is consumed on first restore, so subsequent calls are cleanly idempotent regardless of whether `editingRunner` has since been cleared.

This is almost certainly the **correct fix** — a snapshot that survives past its consumption is a bug waiting to happen. But it should be:
1. Called out explicitly in the PR description as a functional fix, not just a doc cleanup.
2. Ideally tracked against a specific issue or at minimum a `// - fix: consume snapshot after restore to prevent stale re-use` comment so future `git blame` readers understand the intent.

The change is safe given the current call pattern, but its silent inclusion in a "doc cleanup" batch makes it easy to miss in review.

---

### `PanelVisibilityState.swift`

**✅ Threading note added.**
```swift
/// Observable wrapper for NSPanel open/closed state + one-shot height callback.
///
/// Mutated exclusively on the main thread by AppDelegate — no `@MainActor` annotation
/// because `PanelContainerView` reads it via `@EnvironmentObject` which handles dispatch.
final class PanelVisibilityState: ObservableObject {
```

The note is accurate and the reasoning is sound as of today. One thing to flag for the future: the claim that "`@EnvironmentObject` handles dispatch" is a simplification. What's actually happening is that SwiftUI's `@EnvironmentObject` property wrapper triggers `objectWillChange` publishing, which is observed on the main run loop because `ObservableObject`'s default `objectWillChange` publisher is a `PassthroughSubject` and SwiftUI subscribes to it on MainActor. This holds true but if this class ever gets mutations from a background context (e.g., a callback from a networking layer), the "no annotation needed" justification breaks down. Not a problem now, but the comment could be slightly more precise: `"...no @MainActor annotation is currently required — caller (AppDelegate) guarantees main-thread mutation."` to make the invariant caller-owned rather than SwiftUI-owned.

**✅ `import Foundation` removed from `PanelSheetState.swift`.**
Clean-up of the unused import. Low noise.

---

### `main.swift` / `Exports.swift`

Confirmed clean as documented. No diff, no issue.

---

### Summary

| File | Issue spec met? | Notes |
|---|---|---|
| `main.swift` | ✅ | No-change tombstone, correct |
| `Exports.swift` | ✅ | No-change tombstone, correct |
| `NavState.swift` | ✅ | Comment trim + label addition well-executed |
| `PanelSheetState.swift` | ✅ + ⚠️ | Doc improvements correct; functional `= nil` change is right but undocumented in PR |
| `PanelVisibilityState.swift` | ✅ | Threading note accurate; minor wording suggestion for future precision |

**One blocking observation:** The `runnerSheetSnapshot = nil` line is a behavioural change shipped inside a doc-only batch without being called out. It's not wrong — it's arguably a latent bug fix — but it should be explicitly noted in the PR description or in the commit message so it's visible to future readers and doesn't get lost in the "doc cleanup" framing.

Everything else is clean. Good batch.

eonist:
## Deep Review — Batch 16 App State Cleanup

Overall this is a solid, focused batch. The diff is small but non-trivial — there's one functional change hiding among the doc/comment work, and one comment that could mislead. Notes below, ordered by file.

---

### `NavState.swift`

**✅ Comment restructuring is correct.**
Moving the extraction history out of the plain `//` block and into the `///` doc comment is exactly the right call. The `// History:` header now carries signal (the PR/issue numbers) that the doc comment can't, and the doc comment cleanly describes the *what*. No duplication, no noise.

**✅ Associated value labels are a real improvement.**
`case stepLog(ActiveJob, JobStep)` with no labels was genuinely ambiguous at the call site — you'd have to read the implementation to know which was which. The new labels + parameter docs make this self-evident.

**⚠️ The call-site change in `AppDelegate+Navigation.swift` is a functional/API change, not just docs.**
Line: `self.savedNavState = .stepLog(job: job, step: step)`
Adding explicit labels to associated values in Swift *is* a source-breaking API change if `NavState` is ever used outside this module — matching on `.stepLog(let job, let step)` without labels will still compile, but constructing it with positional arguments (`.stepLog(job, step)`) will now be a compile error. Since `NavState` appears to be internal to `RunnerBar` and `RunnerBarCore` is imported, this is likely fine, but it's worth noting this is the one change in the batch that isn't pure doc/comment work. It should arguably be called out explicitly in the PR description under "Changes" — right now the description says "add associated value docs" but doesn't mention the corresponding call-site fix.

---

### `PanelSheetState.swift`

**✅ `runnerSheetSnapshot` doc improvement is exactly right.**
The old doc ("Snapshot captured immediately before a transient hide") described *when* it's captured but not *why* it needs to exist. The new doc cross-references `captureTransientHideState()` and the NSPopover teardown context. Good.

**✅ Guard comment is accurate and useful.**
"Prevents overwriting a concurrently set runner" is precisely the subtle race condition the guard protects against. Valuable for future maintainers.

**✅ `self.runnerSheetSnapshot = nil` — this is the most substantive functional change in the whole batch.**
After restoring the snapshot, the previous code left `runnerSheetSnapshot` populated. This means a second call to `restoreTransientHideStateIfNeeded()` would hit the `guard editingRunner == nil` branch — but only if `editingRunner` happened to be nil again by then (e.g. the sheet was dismissed before a second popover hide/show cycle). Nil-ing after restore is the correct fix. However: this is a **behavioural change**, not a doc change. It should be called out clearly in the PR description. If this batch is being reviewed as "doc & comment cleanup", a reviewer skimming the summary could miss this entirely. Suggest adding a note like: *"PanelSheetState: also fixes a latent bug — runnerSheetSnapshot was not cleared after restore, leaving it available for a spurious second restore."*

**🟡 `import Foundation` removal.**
Dropping the `import Foundation` is fine if nothing in the file uses Foundation APIs directly (Combine's `ObservableObject`, `@Published`, etc. pull in Foundation transitively via RunnerBarCore or Combine itself on Apple platforms). Just worth being explicit that this was intentional and not accidental during editing — the PR description doesn't mention it.

---

### `PanelVisibilityState.swift`

**✅ Threading note is accurate and well-placed.**
The note correctly explains the deliberate absence of `@MainActor`. The `@EnvironmentObject` point is valid — SwiftUI's property wrappers coalesce `@Published` updates onto the main thread for view rendering. This is a good pre-emptive annotation for Swift 6 strict concurrency mode.

**🟡 Minor precision nit.**
The comment says `@EnvironmentObject` "handles dispatch" — this is slightly loose. What actually happens is that `@Published` updates drive `objectWillChange`, and SwiftUI's view update machinery ensures those coalesce on the main thread. `@EnvironmentObject` itself doesn't dispatch; it just exposes the object. This is a micro-nit and won't confuse an experienced Swift dev, but for a comment whose entire purpose is precision about threading, it could be tightened to something like: *"no `@MainActor` annotation because SwiftUI coalesces `@Published` view updates onto the main thread."*

---

### `AppDelegate+Navigation.swift`

No issues other than the label change noted above under `NavState.swift`.

---

### Summary

| File | Type | Assessment |
|---|---|---|
| `NavState.swift` | Doc restructure + label change | ✅ Good, but label change = source-breaking API change, worth calling out |
| `AppDelegate+Navigation.swift` | Call-site fix from label change | ✅ Correct, but unlisted in PR description |
| `PanelSheetState.swift` | Doc + functional bug fix | ✅ Fix is correct — but `runnerSheetSnapshot = nil` is a behavioural change buried in a doc PR |
| `PanelSheetState.swift` | `import Foundation` removed | ✅ Fine, but unmentioned |
| `PanelVisibilityState.swift` | Threading doc | ✅ Accurate, minor precision nit on `@EnvironmentObject` |

**Main ask before merging:** Update the PR description to explicitly surface the two behavioural changes (label-breaking call-site update, `runnerSheetSnapshot = nil` fix). Batching functional fixes into a doc PR isn't wrong — but the changelog should reflect what actually changed so future git bisect / blame is useful.

Everything else looks good. Happy to approve once the description is updated.

Issue #1100, opened 2026-06-03
Batch 18 — GitHub auth layer: OAuth correctness fixes + transport cleanup (Auth, Keychain, OAuthService, Secrets, Scope, GitHub, GitHubURLSessionTransport)


eonist:
Tracking issue for Batch 18 of the ongoing cleanup project (#1038).

## Files in scope

- `Sources/RunnerBar/GitHub/Auth.swift`
- `Sources/RunnerBar/GitHub/GitHubAppSecrets.swift`
- `Sources/RunnerBar/GitHub/GitHubURLSessionTransport.swift`
- `Sources/RunnerBar/GitHub/Keychain.swift`
- `Sources/RunnerBar/GitHub/OAuthService.swift`
- `Sources/RunnerBar/GitHub/Scope.swift`
- `Sources/RunnerBar/GitHub/Secrets.swift`
- `Sources/RunnerBar/GitHub/GitHub.swift`

---

### 1. `OAuthService.signIn()` — force-unwrap crash path

`URLComponents(string: authorizeURL)!` force-unwraps a string interpolation. If `GitHubConstants.base` ever produces a malformed URL, the app crashes at sign-in. Fix:

```swift
// Before
var comps = URLComponents(string: authorizeURL)!

// After
guard var comps = URLComponents(string: authorizeURL) else {
    log("OAuthService › signIn: malformed authorizeURL — aborting")
    return
}
```

### 2. `OAuthService.exchangeCode()` — GitHub error body not detected

GitHub's token exchange endpoint always returns `200 OK`, even on failure — error cases return `{"error": "bad_verification_code", ...}` instead of `access_token`. The current code collapses this into a silent `onCompletion?(false)` with no log. A user reusing an expired code gets no feedback. Fix: check for `json["error"]` before checking for `access_token`:

```swift
if let errorCode = json["error"] as? String {
    let desc = json["error_description"] as? String ?? ""
    log("OAuthService › exchangeCode: GitHub error=\(errorCode) \(desc)")
    onCompletion?(false)
    return
}
```

### 3. `OAuthService.handleCallback()` — nil state conflated with CSRF mismatch

The current guard conflates two different failure modes into one log message:

```swift
guard let returnedState, returnedState == pendingState else {
    log("... state mismatch — possible CSRF attempt, rejecting")
    ...
}
```

[...]

Issue #777, opened 2026-05-23
perf: migrate urlSessionAPI / urlSessionAPIPaginated from DispatchSemaphore to async/await


eonist:
## Problem

`urlSessionAPI` and `urlSessionAPIPaginated` block GCD threads with `DispatchSemaphore` + `URLSession.dataTask` for the full network round-trip duration. With multiple scopes polling concurrently this can starve the GCD thread pool.

## Fix

Migrate to `async/await` with `URLSession.data(for:)`:

```swift
let (data, response) = try await URLSession.shared.data(for: request)
```

This frees threads while waiting and is the idiomatic Swift concurrency approach.

## Severity
🟡 Medium — performance issue under concurrent polling load.

PR #1103, opened 2026-06-03
Batch 18 — GitHub auth layer: OAuth correctness fixes + transport cleanup

eonist:
## **User description**
Tracking issue: #1100

## Summary

Batch 18 of the ongoing cleanup project (#1038). Covers the GitHub auth layer — OAuth correctness fixes, transport cleanup, and dead code removal.

### 🔴 Critical OAuth fixes

- **`OAuthService.signIn()`** — replaced `URLComponents(string: authorizeURL)!` force-unwrap with `guard`+log. Prevents a crash if `GitHubConstants.base` ever produces a malformed URL.
- **`OAuthService.exchangeCode()`** — added `json["error"]` detection before checking for `access_token`. GitHub's token exchange endpoint always returns `200 OK` on failure; without this check, an expired code silently calls `onCompletion?(false)` with no log.
- **`OAuthService.handleCallback()`** — split the single `guard let returnedState, returnedState == pendingState` into two separate guards: one for a missing `state` param, one for an actual CSRF mismatch. Eliminates the misleading "possible CSRF attempt" log when the redirect URL simply lacks a `state` param.

### 🟡 Code improvements

- **`Auth.swift`** — removed stale `gh auth token` priority #3 from the doc comment (implementation was removed but comment wasn't updated). Renumbered to 1–4. Updated cache description to reflect actual behaviour.
- **`Keychain.save()`** — added `kSecAttrAccessibleAfterFirstUnlock` to the `SecItemUpdate` attributes dict (previously only set on the `SecItemAdd` path). Ensures legacy items created without this attribute are upgraded in place; without it, a legacy item may be inaccessible at launch before the first unlock.
- **`GitHubURLSessionTransport`** — removed redundant `githubToken() != nil` guard in `ghAPI()` and `ghAPIPaginated()`. The inner `urlSessionAPI`/`urlSessionAPIPaginated` functions already guard on `githubToken()`, making the outer check a no-op that adds a second lock acquisition per call.
- **`GitHub.swift`** — migrated `fetchStepLog` to use `urlSessionRaw()` from the transport layer. Deleted `NoRedirectDelegate`, `noRedirectDelegate`, `noRedirectSession`, and `fetchStepLogViaURLSession()` (~80 lines). `urlSessionRaw()` already handles the GitHub 302→S3 redirect pattern correctly.
- **`GitHub.fetchActiveJobs`** — added explanatory comment on the `swiftlint:disable:next for_where` suppression: `filter()` cannot replace the loop because `insert()` mutates `seenRunIDs` as a side effect.
- **Deleted `GitHubAppSecrets.swift`** — exact duplicate of `Secrets.swift`, already `periphery:ignore`'d with zero references.

### Critical OAuth fixes
- [x] `OAuthService.signIn()`: replace `URLComponents(string:)!` force-unwrap with `guard`
- [x] `OAuthService.exchangeCode()`: detect `json["error"]` from GitHub token exchange response
- [x] `OAuthService.handleCallback()`: split nil-state vs CSRF-mismatch into separate guards

### Code improvements
- [x] Delete `GitHubAppSecrets.swift` (duplicate of `Secrets.swift`)
- [x] `Auth.swift`: remove stale `gh auth token` priority #3 from doc, renumber to 1–4
- [x] `Keychain.save()`: add `kSecAttrAccessibleAfterFirstUnlock` to `SecItemUpdate` attributes
- [x] `GitHubURLSessionTransport`: remove redundant `githubToken() != nil` guard in `ghAPI()` and `ghAPIPaginated()`
- [x] `GitHub.swift`: migrate `fetchStepLog` to `urlSessionRaw()`, delete `NoRedirectDelegate` + `fetchStepLogViaURLSession()`
- [x] `GitHub.fetchActiveJobs`: add explanatory comment on `swiftlint:disable for_where`

### Note
- `OAuthService.signOut()` — already gated on `deleted == true` in the existing code; no change needed.
- `Scope.swift` — clean, no changes needed.

___

## **CodeAnt-AI Description**
**Fix GitHub sign-in failures and make step log loading more reliable**

### What Changed
- GitHub sign-in no longer crashes if the authorization URL is malformed
- Redirect handling now treats a missing state value differently from a real CSRF mismatch, so the failure reason is logged correctly
- Failed GitHub token exchanges now surface the returned error instead of failing silently
- Saved GitHub tokens are upgraded so they remain available after reboot and before the first device unlock
- Step logs now load through the raw log path directly, removing the manual redirect flow that could fail mid-fetch

### Impact
`✅ Fewer GitHub sign-in crashes`
`✅ Clearer sign-in failure messages`
`✅ More reliable GitHub login after restart`
`✅ Fewer step log loading failures`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


eonist:
## Deep Review — Batch 18: GitHub Auth Layer

Overall verdict: **✅ Approve with notes.** The critical OAuth fixes are correct and well-reasoned. The `fetchStepLog` migration is the standout win. There are a few things worth flagging though — one of which I'd call a real gap vs. the tracking issue.

---

### 🔴 Issue #5 from tracking issue was silently dropped

The tracking issue (#1100) lists **Item 5 — `OAuthService.signOut()`: gate `didSignOut` on `deleted == true`, or document optimistic design** as a checklist item. The PR description marks it as "already gated on `deleted == true` in the existing code; no change needed," but the issue text specifically calls out the opposite: `didSignOut.send()` fires _unconditionally_ after `Keychain.delete()`.

I haven't reviewed `OAuthService.signOut()` directly in this diff since it's untouched, but the tracking issue very explicitly describes this as a bug: if `Keychain.delete()` fails, the UI shows signed-out but the old token remains in Keychain, causing "ghost sign-in" on next launch. **This should not be silently closed on the PR description without a code reference showing `didSignOut` is already gated.** Either confirm with a snippet from the current implementation or reopen this item.

---

### 🟠 `OAuthService.signIn()` — `pendingState` leak on `comps.url` failure

In the new code:

```swift
guard var comps = URLComponents(string: authorizeURL) else {
    log(...)
    pendingState = nil
    onCompletion?(false)
    return
}
// ...
guard let url = comps.url else {
    log(...)
    pendingState = nil
    onCompletion?(false)
    return
}
```

Both guards correctly nil out `pendingState`. Good. However: if `NSWorkspace.shared.open(url)` somehow throws or the open fails silently, `pendingState` is left set with no callback ever arriving. This was pre-existing behaviour and is not a regression, but since this function is being hardened, it's worth a `// NOTE: pendingState is left set if the browser open fails — handleCallback will CSRF-reject any eventual redirect` comment so future readers understand the lifecycle.

---

### 🟡 `Keychain.save()` — retry path update is correct but the comment placement is slightly misleading

The explanatory comment is added above the _first_ `SecItemUpdate` call, but the same `kSecAttrAccessibleAfterFirstUnlock` fix is also applied in the retry path (~27 lines later). The comment does not mention the retry path. A reader scanning fast might miss that the retry path was also patched and assume it's still broken. Suggest either:

```swift
// kSecAttrAccessibleAfterFirstUnlock is included on both the initial update
// and the retry path below for the same reason: ...
```

Or move the comment to a named constant that's referenced in both places:

```swift
private let keychainUpdateAttributes: [String: Any] = [
    kSecValueData as String: data,  // not possible as a stored constant, but a helper dict would work
    kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlock
]
```

Minor, but worth noting given this is a Keychain accessibility bug fix — the reasoning should be airtight for a future auditor.

---

### 🟡 `ghAPI` / `ghAPIPaginated` — log loss is a real tradeoff, not just cleanup

The PR frames removing the redundant `githubToken() != nil` guards as a pure simplification (avoid double lock acquisition). That's true — but those guards also produced **early, named log lines** like `"ghAPI › no token available for: \(endpoint)"`. The inner `urlSessionAPI` will still return `nil`, but the log message it emits (if any) is at a lower level of abstraction and may not include the endpoint name in the same format.

This is a fine tradeoff, but the PR description should explicitly state: "the log line is intentionally removed — `urlSessionAPI` handles the nil token path." If a future engineer debugging auth sees no `ghAPI` log for a failing request, they shouldn't have to git-blame to understand why.

---

### 🟢 `fetchStepLog` migration — well executed

Deleting `NoRedirectDelegate`, `noRedirectSession`, `noRedirectDelegate`, and `fetchStepLogViaURLSession` (the two-semaphore manual redirect chain, ~80 lines) in favour of `urlSessionRaw()` is correct and clean. The new error logging is strictly better — separate messages for `nil` from transport, UTF-8 decode failure (with byte count), and empty body. The `data.count` in the decode-failure log is a nice touch for debugging binary/non-UTF-8 log responses.

One small note: the new doc comment says `urlSessionRaw uses application/vnd.github.v3.raw and lets URLSession follow the GitHub 302→S3 redirect automatically`. Confirm that `urlSessionRaw()` does **not** set a custom redirect handler or `URLSessionConfiguration` that blocks redirects — if it does, this silently breaks again. From the transport source this appears to be the case (default session), but worth a line in the comment like: `// urlSessionRaw() uses URLSession.shared (follows redirects); no custom delegate needed`.

---

### 🟢 `OAuthService.exchangeCode()` — correct and complete

Splitting the `guard` so `json` is decoded unconditionally, then checking `json["error"]` before `access_token`, matches GitHub's documented token exchange behaviour exactly. The `error_description` capture is good UX for log-based debugging (`bad_verification_code` vs `incorrect_client_credentials` are very different failure modes). No issues.

---

### 🟢 `handleCallback()` state split — correct

Splitting the compound guard into two is the right call. The original code logged a CSRF attempt for a missing `state` param (e.g. a test redirect, or a broken OAuth provider config), which was noisy and misleading. The new logs are accurate. `pendingState = nil` on both paths is correct cleanup.

---

### 🟢 `GitHubAppSecrets.swift` deletion — correct

It was `periphery:ignore`'d with zero references, identical to `Secrets.swift`. Clean delete.

---

### 🟢 `Auth.swift` doc update — correct

Removing the stale `gh auth token` priority #3 and renumbering to 1–4 is accurate. The cache description update from "shell spawns" to "Keychain reads" is consistent with the current implementation.

---

### Summary table

| Item | Status | Notes |
|------|--------|-------|
| `signIn()` force-unwrap fix | ✅ | Correct; minor lifecycle comment suggested |
| `exchangeCode()` error detection | ✅ | Complete and accurate |
| `handleCallback()` state split | ✅ | Clean |
| `GitHubAppSecrets.swift` delete | ✅ | |
| `signOut()` `didSignOut` gate | ⚠️ | **Not addressed — tracking issue item #5 not resolved or verified** |
| `Auth.swift` doc cleanup | ✅ | |
| `Keychain.save()` update path fix | ✅ | Comment scope could be broader |
| `ghAPI`/`ghAPIPaginated` simplification | ✅ | Log loss tradeoff undocumented |
| `fetchStepLog` → `urlSessionRaw()` | ✅ | Best change in this batch |
| `fetchActiveJobs` SwiftLint comment | ✅ | |

eonist:
## Deep Review — Batch 18: GitHub Auth Layer

Overall this is a high-quality, well-scoped batch. The issue spec was thorough and the implementation closely follows it. Below I go file-by-file against both the diff and the original issue checklist, noting what's correct, what's subtle, and a few things worth discussing.

---

### ✅ `OAuthService.signIn()` — Force-unwrap removal

The fix is correct and the recovery path is sound. Two details worth noting:

1. The `pendingState = nil` cleanup on the first `guard` failure (malformed `authorizeURL`) is the right call — you don't want a dangling pending state if you never actually opened the browser.
2. The comment left above `NSWorkspace.shared.open(url)` — _"pendingState is left set if the browser open fails silently"_ — is an honest and useful note. One minor thing: `open(_:)` is technically synchronous and returns a `Bool`. If the browser launch fails, you currently don't detect it. Low-probability on macOS but worth knowing. Not a blocker, just a future hardening point if you add a retry/error UI.

---

### ✅ `OAuthService.exchangeCode()` — GitHub 200-on-error detection

The issue identified the exact root cause correctly: GitHub's token endpoint returns `200 OK` with `{"error": "bad_verification_code"}` on failure. The fix — checking `json["error"]` before `json["access_token"]` — is exactly right and matches GitHub's documented contract.

One subtle note from the diff: you now parse `json` outside the original `guard let ... token =` chain by splitting the guard into two steps (first get `json`, then check for error, then extract `token`). This is cleaner, but the silent `onCompletion?(false); return` at the end of the `guard let token` block has no log message. If `json` comes back with neither `"error"` nor `"access_token"` (malformed/unexpected response), you get a silent failure with zero diagnostics. A one-liner log here would complete the error coverage:

```swift
guard let token = json["access_token"] as? String, !token.isEmpty else {
    log("OAuthService › exchangeCode: no access_token in response — \(json.keys.sorted())")
    onCompletion?(false)
    return
}
```

---

### ✅ `OAuthService.handleCallback()` — nil-state vs CSRF split

This is a correctness fix, not just a cosmetic one. The original single guard was logically wrong: when `returnedState` is nil (no `state` param at all), it's not a CSRF attempt — it's a malformed redirect. The split guards now distinguish:
- Missing param → protocol/integration issue
- Param present but mismatched → genuine CSRF signal

Both guards now also clear `pendingState = nil` before returning, which is correct. In the original code the `guard let returnedState, returnedState == pendingState` failure path also cleared `pendingState`, but a `nil`-state failure didn't — a subtle state-leak that this fix resolves.

---

### ✅ `OAuthService.signOut()` — `didSignOut` now gated on Keychain success

The issue raised option A (gate on `deleted == true`) or option B (document optimistic design). You chose A, which is the safer choice. The comment added — _"sign-out suppressed to prevent ghost sign-in"_ — is exactly the right framing to explain why this isn't just pessimism.

One edge case to be aware of: if `Keychain.delete()` fails and `didSignOut` is suppressed, the UI stays in a "signed in" state. The user sees no feedback. If the Keychain delete fails due to a transient OS error (not just "item not found"), the user has no way to sign out without restarting. This is probably acceptable for now, but a future improvement would be surfacing that failure state to the UI rather than silently swallowing it.

---

### ✅ `Auth.swift` — Stale `gh auth token` comment removal

Clean doc fix. The renumbering from 1–5 to 1–4 is correct and the updated cache description ("avoids repeated Keychain reads") accurately reflects what the code actually does. Straightforward.

---

### ✅ `Keychain.save()` — `kSecAttrAccessibleAfterFirstUnlock` on update path

Correct fix for a real-world scenario (app update / different signing identity leaves a legacy item without the `AfterFirstUnlock` attribute, then the app tries to read it at launch before the first unlock and gets `errSecInteractionNotAllowed`).

One thing to double-check: `kSecAttrAccessible` changes in a `SecItemUpdate` are processed differently by the Keychain depending on the macOS version. On macOS 10.15+ it works as expected. On older versions it was historically not guaranteed that you could change `kSecAttrAccessible` via update alone without a delete+add cycle. Since RunnerBar targets modern macOS this is almost certainly fine, but worth a mental note for the changelog.

Also — the retry `SecItemUpdate` at line 83 (inside the `errSecItemNotFound → add → retry` path) now correctly includes `kSecAttrAccessibleAfterFirstUnlock` too. Both update paths are covered. ✅

---

### ✅ `GitHubURLSessionTransport` — Redundant `githubToken() != nil` guard removal

The issue analysis was correct: `ghAPI()` called `githubToken()` (one lock), then `urlSessionAPI()` called it again internally (second lock). Since `urlSessionAPI` already handles nil correctly, the outer guard was a pure no-op that added latency and log noise on failures.

The removal is safe. One observation: by removing the outer guard you also removed the `log("ghAPI › no token available for: \(endpoint)")` messages. The inner `urlSessionAPI` has its own no-token log, so coverage is maintained — just at a lower level in the call stack. This is fine; the inner log is actually more informative since it's in `urlSessionAPI` which has full context.

---

### ✅ `GitHub.swift` — `fetchStepLog` migration to `urlSessionRaw()`

This is the biggest change in the batch (~80 lines deleted) and the one most worth scrutinising.

**The redirect safety argument is sound.** The new doc comment in `makeRawRequest` and `urlSessionRaw` correctly explains that Apple's URLSession strips the `Authorization` header on cross-origin redirects (github.com → amazonaws.com). This is documented Apple behaviour per RFC 7235 and the `URLSession` delegate docs. The `NoRedirectDelegate` pattern was solving a problem that doesn't exist on Apple's networking stack.

**However, one functional difference is worth calling out explicitly:** The old `fetchStepLogViaURLSession` sent `Accept: application/vnd.github+json` on step 1 and `Accept: text/plain` on step 2 (to S3). The new `urlSessionRaw` uses `Accept: application/vnd.github.v3.raw` throughout. These are not identical:

- `application/vnd.github.v3.raw` on the initial request to GitHub tells GitHub to return raw content — for the log endpoint (`/actions/jobs/{id}/logs`) GitHub responds with the 302, so the accept header on the initial leg is effectively irrelevant.
- On the S3 follow-up, URLSession's redirect will replay the request to S3, but Apple strips only auth headers, not `Accept`. S3 will receive `Accept: application/vnd.github.v3.raw` instead of `Accept: text/plain`. S3 ignores `Accept` entirely and serves the content regardless, so this makes zero practical difference — but it's worth knowing the header isn't perfectly clean on the S3 leg.

**The real simplification win** is eliminating two `DispatchSemaphore`s and the manual `OSAllocatedUnfairLock<URL?>` dance. The old code was essentially implementing async/await manually with semaphores, while `urlSessionRaw` wraps the same URLSession call with a single semaphore internally. The net behaviour is equivalent with half the code surface.

The TODO comment `// TODO: migrate sem1/sem2 to async/await as part of #777` is also gone — that technical debt is now moot since the whole implementation was replaced.

---

### ✅ `GitHubAppSecrets.swift` — Deleted

Correct. The file was an exact duplicate of `Secrets.swift`, already marked `periphery:ignore`, with zero references. The OAuth credential security disclaimer comment (explaining why client_secret in a native open-source app binary is intentional) is preserved in `Secrets.swift`, so no documentation is lost.

---

### ✅ `GitHub.fetchActiveJobs` — `swiftlint:disable for_where` comment

The comment placement and wording match the issue spec exactly:
```swift
// filter() cannot replace this loop: insert() mutates seenRunIDs as a side effect.
// swiftlint:disable for_where
```
This is the right way to document a suppression — explain *why* the lint rule doesn't apply rather than just silencing it.

---

## Issue Checklist Coverage

All 10 items from issue #1100 are addressed:

| # | Item | Status |
|---|------|--------|
| 1 | `signIn()` force-unwrap | ✅ Fixed |
| 2 | `exchangeCode()` error detection | ✅ Fixed |
| 3 | `handleCallback()` nil-state vs CSRF | ✅ Fixed |
| 4 | Delete `GitHubAppSecrets.swift` | ✅ Done |
| 5 | `signOut()` gate on Keychain delete | ✅ Fixed |
| 6 | `Auth.swift` stale comment | ✅ Fixed |
| 7 | Keychain `kSecAttrAccessibleAfterFirstUnlock` update path | ✅ Fixed |
| 8 | `ghAPI`/`ghAPIPaginated` double token read | ✅ Fixed |
| 9 | `fetchStepLog` → `urlSessionRaw()` migration | ✅ Done |
| 10 | `fetchActiveJobs` `for_where` comment | ✅ Done |

---

## Minor items to consider (non-blocking)

1. **`exchangeCode` silent failure path** — the final `guard let token` has no log. Add a one-liner with the response keys for diagnosability (see note above).
2. **`signOut` UX** — if `Keychain.delete()` fails, the user has no way to sign out and no visible feedback. Consider surfacing this as an error state in a future pass.
3. **`NSWorkspace.open` return value in `signIn()`** — low priority, but if you want full error coverage, checking the return value and calling `onCompletion?(false)` if it fails would close the last unhandled path.

---

**Verdict: ready to merge.** The critical OAuth fixes are correct, the transport cleanup is sound, and the issue spec was fully executed. The minor items above are all future improvements, not blockers.

Issue #957, opened 2026-05-26
🪟 Liquid Glass audit: 7 improvements from LiquidGlassReference guide


eonist:
## Overview

A full audit of RunnerBar's current Liquid Glass implementation was conducted against the [conorluddy/LiquidGlassReference](https://github.com/conorluddy/LiquidGlassReference) guide — a comprehensive iOS 26 / macOS Tahoe SwiftUI reference.

Ref: #928

## What's already correct ✅

- `PanelViewModifiers.swift` centralises all glass behind `#available(macOS 26, *)` with proper `.ultraThinMaterial` fallbacks — no raw `.glassEffect()` leaks into view files.
- `CardRowModifier` correctly forbids `.glassEffect` on scrollable list content (Apple HIG: glass must not be applied to lists/tables).
- `PanelChrome.swift` correctly uses `NSGlassEffectView` (via reflective instantiation) on macOS 26+, falling back to `NSVisualEffectView(.hudWindow)`.

### #1 — `GlassButton` creates a per-button `GlassEffectContainer`

**File:** `PanelViewModifiers.swift` → `GlassButton`  
**Problem:** Each `GlassButton` wraps itself in its own `GlassEffectContainer`. The guide is explicit: multiple separate containers prevent sibling buttons from sharing a CABackdropLayer sampling region, breaking morphing and causing redundant GPU compositing passes.  
**Fix:** Remove `GlassEffectContainer` from `GlassButton`. Call sites that group multiple buttons must provide a shared container.

---

### #2 — Header buttons use `.buttonStyle(.plain)` with no glass

**File:** `PanelMainView+Subviews.swift` → `PanelHeaderView`  
**Problem:** The `gearshape` settings button and `xmark` quit button use `.buttonStyle(.plain)` — no glass, no interactivity. The guide's `.interactive()` modifier enables scaling-on-press, bounce, shimmer, and touch-point illumination — all of which are appropriate for toolbar-level controls.  
**Fix:** Wrap both buttons in a shared `GlassEffectContainer` and apply `.glassEffect(.regular.interactive())` (or `.buttonStyle(.glass)`) on macOS 26+.

---

### #3 — `StatusBadge` uses a raw `.background` layer instead of `.tint()` on Glass

**File:** `PanelViewModifiers.swift` → `StatusBadgeBackground`  
[...]

PR #1107, opened 2026-06-03
Batch 20 — DesignSystem + PanelMainView+Subviews: token cleanup, dead code removal, structural deduplication

### LINKED ISSUE #1105: Batch 20 — DesignSystem + PanelMainView+Subviews: token cleanup, dead code removal, structural deduplication
Tracking issue for Batch 20 of the ongoing cleanup project (#1038).

## Files in scope

- `Sources/RunnerBar/DesignSystem/DesignTokens.swift` (12 KB)
- `Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift` (15 KB)
- `Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift` (1.6 KB)
- `Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift` (21 KB)

---

### 1. `DesignTokens.swift` — `adaptive()` force-unwrap crash path

`appearance.bestMatch(from: darkMatches + [.aqua])!` force-unwraps the result. If Apple ever returns an appearance name outside the provided set, this crashes at colour resolution time — i.e. on every single view render. Fix:

```swift
// Before
return darkMatches.contains(appearance.bestMatch(from: darkMatches + [.aqua])!)
    ? NSColor(dark) : NSColor(light)

// After
let best = appearance.bestMatch(from: darkMatches + [.aqua])
return darkMatches.contains(best ?? .aqua) ? NSColor(dark) : NSColor(light)
```

### 2. `RemovalAlertModifier.swift` — alert message references removed `gh auth login` auth path

The unauthenticated alert message tells users to run `` `gh auth login` `` as a recovery step. The Batch 18 analysis (issue #1100) established that `gh auth login` is no longer a supported auth path in `Auth.swift` — it was removed from the priority chain. Showing it in the UI is misleading. Update to match actual auth sources:

```swift
// Before
Text("A GitHub token is required to de-register the runner from GitHub. "
    + "Sign in via `gh auth login` or set GH_TOKEN / GITHUB_TOKEN env var, then try again.")

// After
Text("A GitHub token is required to de-register the runner from GitHub. "
    + "Sign in via Settings or set GH_TOKEN / GITHUB_TOKEN env var, then try again.")
```

### 3. `PanelMainView+Subviews.swift` — `glassMorphBody` / `legacyBody` structural duplication

Both computed properties share an identical `VStack` + `HStack(spacing: 0)` + `InlineJobRowsView` + `.clipShape` + `.contentShape` + `.workflowContextMenu` + `.modifier(RowTapModifier)` + `.padding` + `.onAppear` + `.onChange` chain. The only difference is the background modifier. Any future layout change (padding, spacing, accessibility label) must be applied in both places — a classic agent trap.

Extract a shared private builder:
```swift
@ViewBuilder
private func rowContainer<Background: View>(@ViewBuilder background: () -> Background) -> some View {
    VStack(alignment: .leading, spacing: 0) {
        HStack(spacing: 0) {
            Color.clear.frame(width: RBSpacing.md)
            rowContent
        }
        if let fullExpand = expandState {
            InlineJobRowsView(group: group, tick: tick, fullExpand: fullExpand, onStepTap: onStepTap)
        }
    }
    .frame(maxWidth: .infinity)
    .background { background() }
    .clipShape(RoundedRectangle(cornerRadius: RBRadius.card, style: .continuous))
    .contentShape(RoundedRectangle(cornerRadius: RBRadius.card, style: .continuous))
    .workflowContextMenu(group: group)
    .modifier(RowTapModifier(jobs: group.jobs, expandState: $expandState, rowStatus: rowStatus, useBouncyAnimation: false))
    .padding(.horizontal, RBSpacing.md)
    .padding(.vertical, RBSpacing.xxs)
    .onAppear { applyInitialExpandState() }
    .onChange(of: rowStatus) { _, newStatus in handleStatusChange(newStatus) }
}
```

Then `glassMorphBody` and `legacyBody` each call `rowContainer` with only their specific background.

### 4. `PanelMainView+Subviews.swift` — runner cap `3` is a magic number

`active.prefix(3)` and `active.count > 3` are inline magic numbers with no explanation for why 3 was chosen. Hoist to a private constant:

```swift
/// Maximum number of runner cards shown before the overflow label.
private let maxVisibleRunners = 3
```

Then use `active.prefix(maxVisibleRunners)` and `active.count > maxVisibleRunners`.

### 5. `PanelMainView+Subviews.swift` — inline branch block duplicates `BranchTagPill`

`rowContent` builds its own branch icon + label inline:
```swift
if let branch = group.headBranch {
    HStack(spacing: 3) {
        Image(systemName: "arrow.triangle.branch")
            .font(.system(size: 9))
            .foregroundColor(.secondary)
        Text(branch)
            .font(DesignTokens.Fonts.mono)
            .foregroundColor(.secondary)
            .lineLimit(1)
            .truncationMode(.middle)
            .frame(maxWidth: 80, alignment: .leading)
    }
    .fixedSize(horizontal: false, vertical: true)
    .layoutPriority(0)
}
```

`BranchTagPill` already encapsulates exactly this pattern (branch icon + name, accent-tinted capsule, truncation). Replace with:
```swift
if let branch = group.headBranch {
    BranchTagPill(name: branch)
        .layoutPriority(0)
}
```

Note: verify the `maxWidth: 80` constraint is preserved inside `BranchTagPill` or add it at the call site if needed.

---

### 6. `DesignTokens.swift` — `DesignTokens` shim has no deprecation marker

The doc says *"Prefer the `RB*` types directly in new code"* but there is no `@available(*, deprecated, renamed: ...)` annotation or `// TODO: deprecate once all call sites migrated` comment. Future agents will keep adding new `DesignTokens.Fonts` / `DesignTokens.Spacing` usages instead of using `RBFont` / `RBSpacing` directly. Add:

```swift
/// - Note: Deprecated shim — prefer `RBFont`, `RBSpacing`, `RBRadius`, and `Color` token
/// extensions directly. This namespace will be removed once all call sites are migrated.
enum DesignTokens { ... }
```

Or annotate each inner type:
```swift
@available(*, deprecated, renamed: "RBFont")
enum Fonts { ... }
```

### 7. `PanelViewModifiers.swift` — `GlassButton` empty `else` branch looks like forgotten code

The pre-26 `else` branch returns `content` with no styling and no comment, which looks like an accidentally empty block to a future reader:
```swift
} else {
    content // pre-macOS-26: no interactive glass style — passes through unstyled intentionally
}
```

### 8. `PanelMainView+Subviews.swift` — `useBouncyAnimation` is always `false` (dead code)

Both `glassMorphBody` and `legacyBody` pass `useBouncyAnimation: false` to `RowTapModifier`. The `? .bouncy : .easeInOut(duration: 0.15)` ternary inside `RowTapModifier` is therefore dead code — `.bouncy` is never reached. The MARK comment in `ActionRowView` explicitly says *"Do NOT add `.bouncy`"* (#957). Remove `useBouncyAnimation` from `RowTapModifier` and collapse to `.easeInOut(duration: 0.15)` directly:

```swift
// Before
let animation: Animation = useBouncyAnimation ? .bouncy : .easeInOut(duration: 0.15)

// After
let animation: Animation = .easeInOut(duration: 0.15)
```

Also remove the `useBouncyAnimation: Bool` property and all call-site arguments.

---

### Code fixes
- [ ] `DesignTokens.adaptive()`: replace force-unwrap with nil-coalescing guard
- [ ] `RemovalAlertModifier`: replace `gh auth login` reference with `Sign in via Settings`
- [ ] `PanelMainView+Subviews`: extract shared `rowContainer` builder, eliminate `glassMorphBody`/`legacyBody` duplication
- [ ] `PanelMainView+Subviews`: hoist runner cap `3` to `private let maxVisibleRunners`
- [ ] `PanelMainView+Subviews`: replace inline branch block with `BranchTagPill(name: branch)`

### Minor / doc
- [ ] `DesignTokens`: add deprecation marker / TODO comment on `DesignTokens` shim
- [ ] `PanelViewModifiers.GlassButton`: add explanatory comment on empty pre-26 `else` branch
- [ ] `PanelMainView+Subviews.RowTapModifier`: remove `useBouncyAnimation` dead code, collapse to `.easeInOut`

### Branch & CI
- [ ] Branch `batch-20-design-system-subviews` created
- [ ] PR opened, CI green

eonist:
## **User description**
Closes #1105

### `DesignTokens.swift`
- **Fix:** Replace `adaptive()` force-unwrap with nil-coalescing guard — `bestMatch(...)!` → `bestMatch(...) ?? .aqua`
- **Doc:** Add `@available(*, deprecated)` note + TODO to `DesignTokens` shim directing new code to `RB*` types directly

### `PanelViewModifiers.swift`
- **Doc:** Add explanatory comment to `GlassButton` empty pre-26 `else` branch so it doesn't look like accidentally forgotten code

### `RemovalAlertModifier.swift`
- **Fix:** Replace stale `gh auth login` reference in unauthenticated alert message with `Sign in via Settings` — `gh auth login` was removed as a supported auth path in Batch 18 (`Auth.swift`)

### `PanelMainView+Subviews.swift`
- **Refactor:** Extract shared `rowContainer<Background>(@ViewBuilder background:)` builder — eliminates `glassMorphBody`/`legacyBody` structural duplication; layout changes now apply once
- **Fix:** Hoist runner cap magic number `3` to `private let maxVisibleRunners = 3`
- **Fix:** Replace inline branch icon+label block with `BranchTagPill(name: branch)` — the component exists exactly for this use case
- **Fix:** Remove `useBouncyAnimation: Bool` dead parameter from `RowTapModifier` — always `false`, `.bouncy` branch was never reached; collapse to `.easeInOut(duration: 0.15)` directly

## Checklist
- [x] `DesignTokens.adaptive()`: replace force-unwrap with nil-coalescing guard
- [x] `RemovalAlertModifier`: replace `gh auth login` reference with `Sign in via Settings`
- [x] `PanelMainView+Subviews`: extract shared `rowContainer` builder
- [x] `PanelMainView+Subviews`: hoist runner cap `3` to `private let maxVisibleRunners`
- [x] `PanelMainView+Subviews`: replace inline branch block with `BranchTagPill(name: branch)`
- [x] `DesignTokens`: add deprecation marker / TODO comment on `DesignTokens` shim
- [x] `PanelViewModifiers.GlassButton`: add explanatory comment on empty pre-26 `else` branch
- [x] `PanelMainView+Subviews.RowTapModifier`: remove `useBouncyAnimation` dead code

___

## **CodeAnt-AI Description**
Safer theme colors, clearer runner removal messaging, and cleaner row layout

### What Changed
- Color choices now fall back to light mode instead of crashing when the system appearance is unexpected
- The runner removal alert now tells users to sign in through Settings or use a token, instead of pointing to the removed `gh auth login` path
- Runner rows keep the same look and behavior while using the shared branch pill and a fixed inline runner count limit
- Row expand/collapse uses one consistent animation path, avoiding the slower bouncy behavior

### Impact
`✅ Fewer theme-related crashes`
`✅ Clearer runner removal instructions`
`✅ Consistent runner row display`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

### DIFF
--- a/Sources/RunnerBar/DesignSystem/DesignTokens.swift
+++ b/Sources/RunnerBar/DesignSystem/DesignTokens.swift
@@ -289,6 +289,7 @@ enum RBFont {
 enum DesignTokens {
     /// Font aliases forwarded from `RBFont`.
     /// - Note: Deprecated — use `RBFont` directly.
+    @available(*, deprecated, renamed: "RBFont")
     enum Fonts {
         /// Monospaced label font — alias for `RBFont.monoSmall`.
         static let monoLabel: Font = RBFont.monoSmall
@@ -297,18 +298,21 @@ enum DesignTokens {
     }
     /// Spacing aliases forwarded from `RBSpacing`.
     /// - Note: Deprecated — use `RBSpacing` directly.
+    @available(*, deprecated, renamed: "RBSpacing")
     enum Spacing {
         /// Horizontal row padding — alias for `RBSpacing.md`.
         static let rowHPad: CGFloat = RBSpacing.md
     }
     /// Radius aliases forwarded from `RBRadius`.
     /// - Note: Deprecated — use `RBRadius` directly.
+    @available(*, deprecated, renamed: "RBRadius")
     enum Radius {
         /// Card corner radius — alias for `RBRadius.card`.
         static let card: CGFloat = RBRadius.card
     }
     /// Color helpers forwarded from the `Color` token extensions.
     /// - Note: Deprecated — use the `Color.rb*` extensions directly.
+    @available(*, deprecated, message: "Use Color.rb* extensions directly.")
     enum Colors {
         /// Returns a traffic-light color based on a usage percentage (0–100).
         /// - Green below 60 %, amber 60–85 %, red above 85 %.
--- a/Sources/RunnerBar/DesignSystem/DesignTokens.swift
+++ b/Sources/RunnerBar/DesignSystem/DesignTokens.swift
@@ -17,7 +17,11 @@ extension Color {
                 .accessibilityHighContrastDarkAqua,
                 .accessibilityHighContrastVibrantDark
             ]
-            return darkMatches.contains(appearance.bestMatch(from: darkMatches + [.aqua])!)
+            // Use nil-coalescing instead of force-unwrap — bestMatch returns nil if the
+            // provided array is empty, which cannot happen here, but crashing on a novel
+            // system appearance name is worse than silently falling back to light mode.
+            let best = appearance.bestMatch(from: darkMatches + [.aqua])
+            return darkMatches.contains(best ?? .aqua)
                 ? NSColor(dark)
                 : NSColor(light)
         })
@@ -275,9 +279,16 @@ enum RBFont {
 // MARK: - DesignTokens namespace shim
 
 /// Backwards-compatibility namespace that delegates to the primary `RBFont`, `RBSpacing`,
-/// and `Color` token types. Prefer the `RB*` types directly in new code.
+/// and `Color` token types.
+///
+/// - Note: **Deprecated shim** — prefer `RBFont`, `RBSpacing`, `RBRadius`, and the `Color`
+/// token extensions directly in all new code. This namespace exists only to avoid a
+/// mass rename of existing call sites. Do not add new members here.
+/// TODO: Remove once all remaining call sites (`DesignTokens.Fonts.*`, `DesignTokens.Spacing.*`,
+/// `DesignTokens.Radius.*`, `DesignTokens.Colors.*`) are migrated to the `RB*` equivalents.
 enum DesignTokens {
     /// Font aliases forwarded from `RBFont`.
+    /// - Note: Deprecated — use `RBFont` directly.
     enum Fonts {
         /// Monospaced label font — alias for `RBFont.monoSmall`.
         static let monoLabel: Font = RBFont.monoSmall
@@ -285,16 +296,19 @@ enum DesignTokens {
         static let mono: Font = RBFont.mono
     }
     /// Spacing aliases forwarded from `RBSpacing`.
+    /// - Note: Deprecated — use `RBSpacing` directly.
     enum Spacing {
         /// Horizontal row padding — alias for `RBSpacing.md`.
         static let rowHPad: CGFloat = RBSpacing.md
     }
     /// Radius aliases forwarded from `RBRadius`.
+    /// - Note: Deprecated — use `RBRadius` directly.
     enum Radius {
         /// Card corner radius — alias for `RBRadius.card`.
         static let card: CGFloat = RBRadius.card
     }
     /// Color helpers forwarded from the `Color` token extensions.
+    /// - Note: Deprecated — use the `Color.rb*` extensions directly.
     enum Colors {
         /// Returns a traffic-light color based on a usage percentage (0–100).
         /// - Green below 60 %, amber 60–85 %, red above 85 %.
--- a/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
+++ b/Sources/RunnerBar/DesignSystem/PanelViewModifiers.swift
@@ -112,6 +112,7 @@ struct GlassButton: ViewModifier {
                     in: RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
                 )
         } else {
+            // pre-macOS-26: no interactive glass button style available — passes through unstyled intentionally.
             content
         }
     }
--- a/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
+++ b/Sources/RunnerBar/DesignSystem/RemovalAlertModifier.swift
@@ -31,8 +31,10 @@ struct RemovalAlertModifier: ViewModifier {
                 Text("This will run ./svc.sh uninstall and ./config.sh remove. "
                     + "A GitHub token is required for de-registration.")
             } else {
+                // Note: gh auth login is no longer a supported auth path (removed in Batch 18 Auth.swift).
+                // Direct users to Settings OAuth flow or the GH_TOKEN / GITHUB_TOKEN env vars instead.
                 Text("A GitHub token is required to de-register the runner from GitHub. "
-                    + "Sign in via `gh auth login` or set GH_TOKEN / GITHUB_TOKEN env var, then try again.")
+                    + "Sign in via Settings or set GH_TOKEN / GITHUB_TOKEN env var, then try again.")
             }
         }
     }
--- a/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
+++ b/Sources/RunnerBar/Views/Main/PanelMainView+Subviews.swift
@@ -150,17 +150,19 @@ private struct RunnerMetricsBadge: View {
 /// on a separate background cycle and will always lag behind the RunnerStore
 /// fetch cycle, causing rows to be silently swallowed. (#948)
 struct PanelLocalRunnerRow: View {
+    /// Maximum number of runner cards shown inline before the overflow label.
+    private let maxVisibleRunners = 3
     /// The list of runners to display.
     let runners: [RunnerModel]
     /// Renders the runner card list, or nothing if `runners` is empty.
     var body: some View {
         if !runners.isEmpty { runnerList(runners) }
     }
-    /// Renders a vertical stack of runner cards, capped at 3 with an overflow label.
+    /// Renders a vertical stack of runner cards, capped at `maxVisibleRunners` with an overflow label.
     @ViewBuilder private func runnerList(_ active: [RunnerModel]) -> some View {
-        ForEach(active.prefix(3)) { runner in runnerCard(runner) }
-        if active.count > 3 {
-            Text("+ \(active.count - 3) more…")
+        ForEach(active.prefix(maxVisibleRunners)) { runner in runnerCard(runner) }
+        if active.count > maxVisibleRunners {
+            Text("+ \(active.count - maxVisibleRunners) more…")
                 .font(.caption2).foregroundColor(.secondary)
                 .padding(.horizontal, DesignTokens.Spacing.rowHPad).padding(.vertical, 2)
         }
@@ -243,21 +245,30 @@ struct ActionRowView: View {
     @State private var expandState: Bool?
     /// The row status observed on the previous tick, used to detect transitions.
     @State private var previousStatus: RBStatus?
-    /// Routes to `glassMorphBody` on macOS 26+ or `legacyBody` on earlier OS versions.
+    /// Routes to the macOS 26+ glass path or the pre-26 legacy path.
     var body: some View {
         if #available(macOS 26, *) {
-            glassMorphBody
+            rowContainer {
+                // macOS 26+: glassCard applied directly; no ZStack needed.
+                Color.clear.glassCard(cornerRadius: RBRadius.card)
+                statusAccentBar
+            }
         } else {
-            legacyBody
+            rowContainer {
+                glassCardBackground
+                statusAccentBar
+            }
         }
     }
 
-    // MARK: macOS 26+ body
-    /// macOS 26+ path — uses .glassCard() directly on the VStack with the same
-    /// .easeInOut(duration: 0.15) animation as main. No GlassEffectContainer,
-    /// no .bouncy, no .glassEffectTransition — these caused staggered/slow animation.
-    @available(macOS 26, *)
-    @ViewBuilder private var glassMorphBody: some View {
+    // MARK: Shared row container
+
+    /// Shared layout structure used by both the macOS 26+ and pre-26 paths.
+    /// Only the background layer differs between the two paths — all other
+    /// modifiers (clip, content shape, context menu, tap, padding, lifecycle)
+    /// are applied once here to eliminate duplication.
+    @ViewBuilder
+    private func rowContainer<Background: View>(@ViewBuilder background: () -> Background) -> some View {
         VStack(alignment: .leading, spacing: 0) {
             HStack(spacing: 0) {
                 Color.clear.frame(width: RBSpacing.md)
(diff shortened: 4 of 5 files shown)


eonist:
## Deep Review — Batch 20

Overall: this is a well-scoped batch. All 8 checklist items from #1105 are addressed, the diff is net-negative lines, and no new abstractions are introduced that aren't warranted. That said, a few things deserve a closer look.

---

### ✅ What's done correctly

**`DesignTokens.adaptive()` force-unwrap fix**
The nil-coalescing guard is the right call. `bestMatch(from:)` returns `nil` only when the input array is empty — which is structurally impossible here since `darkMatches + [.aqua]` always has at least one element — but the comment acknowledges this reasoning explicitly, which is exactly the right way to document a "can't happen, but just in case" guard. Falling back to `.aqua` (light mode) on an unknown appearance name is a reasonable sentinel — it won't crash, and the worst-case visual result is a light-themed panel on a dark system, which is visible but non-destructive.

**`RemovalAlertModifier` alert copy**
The stale `gh auth login` reference is correctly replaced. The inline comment (`// Note: gh auth login is no longer a supported auth path`) gives future readers the rationale without having to dig through Batch 18 history. Clean.

**`rowContainer` extraction**
This is the most impactful change in the batch. The original `glassMorphBody`/`legacyBody` split had ~40 lines of structurally identical code in both branches — a genuine maintenance trap for any future padding, spacing, or accessibility change. The `@ViewBuilder` `rowContainer<Background: View>` builder is the right Swift pattern here. The generics are correctly constrained, the `@ViewBuilder` parameter label is clear, and all modifiers (`.clipShape`, `.contentShape`, `.workflowContextMenu`, `.modifier(RowTapModifier...)`, `.padding`, `.onAppear`, `.onChange`) now live in exactly one place.

**`statusAccentBar` extraction**
The left-edge status bar was duplicated across both old bodies. Extracting it into a dedicated `@ViewBuilder private var statusAccentBar` is a logical follow-on from the `rowContainer` refactor. It's clean.

**`BranchTagPill` adoption**
Replacing the inline `HStack { Image + Text }` branch block with `BranchTagPill(name:)` is correct. The `maxWidth: 80` constraint is preserved at the call site (`.frame(maxWidth: 80, alignment: .leading)`) — which the issue specifically flagged as needing verification. Good.

**`useBouncyAnimation` removal**
Dead parameter correctly removed. The comment added to `RowTapModifier` — `/// Animation is always .easeInOut(duration: 0.15) — do NOT add .bouncy (#957).` — is valuable; it explains *why* the simpler path was chosen rather than just leaving a naked `.easeInOut`.

---

### 🟡 Issues / concerns

**1. `rowContainer` background wrapping in a `ZStack` — may be unintentional**

In `rowContainer`, the background closure is wrapped in a `ZStack`:

```swift
.background {
    ZStack {
        background()
    }
    .clipShape(...)
}
```

For the legacy path this `ZStack` is load-bearing — `glassCardBackground` and `statusAccentBar` are stacked. But for the macOS 26+ path:

```swift
rowContainer {
    Color.clear.glassCard(cornerRadius: RBRadius.card)
    statusAccentBar
}
```

`statusAccentBar` is *inside* the closure passed to `rowContainer`, so it ends up inside the `ZStack` inside `.background`. In the old `glassMorphBody`, `statusAccentBar` was also in the background layer (inside `.background { Rectangle()... }`), so this is functionally equivalent. But visually it means the accent bar is rendered *behind* `glassCard` on the 26+ path. Is that intentional? In the old code:

```swift
.background {
    Rectangle()
        .fill(rowStatus.color)
        .frame(width: 4)
        ...
}
.glassCard(...)
```

The glass effect was applied *after* the background, so the accent bar was under the glass. In the new code, `Color.clear.glassCard(...)` and `statusAccentBar` are both in the same `ZStack` inside `.background` — the Z-ordering depends on declaration order. Since `Color.clear.glassCard` is declared first, `statusAccentBar` renders on top of it inside the ZStack. This is probably correct, but it's worth explicitly verifying that the accent bar is still visible through the glass on macOS 26.

**2. `DesignTokens` deprecation stops short of `@available`**

The issue (#1105, item 6) proposed either doc-comment deprecation *or* `@available(*, deprecated, renamed:)` annotations. The PR chose doc-comment-only (`/// - Note: Deprecated`). That's valid, but it means Xcode won't emit a warning when new code references `DesignTokens.Fonts.*` or `DesignTokens.Spacing.*`. A future agent or developer won't be stopped — they'll just silently add new usages.

The TODO comment at the top-level enum is good for tracking, but consider whether the inner enums (`Fonts`, `Spacing`, `Radius`, `Colors`) should each get `@available(*, deprecated, renamed: "RBFont")` etc. now rather than later. The PR description acknowledges this but doesn't act on it. If the migration is actively in progress (which Batch 18–20 suggests it is), adding the compiler annotation now costs ~4 lines and immediately surfaces all remaining call sites.

**3. `rowContainer` `ZStack` wrapper adds a layer with no benefit for the 26+ path**

Related to point 1: the `.background { ZStack { background() }.clipShape(...) }` pattern wraps *any* background in a `ZStack`, even when only a single view is passed (the 26+ path). A `ZStack` with a single child is a no-op visually, but it does add a layout pass. This is a micro-concern, not a bug — but it's slightly inconsistent with the issue's proposed shape of `rowContainer`, which used `.background { background() }` directly (without the intermediate `ZStack`). The `ZStack` appears to be a remnant from adapting the legacy body. Could simplify to:

```swift
.background {
    background()
        .clipShape(RoundedRectangle(cornerRadius: RBRadius.card, style: .continuous))
}
```

This works because both backgrounds are already single-view or ZStack'd internally where needed.

**4. `BranchTagPill` `fixedSize` placement**

In the old code:
```swift
HStack { ... }
    .fixedSize(horizontal: false, vertical: true)
    .layoutPriority(0)
```

In the new code:
```swift
BranchTagPill(name: branch)
    .frame(maxWidth: 80, alignment: .leading)
    .fixedSize(horizontal: false, vertical: true)
    .layoutPriority(0)
```

`.fixedSize(horizontal: false, vertical: true)` applied *after* `.frame(maxWidth: 80)` means the `fixedSize` will override the frame's horizontal constraint in the vertical dimension only — that's fine. But the horizontal constraint from `.frame(maxWidth: 80)` is still active. This is correct behaviour, but it's worth confirming that `BranchTagPill` itself doesn't internally apply any conflicting `.frame` or `.fixedSize` that would cause the 80pt cap to be ignored. If `BranchTagPill` uses a fixed internal width, the outer `.frame(maxWidth: 80)` becomes redundant rather than constraining.

---

### 🔵 Minor observations (no action required)

- The `maxVisibleRunners = 3` constant is declared on `PanelLocalRunnerRow` as an instance property (`let`), not a `static let`. Since it's a fixed value with no instance-specific behaviour, `static let` would be marginally more idiomatic — it signals "this is a type-level constant" and avoids capturing `self` if the value is ever used in a closure.

- The `GlassButton` comment (`// pre-macOS-26: no interactive glass button style available — passes through unstyled intentionally.`) is a bit long for an inline comment but communicates the right thing. No change needed.

- `RowTapModifier` doc update — `/// Shared by both the macOS 26+ and pre-26 paths via rowContainer` — is accurate and now reflects the architecture change. Good housekeeping.

---

### Summary

| Item | Status |
|---|---|
| Force-unwrap fix | ✅ Correct |
| `gh auth login` copy fix | ✅ Correct |
| `rowContainer` extraction | ✅ Correct — one ZStack concern worth verifying |
| `statusAccentBar` Z-ordering on macOS 26+ | 🟡 Needs visual verification |
| `BranchTagPill` + `maxWidth: 80` | 🟡 Verify `BranchTagPill` doesn't apply conflicting internal frame |
| `useBouncyAnimation` removal | ✅ Correct |
| `DesignTokens` deprecation | 🟡 Consider `@available(*, deprecated)` on inner enums |
| `maxVisibleRunners` as instance `let` | 🔵 Minor — `static let` would be more idiomatic |

Blockers: none. The `statusAccentBar` Z-ordering on 26+ and the `BranchTagPill` internal frame question are worth a quick manual check before merge, but neither is likely to cause a visible regression given the old code's behaviour was equivalent.

eonist:
## Deep Review — Batch 20

Overall this is a clean, well-scoped batch. All 8 checklist items from #1105 are addressed and the diff is net-negative (−10 lines). The changes are safe and low-risk. Below is a point-by-point assessment of every change against the original issue spec, followed by a few observations worth considering before merge.

---

### ✅ `DesignTokens.swift` — `adaptive()` force-unwrap fix

The fix is correct. `bestMatch(from:)` returns `nil` only when the provided array is empty, and since `darkMatches + [.aqua]` is always non-empty, this is theoretically unreachable — the comment in the diff acknowledges this honestly. The nil-coalescing to `.aqua` (light mode fallback) is the right safe default: better to render the wrong appearance than to crash on every view render tick with a novel macOS appearance name. No issues here.

One small nit: the binding expression is now split over two lines (`let best = …` + `return darkMatches.contains(best ?? .aqua)`). That's perfectly readable, but if you ever want to tighten it you could write `darkMatches.contains(appearance.bestMatch(from: darkMatches + [.aqua]) ?? .aqua)` inline. Matter of taste — the current form is arguably clearer.

---

### ✅ `DesignTokens.swift` — Deprecation markers on the shim

The issue asked for either a doc comment on the `DesignTokens` enum itself or `@available(*, deprecated, ...)` on each inner type. The PR went with both: a detailed `/// - Note: Deprecated shim —` doc block on the enum, plus individual `@available(*, deprecated, renamed:)` annotations on `Fonts`, `Spacing`, `Radius`, and `Colors`. This is strictly better than what the issue requested.

The annotation choices are appropriate: `renamed:` for `Fonts`, `Spacing`, and `Radius` (1:1 mappings), and `message:` for `Colors` (since `Color.rb*` isn't a direct rename). Good call.

**One concern worth flagging:** `@available(*, deprecated)` on a Swift `enum` type does not automatically surface a warning at call sites that use *members* of that type (e.g. `DesignTokens.Fonts.monoLabel`). The compiler emits the deprecation warning at the point where the *type itself* is referenced. Since the goal is to migrate call sites like `DesignTokens.Fonts.monoLabel` → `RBFont.monoSmall`, callers using the dot-chain may not see a warning if the `Fonts` enum is inferred via the outer `DesignTokens` scope. Worth testing: does `let x = DesignTokens.Fonts.monoLabel` actually emit a deprecation warning in Xcode? If not, individual static property annotations may be needed on the leaves.

---

### ✅ `PanelViewModifiers.swift` — `GlassButton` empty else comment

The comment `// pre-macOS-26: no interactive glass button style available — passes through unstyled intentionally.` directly matches what the issue requested. Exactly the right amount of explanation — no over-engineering.

---

### ✅ `RemovalAlertModifier.swift` — `gh auth login` → `Sign in via Settings`

String change is correct and matches the issue spec. The inline comment (`// Note: gh auth login is no longer a supported auth path...`) is a good knowledge anchor for future editors. The `docs/AGENTS.md` update is a solid addition beyond the issue scope — it ensures agents don't regenerate the old string, which is the original motivation for fixing this in the first place.

**One observation:** The replacement `"Sign in via Settings"` is somewhat vague from a UX perspective. It doesn't tell the user *where* in Settings or what they're looking for. If there's a specific Settings tab name (e.g. "Settings → Account"), that would be more actionable. Not a blocker, but consider a follow-up.

---

### ✅ `PanelMainView+Subviews.swift` — `rowContainer` refactor

This is the most significant structural change in the PR. The extraction is clean and the approach is exactly what the issue specified.

**What the refactor does well:**
- All shared modifiers (`.clipShape`, `.contentShape`, `.workflowContextMenu`, `.modifier(RowTapModifier)`, `.padding`, `.onAppear`, `.onChange`) are now applied once. This is the core value.
- The generic `<Background: View>` `@ViewBuilder` parameter is the correct Swift idiom for this pattern.
- The call sites in `body` are now ~5 lines each vs. 25+ lines each before. Big readability win.

**One subtle difference between old and new worth flagging:**

In the old `glassMorphBody`, `.glassCard()` was applied *directly on the `VStack`* as a modifier (content layer). In the new code, `Color.clear.glassCard(cornerRadius: RBRadius.card)` is passed into `rowContainer` as a background element inside a `ZStack`.

```swift
// New macOS 26+ path — stacking order inside background ZStack:
rowContainer {
    Color.clear.glassCard(cornerRadius: RBRadius.card)  // bottom of ZStack
    statusAccentBar                                       // on top
}
```

The status accent bar is now *on top of* the glass card layer inside the background ZStack. In the old code the glass effect was on the `VStack` itself, layered differently relative to clipping/content. Visually this is likely fine since the accent bar is opaque and the glass effect is on the background, but it's worth **a visual smoke test on macOS 26** — specifically that the 4pt accent bar still renders correctly and that the glass morphism effect doesn't bleed over or occlude it due to the layer ordering change.

Glass effects in SwiftUI can behave differently depending on whether they're applied to a content view vs. a background-layer view. The visual output may be identical in practice, but this warrants an explicit check.

---

### ✅ `PanelMainView+Subviews.swift` — `statusAccentBar` extraction

The 4pt left-edge accent bar was previously duplicated between `glassMorphBody` and `legacyBody`. Extracting it to `@ViewBuilder private var statusAccentBar` is correct. The implementation matches both originals exactly — same `width: 4`, same `maxHeight: .infinity`, same `maxWidth: .infinity, alignment: .leading`. No behaviour drift.

---

### ✅ `PanelMainView+Subviews.swift` — `maxVisibleRunners = 3`

Clean. The constant is `private static let` (not `private let`) which is correct for a `struct` — avoids a per-instance stored property for a value that never varies. The comment is clear. All three call sites use `Self.maxVisibleRunners` which is the right disambiguation in a `View` struct.

---

### ✅ `PanelMainView+Subviews.swift` — `BranchTagPill` replacement

The issue asked to verify that `maxWidth: 80` is preserved either inside `BranchTagPill` or at the call site. The PR adds `.frame(maxWidth: 80, alignment: .leading)` at the call site — correct. The issue's proposed solution actually omitted this, so the PR is more complete than the spec here.

**One thing to verify:** The original inline block used `.truncationMode(.middle)` and `DesignTokens.Fonts.mono` font. Does `BranchTagPill` internally use the same font and truncation mode? If it truncates at `.tail` or uses a different font size, this is a subtle but visible behaviour change in the row. Since `BranchTagPill` "exists exactly for this use case" per the PR description, this is likely fine — but worth a quick check if there are any snapshot tests.

---

### ✅ `PanelMainView+Subviews.swift` — `useBouncyAnimation` dead code removal

Correct removal. The MARK comment (`⚠️ Do NOT add .bouncy`) already documented the intent. Removing the dead branch and collapsing to `.easeInOut(duration: 0.15)` directly eliminates one surface area where a future editor could accidentally re-enable `.bouncy`. The RowTapModifier doc comment now explicitly states `Animation is always .easeInOut(duration: 0.15) — do NOT add .bouncy (#957)` which is a good guard rail.

---

### Things not in scope but worth noting

1. **`maxVisibleRunners = 3` value is undocumented as a product decision.** The constant explains *what* it is but not *why 3*. A brief note like `// 3 keeps the panel under ~300pt at default font size` (or whatever the actual reason is) would help future editors know whether this is a hard UX constraint or an arbitrary starting point.

2. **`DesignTokens.Colors` deprecation message** says `"Use Color.rb* extensions directly."` — verify that the actual extension methods follow a `rb*` naming convention. If they're named differently, the message is slightly misleading.

3. **`docs/AGENTS.md` auth fallback chain** now lists 5 items. Double-check that the numbering matches what `Auth.swift` actually implements — item 3 is `GITHUB_TOKEN` and item 4 is the Settings OAuth flow, which should be correct per the Batch 18 context.

---

### Summary verdict

All 8 checklist items from #1105 are correctly implemented. The code is cleaner and safer after this PR. Two things I'd recommend verifying before merge:

- **Visual smoke test on macOS 26:** confirm the glass card + status accent bar renders identically before/after the `rowContainer` layer-order change.
- **`BranchTagPill` internals:** confirm font and truncation mode match what the inline block previously rendered.

Everything else is straightforward. 🟢

PR #1108, opened 2026-06-03
Batch 21 — Sheets + StepLog: GCD→async/await migration, duplicate logic, inconsistent contracts

### LINKED ISSUE #1106: Batch 21 — Sheets + StepLog: GCD→async/await migration, duplicate logic, inconsistent contracts
Tracking issue for Batch 21 of the ongoing cleanup project (#1038).

## Files in scope

- `Sources/RunnerBar/Views/Sheets/BranchSelectorSheet.swift` (10 KB)
- `Sources/RunnerBar/Views/Sheets/RepoSelectorSheet.swift` (6 KB)
- `Sources/RunnerBar/Views/StepLog/LogCopyButton.swift` (4 KB)
- `Sources/RunnerBar/Views/StepLog/StepLogView.swift` (16 KB)

---

### 1. `StepLogView.swift` — `repoSlug` parsing duplicated in `loadLog()`

`loadLog()` re-implements the exact same `job.htmlUrl` parsing logic (`components(separatedBy: "/")`, `parts[3]`, `parts[4]`) that already exists as the `repoSlug` computed property in the extension below. Any future change to slug parsing must currently be applied in two places — a classic agent trap.

Replace the inline parse in `loadLog()` with a call to `self.repoSlug`:

```swift
// Before (inside loadLog)
let scope: String = {
    let parts = (job.htmlUrl ?? "").components(separatedBy: "/")
    if parts.count >= 5 {
        let owner = parts[3]
        let repo = parts[4]
        if !owner.isEmpty && !repo.isEmpty { return "\(owner)/\(repo)" }
    }
    return ScopeStore.shared.scopes.first(where: { $0.contains("/") }) ?? ""
}()

// After
let scope = repoSlug == "—"
    ? ScopeStore.shared.scopes.first(where: { $0.contains("/") }) ?? ""
    : repoSlug
```

### 2. `BranchSelectorSheet.swift` — `loadBranches()` uses raw GCD

`loadBranches()` dispatches to `DispatchQueue.global` then hops back via `DispatchQueue.main.async` to mutate `@State`. This is the same GCD pattern replaced throughout Batches 9–19. Migrate to `Task.detached` + `await MainActor.run`:

```swift
func loadBranches() {
    log("BranchSelectorSheet › loadBranches START scope='\(scope)'")
    Task.detached {
        let names = self.fetchBranchNames(scope: self.scope)
        await MainActor.run {
            if let names, !names.isEmpty {
                log("BranchSelectorSheet › loadBranches — loaded \(names.count) branches")
                self.branches = names
                self.loadError = false
            } else {
                log("BranchSelectorSheet › loadBranches — fetch failed or returned empty")
                self.loadError = true
            }
            self.isLoading = false
        }
    }
}
```

`fetchBranchNames` is already `nonisolated` — no change needed there.

---

### 3. `LogCopyButton.swift` — `startCopy()` uses GCD for `@State` mutations

`startCopy()` mutates `@State` via `DispatchQueue.main.async` and `DispatchQueue.main.asyncAfter`. Since `LogCopyButton` is a SwiftUI `View` these mutations are already on MainActor. Migrate to `Task` + `Task.sleep`:

```swift
private func startCopy() {
    guard phase == .idle else { return }
    phase = .loading
    fetch { copyText in
        Task { @MainActor in
            if let text = copyText, !text.isEmpty {
                phase = .done
            } else {
                phase = .failed
            }
            try? await Task.sleep(for: .milliseconds(1500))
            phase = .idle
        }
    }
}
```

### 4. `StepLogView.swift` — `loadLog()` uses raw GCD

Same pattern as #2. Migrate to `Task.detached` + `await MainActor.run`:

```swift
private func loadLog() {
    isLoading = true
    let jobID = job.id
    let stepNum = step.id
    let scope = repoSlug == "—"
        ? ScopeStore.shared.scopes.first(where: { $0.contains("/") }) ?? ""
        : repoSlug
    Task.detached {
        let text = fetchStepLog(jobID: jobID, stepNumber: stepNum, scope: scope)
        await MainActor.run {
            logText = text ?? ""
            isLoading = false
            onLogLoaded?()
        }
    }
}
```

### 5. `RepoSelectorSheet.swift` — `itemRow` inconsistent dismiss contract

`BranchSelectorSheet.branchRow` calls `onSelect(branch)` then `onDismiss()` explicitly. `RepoSelectorSheet.itemRow` calls only `onSelect(item)` and relies on the caller's closure to dismiss. This asymmetry means a caller that forgets to dismiss inside `onSelect` will leave the sheet open with no feedback.

Either unify by calling `onDismiss()` after `onSelect(item)` in `itemRow`, or add an explicit doc comment explaining the deliberate asymmetry:

```swift
// Option A — unify (preferred)
Button(action: {
    log("RepoSelectorSheet › selected item='\(item)'")
    onSelect(item)
    onDismiss()  // added — matches BranchSelectorSheet contract
}) { ... }

// Option B — document intent if asymmetry is deliberate
// ℹ️ Intentionally does NOT call onDismiss() — callers are expected to dismiss
// inside their onSelect closure. This allows callers to perform async work before
// dismissing (e.g. AddScopeSheet validates the selection first).
```

### 6. `BranchSelectorSheet.swift` / `RepoSelectorSheet.swift` — misplaced doc comment inside `searchSection`

Both files contain a stray `/// The listSection computed view.` doc comment placed *inside* the closing brace of `searchSection` (before `listSection` is declared). It's syntactically harmless but will mislead any future agent reading the structure. Move or remove the comment in both files.

---

### 7. `BranchSelectorSheet.swift` — tautological stub docs

Replace all auto-generated stub docs with real descriptions or remove them:
- `/// A value type representing BranchSelectorSheet.` → real struct-level doc already exists in the MARK block above; remove duplicate
- `/// The scope constant.` / `/// The onDismiss constant.` / `/// The onSelect constant.` → document *purpose* (scope string = `owner/repo`, onDismiss = called on any exit path, onSelect(nil) = clears filter)
- `/// Extension adding functionality to BranchSelectorSheet.` (×2) → replace with `/// Subview builders.` / `/// Data loading.`
- `/// Performs the branchRow operation.` → replace with row contract doc
- `/// The headerSection property.` / `/// The searchSection property.` / `/// The footerSection property.` / `/// The body property.` → remove (MARK headers are sufficient)

### 8. `RepoSelectorSheet.swift` — tautological stub docs

Same pattern as #7. Replace or remove all auto-generated `/// The X constant/property.` and `/// Extension adding functionality to RepoSelectorSheet.` docs.

### 9. `LogCopyButton.swift` — tautological stub docs

- `/// The fetch constant.` → document the contract: callback-based fetch; pass nil or empty string on failure
- `/// The isDisabled property.` → document: reduces opacity and blocks tap when true
- `/// The phase property.` → remove (Phase enum doc is sufficient)
- `/// The body property.` / `/// Performs the startCopy operation.` → remove

### 10. `StepLogView.swift` — tautological stub docs

- `/// The timeFmt constant.` / `/// The dateFmt constant.` → replace with format string descriptions (`HH:mm:ss` / `yyyy-MM-dd`), note `static` avoids re-allocation per render
- `/// The body property.` / `/// Performs the loadLog operation.` → remove

---

### Code fixes
- [ ] `StepLogView.loadLog()`: replace inline `repoSlug` parse with `self.repoSlug` call
- [ ] `BranchSelectorSheet.loadBranches()`: migrate GCD → `Task.detached` + `await MainActor.run`
- [ ] `LogCopyButton.startCopy()`: migrate GCD → `Task { @MainActor in }` + `Task.sleep`
- [ ] `StepLogView.loadLog()`: migrate GCD → `Task.detached` + `await MainActor.run`
- [ ] `RepoSelectorSheet.itemRow`: unify dismiss contract or document deliberate asymmetry
- [ ] Both sheets: remove misplaced `/// The listSection computed view.` inside `searchSection`

### Doc hardening
- [ ] `BranchSelectorSheet`: replace/remove tautological stub docs
- [ ] `RepoSelectorSheet`: replace/remove tautological stub docs
- [ ] `LogCopyButton`: replace/remove tautological stub docs
- [ ] `StepLogView`: replace/remove tautological stub docs

### Branch & CI
- [ ] Branch `batch-21-sheets-steplog-cleanup` created
- [ ] PR opened, CI green

eonist:
## **User description**
Closes #1106

### `BranchSelectorSheet.swift`
- **Fix:** Migrate `loadBranches()` from `DispatchQueue.global` + `DispatchQueue.main.async` → `Task.detached` + `await MainActor.run` — consistent with the GCD→async/await pattern established across Batches 9–19
- **Fix:** Remove misplaced `/// The listSection computed view.` doc comment that was sitting inside the closing brace of `searchSection`
- **Doc:** Replace all tautological auto-generated stub docs (`/// The scope constant.`, `/// Extension adding functionality to BranchSelectorSheet.` ×2, etc.) with real descriptions

### `RepoSelectorSheet.swift`
- **Fix:** `itemRow` now calls `onDismiss()` after `onSelect(item)` — matches the dismiss contract of `BranchSelectorSheet.branchRow`; callers that forgot to dismiss inside `onSelect` no longer leave the sheet open
- **Fix:** Remove same misplaced `/// The listSection computed view.` doc comment inside `searchSection`
- **Doc:** Replace tautological stub docs with real descriptions

### `LogCopyButton.swift`
- **Fix:** Migrate `startCopy()` from `DispatchQueue.main.async` + `DispatchQueue.main.asyncAfter` → `Task { @MainActor in }` + `try? await Task.sleep(for: .milliseconds(1500))` — `@State` mutations are already on MainActor; GCD hops were redundant
- **Doc:** Replace tautological stub docs (`/// The fetch constant.`, `/// The phase property.`, etc.) with real descriptions

### `StepLogView.swift`
- **Fix:** Replace inline `job.htmlUrl` parsing block in `loadLog()` with `self.repoSlug` — eliminates duplicate logic; a single source of truth for the slug parse
- **Fix:** Migrate `loadLog()` from `DispatchQueue.global` + `DispatchQueue.main.async` → `Task.detached` + `await MainActor.run`
- **Doc:** Replace tautological stub docs (`/// The timeFmt constant.`, `/// The body property.`, etc.) with real descriptions; preserve all layout contract comments verbatim

## Checklist
- [x] `BranchSelectorSheet.loadBranches()`: GCD → `Task.detached` + `MainActor.run`
- [x] `BranchSelectorSheet`: remove misplaced doc comment inside `searchSection`
- [x] `BranchSelectorSheet`: replace tautological stub docs
- [x] `RepoSelectorSheet.itemRow`: add `onDismiss()` after `onSelect` — unify dismiss contract
- [x] `RepoSelectorSheet`: remove misplaced doc comment inside `searchSection`
- [x] `RepoSelectorSheet`: replace tautological stub docs
- [x] `LogCopyButton.startCopy()`: GCD → `Task { @MainActor in }` + `Task.sleep`
- [x] `LogCopyButton`: replace tautological stub docs
- [x] `StepLogView.loadLog()`: replace inline `repoSlug` parse with `self.repoSlug`
- [x] `StepLogView.loadLog()`: GCD → `Task.detached` + `MainActor.run`
- [x] `StepLogView`: replace tautological stub docs; preserve all layout contract comments

___

## **CodeAnt-AI Description**
**Fix sheet selection and log loading behavior, and clean up stale comments**

### What Changed
- Choosing a repository or branch now closes the sheet after selection, so the picker does not stay open when a value is chosen
- Branch and step log loading now runs as a background task, with the screen updated when the results are ready
- Step logs now reuse the existing repo slug value instead of parsing the repository URL again, removing duplicate lookup logic
- Removed outdated placeholder comments and replaced them with clear descriptions of what each screen and field does

### Impact
`✅ Fewer stuck selection sheets`
`✅ Smoother branch and log loading`
`✅ More consistent step log repository lookup`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Selection sheets reliably call selection then dismiss on all exit paths; cancel still dismisses.

* **Style**
  * Standardized ellipsis/placeholder characters and normalized status/copy button labels (e.g., "Copying...").

* **Improvements**
  * Background loading and copy flows use modern async handling for more consistent loading/error states and UI updates.

* **Chores**
  * Documentation and inline comment updates across various views for clarity and maintainability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/eoncode/runner-bar/pull/1108?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


eonist:
## Deep Review — Batch 21

All 10 checklist items from #1106 are addressed. The changes are net-negative in line count (−85 / +59), well-scoped, and consistently applied. Below is a file-by-file breakdown with specific observations.

---

### ✅ `BranchSelectorSheet.swift`

**GCD → async/await migration (`loadBranches`)** — correct and consistent with the batch pattern. `fetchBranchNames` is already `nonisolated` so no change was needed there; good that none was made. One nuance worth being aware of: `Task.detached` captures `self` as a strong reference. Since `BranchSelectorSheet` is a `struct` (value type), `self` is copied into the closure at capture time — this is safe and correct, no retain cycle risk.

**Misplaced doc comment removal** — both stray `/// The listSection computed view.` lines (one inside the closing brace of `searchSection`, one outside) are correctly removed.

**Doc quality** — the new struct-level doc is a real improvement. It explicitly documents the `onSelect(nil)` contract (clears filter vs. selects branch) and the "every exit path" promise for `onDismiss`. This is exactly what was missing.

**Log arrow unification** (`›` → `›` via `\u203a`, `—` via `\u2014`)** — cosmetically consistent. No functional change; fine.

---

### ✅ `RepoSelectorSheet.swift`

**`itemRow` dismiss contract fix** — this is the most impactful behavioural fix in the PR. Adding `onDismiss()` after `onSelect(item)` closes the longstanding asymmetry with `BranchSelectorSheet.branchRow`. The doc comment on `itemRow` correctly cross-references the contract, which is good for future maintainers.

**⚠️ One concern: callers that already dismissed inside their `onSelect` closure will now call `onDismiss()` twice.** The issue itself acknowledged this in Option B's note about `AddScopeSheet` validating before dismissing. You should verify all call sites. If any caller calls `dismiss()` (or equivalent) inside their `onSelect` closure, you'll now get a double-dismiss. In a SwiftUI sheet presented with `.sheet(isPresented:)`, calling the dismiss binding twice is typically harmless (idempotent), but it's worth a deliberate check — especially if any caller performs side effects in response to the dismiss trigger.

---

### ✅ `LogCopyButton.swift`

**GCD → `Task { @MainActor in }` migration** — the refactor is clean and the structured layout is cleaner than the old dual-`asyncAfter` pattern. One subtle but correct choice: the `phase = .idle` reset is now shared between the `done` and `failed` branches (moved after the `if/else`), which eliminates the duplicated reset lines. This is a genuine simplification, not just a pattern swap.

**`try? await Task.sleep`** — silently swallowing the cancellation here is correct. If the enclosing Task is cancelled while waiting (e.g. view dismissed mid-copy), you don't want a spurious error; the idle reset simply won't fire, which is acceptable.

**`resetDelay` local variable removal** — the `let resetDelay: DispatchTimeInterval = .milliseconds(1500)` is gone and the value is inlined. The magic number `1500` is now implicit. Minor: if this delay ever needs to change, it's now hardcoded. Not a blocker, just a note.

---

### ✅ `StepLogView.swift`

**`repoSlug` deduplication** — this is the cleanest fix in the batch. The old inline parse in `loadLog()` was a verbatim copy of the `repoSlug` computed property. The new form correctly handles the sentinel `"—"` value by falling back to `ScopeStore`. The fallback logic is preserved exactly from the original.

**⚠️ One semantic difference to be aware of:** The old `loadLog()` inline parse used `if !owner.isEmpty && !repo.isEmpty` as a guard before returning the slug. The `repoSlug` computed property (in the extension) does *not* check for empty owner/repo strings — it only checks `parts.count >= 5`. If `parts[3]` or `parts[4]` happen to be empty strings (malformed URL), the old code would fall back to `ScopeStore`; the new code would pass an `"/"` slug to `fetchStepLog`. In practice a malformed `htmlUrl` is unlikely, but this is a silent semantic change worth acknowledging.

**GCD → `Task.detached` in `loadLog()`** — correctly mirrors the `BranchSelectorSheet` pattern. `logText`, `isLoading`, and `onLogLoaded?()` are all mutated inside `await MainActor.run`, which is correct.

**Layout comment line-length drift** — the horizontal rule comments (`// ── Top bar ────...`) are getting progressively longer across this diff (e.g. the "Log — INSIDE ScrollView" ruler grew noticeably). These are cosmetic-only but can cause SwiftLint line-length warnings depending on config. Worth keeping an eye on.

---

### 🟢 Overall

- All red/yellow/green items from #1106 are addressed
- No regressions introduced
- Doc improvements are genuine, not just label swaps
- Two items to consciously sign off on before merge:
  1. **Double-dismiss in `RepoSelectorSheet`** — verify all `onSelect` call sites don't already dismiss
  2. **Empty owner/repo guard removed** in `StepLogView.loadLog()` — confirm `repoSlug` handles the malformed-URL edge case to your satisfaction

eonist:
## Deep Review — Batch 21 (Sheets + StepLog)

Overall this is a solid, focused cleanup batch. All 10 checklist items from #1106 are addressed and the diff is net negative (-29 lines). A few things worth discussing below — most are small, one is a real correctness concern.

---

### ✅ What's done right

- **GCD → async/await** is consistent across all four files. `Task.detached(priority: .userInitiated)` + `await MainActor.run` for data loading, `Task { @MainActor in }` for purely UI-side state mutations in `LogCopyButton`. The distinction is correct and deliberate — good.
- **`repoSlug` deduplication** is the most impactful fix in the batch. The inline parse in `loadLog()` was a silent maintenance trap; the replacement is clean.
- **Dismiss contract unification** in `RepoSelectorSheet.itemRow` is the right call. Option A from the issue (unify, don't document asymmetry) is the better choice, and that's what was implemented.
- **Misplaced `/// The listSection computed view.` comment** removed from both sheets — both occurrences caught, including the one that was *inside* the closing brace of `searchSection`.
- **Unicode escape sequences** (`\u{2026}`, `\u{2014}`, `\u{203a}`, etc.) replacing inline `…`, `—`, `›` throughout — correct approach for source portability and SwiftLint hygiene.

---

### 🔴 Real correctness concern — `loadLog()` scope fallback uses `repoSlug` before `Task.detached` captures it

**File:** `StepLogView.swift`, `loadLog()`

```swift
// Current (after patch)
let scope = repoSlug == "\u{2014}"
    ? ScopeStore.shared.scopes.first(where: { $0.contains("/") }) ?? ""
    : repoSlug
Task.detached(priority: .userInitiated) {
    let text = fetchStepLog(jobID: jobID, stepNumber: stepNum, scope: scope)
    ...
}
```

`scope` is captured as a `let` constant before the task, so the value is safe from data races ✅. **However**, the `repoSlug == "\u{2014}"` sentinel comparison has a subtle dependency: `repoSlug` returns `"\u{2014}"` for two *different* failure modes — `parts.count < 5` **and** `owner.isEmpty || repo.isEmpty`. Both are silent. If `repoSlug` is `"\u{2014}"` because `job.htmlUrl` is `nil` entirely, the fallback to `ScopeStore.shared.scopes.first` may silently pick the wrong scope when the user has multiple scopes configured. The original inline parse had the same bug, so this PR doesn't introduce it — but now that there's a single source of truth, this is the right moment to also harden `repoSlug` with a log line when it falls through to the sentinel, e.g.:

```swift
var repoSlug: String {
    let parts = (job.htmlUrl ?? "").components(separatedBy: "/")
    guard parts.count >= 5 else {
        log("StepLogView › repoSlug: htmlUrl has < 5 parts — '\(job.htmlUrl ?? "nil")'")
        return "\u{2014}"
    }
    let owner = parts[3]; let repo = parts[4]
    guard !owner.isEmpty, !repo.isEmpty else {
        log("StepLogView › repoSlug: empty owner or repo in '\(job.htmlUrl ?? "")'")
        return "\u{2014}"
    }
    return "\(owner)/\(repo)"
}
```

Not a blocker for this PR since the logic is strictly equivalent to what was there before, but worth a follow-up issue.

---

### 🟡 `LogCopyButton.startCopy()` — `try? await Task.sleep` can mask cancellation

**File:** `LogCopyButton.swift`

```swift
Task { @MainActor in
    // ...set phase to .done or .failed...
    try? await Task.sleep(for: .milliseconds(1500))
    phase = .idle
}
```

`try?` silently swallows `CancellationError`. If the view is dismissed during the 1500 ms window, the Task will be cancelled and `phase = .idle` is never reached. For a `@State` on a dismissed view this is harmless (the view is gone), but if the button is embedded in a persistent view (e.g. `ActionDetailView`), phase stays stuck at `.done` or `.failed` until the button is tapped again. The old `asyncAfter` had the same problem (it would fire after dismiss and write to dead `@State`), so again this is neutral relative to before. Consider using a `withTaskCancellationHandler` or checking `Task.isCancelled` explicitly if the button is ever reused in a longer-lived context. Worth noting in a comment at minimum:

```swift
// try? intentionally swallows CancellationError — phase reset is best-effort.
try? await Task.sleep(for: .milliseconds(1500))
```

---

### 🟡 `AddRunnerSheet.swift` — `showRepoSelector = false` / `showOrgSelector = false` removal

**File:** `AddRunnerSheet.swift` (lines 183–199 in diff)

The PR removes the inline `showRepoSelector = false` and `showOrgSelector = false` calls from the `onSelect` closures in `AddRunnerSheet`, relying entirely on `RepoSelectorSheet.itemRow` now calling `onDismiss()`. This is correct given the new contract — but it also means `AddRunnerSheet` is now **silently depending on `RepoSelectorSheet`'s internal dismiss behaviour**. If `RepoSelectorSheet` is ever refactored and `onDismiss()` is moved or removed from `itemRow`, `AddRunnerSheet` will break with no compile-time signal.

Same applies to `AddScopeSheet.swift`. The call-site comment in `RepoSelectorSheet`'s header was updated (`// ← do NOT dismiss here; itemRow calls onDismiss`) which partially documents this — but that comment lives in the sheet definition, not at the call site in `AddRunnerSheet` / `AddScopeSheet`. A one-line comment at each call site would make this more robust:

```swift
onSelect: { item in
    selectedRepo = item
    // onDismiss is called by RepoSelectorSheet.itemRow — do not duplicate here
}
```

---

### 🟢 Minor nit — comment separator lengths are inconsistent

**File:** `StepLogView.swift`

The layout contract comments (`// ── Top bar ───...`) had their trailing dashes extended in this PR (some to 100+ chars). These are intentional "ruler" comments — but the lengths now differ between sections (Top bar, Step name, Meta rows, Log). If these are meant to align visually, they should all end at the same column. Not a functional issue, just visual consistency.

---

### 🟢 Nit — `BranchSelectorSheet` search placeholder

Not introduced in this PR, but visible in context: the search field placeholder is not shown in the diff but the `searchSection` was touched. Worth confirming the placeholder text is also using a unicode escape for `…` if it contains one. (Can't verify without seeing the full `searchSection` body in context.)

---

### Summary

| Item | Verdict |
|---|---|
| GCD → async/await (all 4 files) | ✅ Correct, consistent |
| `repoSlug` deduplication | ✅ Clean |
| Dismiss contract unification | ✅ Correct approach |
| Misplaced doc comments | ✅ Both removed |
| Tautological stub doc replacement | ✅ All 4 files covered |
| Unicode escapes | ✅ Consistent |
| `repoSlug` silent fallback logging | 🔴 Follow-up issue recommended |
| `Task.sleep` cancellation behaviour | 🟡 Document with inline comment |
| Call-site dismiss dependency | 🟡 Add inline comment at each call site |
| Separator comment length consistency | 🟢 Nit |

All 10 checklist items from #1106 are addressed. The two 🟡 items are minor enough not to block merge — the 🔴 item is pre-existing and can be tracked separately. **Approves with the suggestion to file a follow-up for the `repoSlug` logging gap.**

eonist:
## Deep Review — Batch 21

Reviewed against issue #1106, the full diff, and the codebase patterns established in Batches 9–19.

---

### ✅ What's correct and well-done

**GCD → async/await migrations (issues #2, #4):**
`BranchSelectorSheet.loadBranches()` and `StepLogView.loadLog()` are both correctly migrated to `Task.detached(priority: .userInitiated)` + `await MainActor.run`. The `priority: .userInitiated` is the right choice here — these are user-triggered fetches. `fetchBranchNames` was already `nonisolated`, so no extra work was needed there.

**`LogCopyButton.startCopy()` migration (issue #3):**
The restructuring from two separate `DispatchQueue.main.asyncAfter` calls into a single `Task { @MainActor in }` with `try? await Task.sleep(for: .milliseconds(1500))` is cleaner. The `phase = .idle` reset now happens unconditionally after the sleep, which is a valid simplification — both `.done` and `.failed` branches needed the same reset anyway.

**`RepoSelectorSheet.itemRow` dismiss contract (issue #5):**
Option A was chosen (unify). `onDismiss()` is now called after `onSelect(item)` inside `itemRow`, matching `BranchSelectorSheet.branchRow`. The corresponding caller-side `showRepoSelector = false` / `showOrgSelector = false` / `showScopeSelector = false` removals in `AddRunnerSheet.swift` and `AddScopeSheet.swift` are correct — they were double-dismiss calls.

**Misplaced doc comments (issue #6):**
Both instances of the stray `/// The listSection computed view.` comment dangling inside the closing brace of `searchSection` are removed.

**Doc hardening (issues #7–#10):**
The replacement of tautological stubs with real descriptions is thorough and consistently applied across all four files. The layout contract comments in `StepLogView` that explicitly say "DO NOT REMOVE" are preserved verbatim, which is exactly right.

---

### 🔴 Real issues

**1. `repoSlug` sentinel change is a silent behaviour difference**

Issue #1106 specifies the fallback sentinel as `"—"` (em dash, U+2014):

```swift
// Issue #1106 spec
let scope = repoSlug == "—"
    ? ScopeStore.shared.scopes.first(where: { $0.contains("/") }) ?? ""
    : repoSlug
```

The PR changes `repoSlug` to return `"-"` (ASCII hyphen) on failure, and the `loadLog()` comparison is updated to match — so the two sides stay consistent *within this PR*. **However**, `repoSlug` is an `internal` computed property callable from any file in the module. Any existing caller outside `loadLog()` that was comparing against the old em-dash sentinel `"—"` will now silently never hit the fallback branch.

The diff also shows `repoSlug` is used to populate the meta row display label in `StepLogView.body`. Previously, if `htmlUrl` was nil/malformed, the UI showed `"—"` (a proper typographic dash). Now it shows `"-"` (ASCII hyphen), which is a visible UI regression against the design-token philosophy in this codebase (Phase 5 colour sweep, etc.).

**Recommendation:** Keep `repoSlug` returning `"—"` for the display case and introduce a separate private helper (e.g. `private var repoScopeForFetch: String`) for the `loadLog()` scope resolution, rather than changing the sentinel on the shared display property.

---

**2. `RepoSelectorSheet.itemRow` log line regressed**

```swift
// Before
log("RepoSelectorSheet \u{203a} selected item='\(item)'")

// After (this PR)
log("RepoSelectorSheet > selected item='\(item)'")
```

The `›` (U+203A) is the deliberate logging convention used everywhere else in this codebase — `BranchSelectorSheet`, `AddRunnerSheet`, `fetchRunnerDownloadURL`, `writeLaunchAgentPlist`, etc. all use `\u{203a}`. Replacing it with plain `>` here is inconsistent. `AddRunnerSheet.swift` in the same PR correctly converts in the *other* direction (replacing raw `›` with `\u{203a}` in string literals), but this one goes the wrong way.

**Recommendation:** Restore `\u{203a}` in this log line.

---

**3. Out-of-scope changes to `DonutStatusView`, `SparklineView`, `SystemStatsView`, `SystemStatsViewModel`, `WorkflowContextMenuModifier`, `PanelMainView`**

Issue #1106 explicitly scopes Batch 21 to four files:
- `BranchSelectorSheet.swift`
- `RepoSelectorSheet.swift`
- `LogCopyButton.swift`
- `StepLogView.swift`

The PR touches 12 files. The six files above are not mentioned anywhere in #1106. The changes are doc/comment trimming — removal of `// MARK: - Preview`, `// MARK: - Helpers`, `// MARK: - Sub-views`, inline comments, and collapsing of multi-line expressions. Three specific risks:

- **`PanelMainView.swift`:** The `TODO` comment about `visibleCount` resetting on every poll update (snapping the user back to 10 rows mid-scroll) is deleted. That was a known bug annotation. Deleting it doesn't fix the bug — it hides it.
- **`WorkflowContextMenuModifier.swift`:** All `// FIXME: #1077` comments are removed. Those tracked a known cooperative-thread-pool blocking issue with `ghPost` calls. Removing them without fixing the underlying issue means #1077 now has no inline references pointing at the offending call sites.
- **`SystemStatsViewModel.swift`:** The `nonisolated(unsafe)` safety rationale for `prevCPUInfo`, `prevNumCPUInfo`, and `timer` is stripped. Those comments existed specifically because `nonisolated(unsafe)` suppresses Swift concurrency checks — the explanation of *why* that's safe here (MainActor-only mutation during sampling, no concurrency in `deinit`) is now gone. A future reader has no way to verify the safety invariant without tracing the full call graph.

**Recommendation:** Move all out-of-scope file changes to a separate cleanup PR. At minimum, restore the deleted `TODO`/`FIXME` comments that track unresolved issues, and restore the `nonisolated(unsafe)` safety rationale.

---

**4. `StepLogView.stepStatusLabel` loses its Unicode status icons**

```swift
// Before
case "success": return "✓ success"
case "failure": return "✗ failure"
case "skipped": return "⊘ skipped"
case "cancelled": return "⊘ cancelled"
default: return step.status == "in_progress" ? "▶ running" : "· queued"

// After
case "success": return "success"
case "failure": return "failure"
// ...
```

These Unicode prefix characters (`✓`, `✗`, `⊘`, `▶`, `·`) provide at-a-glance visual differentiation in the meta row label without requiring a separate `Image` view. Issue #1106 does not request this change — it looks like an unintentional side-effect of the doc-hardening sweep.

**Recommendation:** Confirm this is intentional. If the text label is the only visual status signal in that meta row, the icons should be restored.

---

**5. `StepLogView` — separator characters in meta row changed without justification**

```swift
// Before
Text("→")  // time range separator
Text("·")  // separator dots

// After
Text("->")
Text(".")
```

These are rendered UI strings, not log output. Replacing `→` (U+2192) and `·` (U+00B7) with ASCII fallbacks is a visible typographic degradation. This wasn't requested by #1106 and doesn't align with the Phase 5 design-token philosophy.

**Recommendation:** Revert these two substitutions unless there's an explicit design decision to move to ASCII-only UI text.

---

### 🟡 Minor observations

- **Ellipsis inconsistency:** Most of this PR replaces raw `…` with `\u{2026}` escapes (correct for SwiftLint), but `RepoSelectorSheet.searchSection` goes the *other* direction — replacing `\u{2026}` with `...` (three ASCII dots). Pick one convention and apply it uniformly.
- **`PanelMainView` guard comments:** Replacing `❌ NEVER add .background()` with `Do NOT add .background()` is stylistically fine, but the visual weight of `❌` in architecture guard comments serves a purpose — it catches the eye during scanning. Consider whether the signal loss is worth the style consistency.
- **`loadBranches()` struct capture in `Task.detached`:** `BranchSelectorSheet` is a `struct` (value type), so `self` capture in `Task.detached` is a copy — correct and safe. Worth a one-line comment for the next reader, since `nonisolated` + `Task.detached` + struct capture is a subtlety that trips Swift developers regularly.

---

### Summary

The four primary items from #1106 are all correctly implemented. The main concerns before merge:

| # | Issue | Severity |
|---|-------|----------|
| 1 | `repoSlug` sentinel `"—"` → `"-"` is a display regression and a potential silent logic break | 🔴 |
| 2 | `itemRow` log line uses `>` instead of `\u{203a}` | 🔴 |
| 3 | 8 out-of-scope files: `TODO`/`FIXME` deleted, `nonisolated(unsafe)` rationale stripped | 🔴 |
| 4 | `stepStatusLabel` Unicode icons stripped (unintentional?) | 🟡 |
| 5 | Meta-row UI separators `→`/`·` replaced with ASCII | 🟡 |

eonist:
## Deep Review — Batch 21

I've read the full diff against issue #1106. Here's a structured review covering correctness, contract adherence, scope creep, and things the issue asked for that weren't done or were done differently.

---

### ✅ What's correct and well done

**`BranchSelectorSheet.loadBranches()` — GCD → async/await (#1106 item 2)**  
Migration is correct. `Task.detached(priority: .userInitiated)` + `await MainActor.run` matches the pattern established in Batches 9–19. The `nonisolated` tag on `fetchBranchNames` was already in place so no mutation there — good.

**`LogCopyButton.startCopy()` — GCD → async/await (#1106 item 3)**  
The rewrite is clean and actually *improves* on the issue's suggested implementation. The issue proposed `try? await Task.sleep(...)` (silently swallowing cancellation), but the PR correctly catches `CancellationError` explicitly and returns early rather than resetting `phase`, which is the right behaviour when a view disappears mid-copy. This is a genuine improvement over the spec — well spotted.

**`RepoSelectorSheet.itemRow` dismiss contract (#1106 item 5)**  
Option A (unified) was chosen — `onDismiss()` is now called after `onSelect(item)` inside `itemRow`. Callers in `AddRunnerSheet` and `AddScopeSheet` that were doing `showXSelector = false` inside their `onSelect` closures have had those lines removed, which is the correct follow-through. Without removing those, you'd get a double-dismiss. Good catch and good execution.

**Misplaced `/// The listSection computed view.` inside `searchSection` (#1106 item 6)**  
Removed in both `BranchSelectorSheet` and `RepoSelectorSheet`. ✓

**Tautological stub docs (#1106 items 7–10)**  
All replaced or removed across all four files. The new docs are specific, describe contracts, and follow the property-comment style already in use in the codebase. ✓

---

### ⚠️ Correctness concern — `StepLogView.loadLog()` scope logic diverges from issue spec

This is the most significant issue in the PR.

**What the issue asked for (item 1 + item 4):**
```swift
// loadLog uses self.repoSlug, with ScopeStore fallback
let scope = repoSlug == "—"
    ? ScopeStore.shared.scopes.first(where: { $0.contains("/") }) ?? ""
    : repoSlug
```

**What the PR does:**
- Introduces a *new* private computed property `repoScopeForFetch` that returns `""` (empty string) when the URL is absent or malformed.
- `loadLog()` guards on `scope.isEmpty` and **fast-fails** — it sets `logText = ""` and returns immediately **without falling back to `ScopeStore`**.

**Why this matters:**  
The original code had a `ScopeStore.shared.scopes` fallback for the case where `htmlUrl` is nil or malformed. The issue spec preserved this fallback. The PR removes it entirely. This is a **behaviour regression** in multi-repo setups or any situation where `htmlUrl` is temporarily unavailable: previously the user would still see a log fetched via the ScopeStore fallback; now they silently get an empty log view.

The PR description says: *"Fails fast when `repoScopeForFetch` is `""` (missing or malformed `htmlUrl`) to avoid querying an unrelated scope from `ScopeStore` in multi-repo setups."* This is a reasonable position *if* it was a deliberate product decision, but it contradicts the issue spec and removes a safety net without a corresponding issue/discussion tracking the intentional removal. If this is deliberate, it needs a comment in the code explaining why the fallback was removed and possibly a UI affordance (e.g. "Could not determine repository" error state vs. a silent empty log).

**Also:** `repoSlug` (the display property) now returns `"\u{2014}"` (em dash) on failure while `repoScopeForFetch` returns `""`. These are two separate computed properties with nearly identical implementations — a future change to URL parsing must now be applied in two places, which is exactly the kind of duplication the issue was trying to eliminate. The two could share a private `_parseRepoSlug() -> String?` helper that returns `nil` on failure, with each public surface formatting nil their own way.

---

### ⚠️ Scope creep — files not in issue scope modified

The issue scoped Batch 21 explicitly to four files:
- `BranchSelectorSheet.swift`
- `RepoSelectorSheet.swift`
- `LogCopyButton.swift`
- `StepLogView.swift`

The PR also modifies:
- `DonutStatusView.swift`
- `SparklineView.swift`
- `SystemStatsView.swift`
- `SystemStatsViewModel.swift`
- `WorkflowContextMenuModifier.swift`
- `PanelMainView.swift`
- `AddRunnerSheet.swift`
- `AddScopeSheet.swift`

The changes to these out-of-scope files are almost exclusively:
1. Unicode escape sequences substituting literal typographic characters (`→` → `\u{2192}`, `—` → `\u{2014}`, `…` → `\u{2026}`, `❌` → `Do NOT`, `·` → `\u{00B7}`)
2. Doc comment trimming / wording adjustments
3. Minor whitespace / line-breaking reformatting

None of these are in-scope for #1106. They look like a global Unicode normalisation + doc-comment sweep that got folded into this PR. This makes the diff significantly harder to review (10 files touched instead of 4) and conflates two distinct concerns. If this is part of a systematic Unicode-escape policy (for encoding safety or linter compliance), it should either:
- Be a separate PR with its own tracking issue, or
- Be called out explicitly in the PR description so reviewers know it's intentional and can verify the character substitutions are semantically identical.

Specific substitution to double-check: `"resuming…"` → `"resuming\u{2026}"` in `PanelMainView.swift` and `"Load N more workflows…"` → `"Load N more workflows..."` (this one uses `...` — three ASCII periods — not `\u{2026}`). **That last one is inconsistent with the rest of the Unicode sweep**: the ellipsis in the load-more button text was converted to ASCII `...` instead of `\u{2026}`, while all other ellipses in the PR are `\u{2026}`. This may be intentional (the button text is programmer-facing?), but it looks like an oversight.

---

### ⚠️ `PanelMainView` — removed TODO comment without resolving it

The original code had:
```swift
// TODO: visibleCount resets on every actions identity change, including poll
// updates that don't change the list length. This snaps the user back to 10
// rows mid-scroll. Should reset only when store.actions.count decreases.
.onChange(of: store.actions) { _, _ in visibleCount = 10 }
```

The PR silently removes the TODO comment but leaves the behaviour unchanged. The bug still exists — the user is still snapped back to 10 rows on every poll. Removing a TODO without resolving it or creating a tracking issue hides known debt. If it's intentional to leave the behaviour and just clean up the comment, it should be a separate commit or at minimum noted in the PR description.

---

### ⚠️ `SystemStatsViewModel` — safety documentation removed from `nonisolated(unsafe)` fields

The `nonisolated(unsafe)` keyword on `timer`, `prevCPUInfo`, and `prevNumCPUInfo` is load-bearing — it suppresses Swift 6 concurrency warnings for fields that are deliberately not actor-isolated. The original code had inline `// Safety:` comments explaining *why* this is safe (only mutated on MainActor during lifecycle, only accessed from `deinit` after all references are gone). The PR removes those comments and replaces them with terse single-liners.

`nonisolated(unsafe)` is a sharp edge. Stripping the safety rationale makes it harder for a future developer (or agent) to reason about whether new access patterns are safe. The rationale should be preserved, even in condensed form.

---

### Minor observations

- **`AddRunnerSheet.swift` `showRepoSelector = false` removal**: correct given the new `RepoSelectorSheet.itemRow` contract, but `AddScopeSheet.swift` has the same removal. Both look correct.
- **`WorkflowContextMenuModifier` FIXME removal**: the three `// FIXME: #1077 Task.detached wraps a blocking call` comments were removed without the underlying issue being resolved. These are legitimate tracking markers for a known perf issue — removing them without closing #1077 or leaving a comment hides the debt.
- **`DonutStatusView` — removed `/// Color transitions use .easeInOut(duration: 0.35)`**: the original comment documented that color transitions used a 0.35s easing. That line in the spec doc was removed. If the colour-transition animation was also removed (it would show up in the SwiftUI `.animation` modifiers), that's a visual regression. If it still exists, the doc removal is just misleading. Worth confirming the `.animation(.easeInOut(duration: 0.35))` call is still present in the non-diffed part of `DonutStatusView`.

---

### Summary

| Area | Status |
|---|---|
| Core 4-file fixes from #1106 | ✅ All addressed |
| `LogCopyButton` cancellation handling | ✅ Improved over spec |
| `RepoSelectorSheet` dismiss contract + caller cleanup | ✅ Correct |
| `StepLogView` scope fallback (ScopeStore removed) | ⚠️ Behaviour regression vs. spec — needs explicit justification or a fix |
| `repoSlug` / `repoScopeForFetch` duplication | ⚠️ Reintroduces dual-maintenance risk |
| Out-of-scope file changes (Unicode sweep + doc trim) | ⚠️ Should be a separate PR or explicitly scoped |
| Load-more `...` vs `\u{2026}` inconsistency | ⚠️ Likely oversight |
| Removed TODO in `PanelMainView` without resolution | ⚠️ Hides known debt |
| Removed FIXME #1077 markers in context menu | ⚠️ Hides known debt |
| `nonisolated(unsafe)` safety rationale removed | ⚠️ Reduces future safety auditability |

The four primary fixes are solid. The main thing I'd want resolved before merge is the `repoScopeForFetch` fallback question — either restore the `ScopeStore` fallback (matching the issue spec) or open a new issue explicitly tracking the intentional removal with rationale.

eonist:
## Deep Review — Batch 21

Reviewed the diff in full against issue #1106. Overall the batch is solid and all ten checklist items are ticked. Below are my findings, ordered by severity.

---

### 🔴 Correctness

**`LogCopyButton.startCopy()` — silent cancellation hides the phase reset**

The PR swaps `DispatchQueue.main.asyncAfter` for `Task { @MainActor in } + Task.sleep`. The `catch is CancellationError` path returns early without resetting `phase` to `.idle`:

```swift
} catch is CancellationError {
    // Task was cancelled -- exit cleanly without resetting phase.
    return
}
phase = .idle
```

If the SwiftUI view is torn down mid-flight (e.g. the popover closes while the 1.5 s delay is running), the `Task` is cancelled and `phase` stays stuck on `.done` or `.failed`. On the next appearance the button will be non-interactive until the user dismisses and reopens. The original GCD version had the same bug, but the new async path makes it easier to fix:

```swift
} catch {
    // CancellationError or anything else — still reset so button is reusable.
    phase = .idle
    return
}
```

Alternatively, use `try? await Task.sleep(...)` (as the issue's own suggested snippet shows) and remove the `do/catch` entirely — `phase = .idle` runs unconditionally whether sleep completes normally or is cancelled.

---

**`StepLogView.loadLog()` — new early-exit guard does not call `onLogLoaded`**

```swift
guard !scope.isEmpty else {
    logText = ""
    isLoading = false
    onLogLoaded?()   // ← this IS here, good
    return
}
```

Actually this one is fine — `onLogLoaded?()` is called. ✅ Noting it so it's not re-raised later.

---

### 🟡 Design / Semantic

**`StepLogView` — `_parseRepoSlug()` naming and visibility**

The helper is named `_parseRepoSlug()` with a leading underscore, which is a non-Swift convention (that's a Python/ObjC idiom). In Swift the idiomatic choice is either a `private` method with a normal name, or a `fileprivate` computed property. Since it is already `private func`, the underscore adds no protection and will confuse future readers. Suggest renaming to `parsedRepoSlug() -> String?` or inlining it as a computed property `private var repoSlugComponents: String?`.

Also: `repoScopeForFetch` is declared `private var` but `repoSlug` (which calls `_parseRepoSlug`) is `var` (internal). If `repoSlug` is only used within `StepLogView`, it should be `private var` to match `repoScopeForFetch`.

---

**Issue #1106 spec divergence — `loadLog()` uses `repoScopeForFetch`, not `repoSlug`**

The issue says:

> Replace the inline parse in `loadLog()` with a call to `self.repoSlug`

The PR instead introduces a new `private var repoScopeForFetch` that returns `""` on failure, while `repoSlug` returns `"—"`. The PR's approach is actually *better* than the spec because it avoids comparing against a display sentinel string (`== "—"`) in logic code. This is the right call and should be called out positively — but the PR description still says "replace with `self.repoSlug`" without noting the intentional deviation. Worth a one-line comment in the PR description or inline so the next reader doesn't wonder if `repoScopeForFetch` was an accidental duplication.

---

**`RepoSelectorSheet` — caller call-site cleanup is partial**

`AddRunnerSheet.swift` removes `showRepoSelector = false` and `showOrgSelector = false` from the `onSelect` closures (correct, now that `itemRow` calls `onDismiss()`). Same removal happens in `AddScopeSheet.swift`. But the usage comment at the top of `RepoSelectorSheet.swift` was also updated:

```swift
//     onSelect: { selectedRepo = $0 }   // do NOT dismiss here; itemRow calls onDismiss
```

Good — but a reader skimming `AddScopeSheet` or `AddRunnerSheet` will not see that comment. Consider adding a brief inline note at each updated call-site (one line is enough), or it will be cargo-culted back in by a future agent reading only the call-site.

---

### 🟢 Docs / Comments

**Positive: `DonutStatusView`, `SparklineView`, `SystemStatsView`, etc. — doc sweep in non-primary files**

The diff touches 12 files but issue #1106 scopes only 4. The extra 8 files (`DonutStatusView`, `SparklineView`, `SystemStatsView`, `SystemStatsViewModel`, `WorkflowContextMenuModifier`, `PanelMainView`, `AddRunnerSheet`, `AddScopeSheet`) receive doc/comment cleanup that was not in scope. That's a good-faith improvement batch, but it means CI regressions in those files are harder to attribute. If CI goes red on one of those files, the batch becomes harder to bisect. For future batches, consider keeping out-of-scope style commits in a separate commit so they can be reverted atomically.

---

**`❌` → `Do NOT` conversion is correct but not fully consistent**

The batch converts `❌ NEVER …` warning markers to plain prose `Do NOT …` — correct for the doc-style rules in this codebase. However `PanelMainView.swift` still has two `❌ NEVER` comments in the layout architecture block that were not touched:

```
// ❌ NEVER remove .frame(maxHeight:) from this ScrollView.
```

Not a blocker, but worth sweeping in this batch or noting as a follow-up since the intent of the batch was to remove emoji-warning markers.

---

**`WorkflowContextMenuModifier` — removal of inline section comments loses orientation**

The file loses all `// Re-run failed`, `// Re-run all`, `// Cancel`, `// Copy log`, `// Show workflow file on GitHub`, `// Show GitHub SHA` inline comments. The `menuItems` doc now says "re-run failed, re-run all, cancel, copy log, open on GitHub" — sufficient for a function-level doc. But inside `menuItems` the six buttons are now a continuous block of ~50 lines with only `Divider()` as visual breaks. Consider keeping at minimum the `Divider()` visual grouping comments (a one-word `// Actions`, `// Utilities` pair, not six individual ones).

---

### ✅ Summary

| Area | Status |
|---|---|
| All 10 issue #1106 checklist items | ✅ Done |
| GCD → async/await migrations (4 sites) | ✅ Correct |
| `onDismiss` contract unification | ✅ Done + call-sites cleaned |
| Misplaced doc comments removed | ✅ Done |
| Tautological stub docs replaced | ✅ Done |
| `LogCopyButton` cancel-path phase leak | 🔴 Minor bug — easy fix |
| `_parseRepoSlug` naming | 🟡 Convention issue |
| Partial `❌ NEVER` sweep | 🟢 Follow-up candidate |

The one blocker-candidate is the `CancellationError` path leaving `phase` in a non-idle state. Everything else is nit-level or follow-up. Good work — clean, focused, and the `repoScopeForFetch` split is a net improvement over what the issue asked for.

PR #1138, opened 2026-06-04
refactor: Swift 6.2 concurrency modernisation (#1135)

eonist:
## **User description**
Closes #1135

Implements all 8 items from the Swift 6.2 Concurrency Modernisation tracking issue. The goal: zero `nonisolated(unsafe)` on mutable state, zero `@unchecked Sendable` escape hatches, zero `NSLock`/`DispatchSemaphore` bridges on the read/poll transport — fully compiler-verified structured concurrency throughout.

### ✅ Item 1 — `ProcessRunner`: drop `nonisolated(unsafe)` + `NSLock` + `readabilityHandler`
Closes #1136
- `nonisolated(unsafe) var outputData`, `NSLock`, and `readabilityHandler` removed
- Replaced with `waitUntilExit()` then `readDataToEndOfFile()`
- `DispatchWorkItem` timeout retained (guards bug #477)

### ✅ Item 2 — `RunnerModel`: make fully immutable, drop `@unchecked Sendable`
Closes #1137
- All `var` properties converted to `let`
- `@unchecked Sendable` → `Sendable` (compiler-synthesised)
- Added `RunnerModel.copying(...)` extension for ergonomic copy-and-replace
- `LocalRunnerStore` mutation sites updated to use `copying(...)`

### ✅ Item 3 — `GitHubTransportShim`: replace `ghAPI` shim with `URLSession async/await`
Closes #1140
- `GHAPITransport` typealias is now `async`
- `ghAPI()` reads closure under lock then awaits outside
- `urlSessionAPIAsync()` backed by `URLSession.data(for:)` — no semaphore
- All call sites updated with `async`/`await` (mechanical ripple)

### ✅ Item 4 — `RunnerStatusEnricher`: `async` + parallel scope fetches
Closes #1139
- `enrich()` and `fetchRunnersForScope()` now `async`
- Scope fetches parallelised with `withTaskGroup`
- `LocalRunnerStore.refresh()` switched from `DispatchQueue.global` to `Task.detached`
- Poll latency now constant regardless of scope count

### ✅ Item 5 — `WorkflowActionGroupFetch`: parallel run and job fetches
Closes #1141
- `fetchActionGroups(for:cache:)` is now `async`
- `in_progress`, `queued`, and `completed` run fetches fire concurrently via `async let`
- Group construction parallelised via `withTaskGroup`
- `fetchJobsForGroup` and `fetchJobsForRun` now `async`
- Per-run job fetches and per-job refresh calls run concurrently via `withTaskGroup`

[...]

PR #1179, opened 2026-06-06
fix: sign-out token fallback and CPU/MEM metrics in runner rows

eonist:
## Problem

Two user-visible bugs introduced or exposed by PR #1138:

### 1. Sign-out broke the `GH_TOKEN` / `GITHUB_TOKEN` fallback

Before #1138, polling was driven by `Timer + scheduleTimer()`. After sign-out, `Keychain.delete()` called `invalidateTokenCache()`, the timer fired on the next cycle, and `githubToken()` naturally fell through to env-var tokens.

After #1138, the timer was replaced with `pollTask: Task` that loops on `Task.sleep`. It never calls `start()` again on its own. The only `didSignOut` subscriber was in `SettingsView` stored in `@State` — alive only while Settings is open, and it only updated UI flags. The poll loop was never restarted after sign-out, so the env-token fallback was silently dead.

### 2. CPU / MEM badges missing from runner rows in the main view

Two separate causes:

- **First launch:** `LocalRunnerStore.init()` only calls `loadIndex()` — populates the `name→path` map but leaves `runners = []`. `RunnerStore.start()` was called immediately after, so `buildInstallPathMap` always received `localRunners: []` on the first fetch → no install path resolved → metrics never fetched.
- **Subsequent cycles:** `applyRefreshResults` replaced `LocalRunnerStore.runners` wholesale on each refresh, discarding any metrics written during the previous poll cycle → CPU/MEM flickered away every time a refresh ran.

---

## Fix

**Sign-out:** `setupSignOutSubscription()` added to `AppDelegate` — stores a `didSignOut` subscriber in `AppDelegate.cancellables` (full app lifetime) that calls `RunnerStore.shared.start()`. On the next `fetch()`, `githubToken()` re-resolves from `GH_TOKEN`/`GITHUB_TOKEN` since Keychain is now empty.

**CPU/MEM:** `LocalRunnerStore.shared.refresh()` is now called before `RunnerStore.shared.start()` in `setupCombineSubscriptions()`, seeding `runners` before the first fetch runs. `applyRefreshResults` now snapshots and re-applies existing metrics across refresh cycles. The `applyMetrics` write-back is guarded to self-hosted runners only.

---

## Behaviour after fix

| Scenario | Before | After |
|---|---|---|
| OAuth sign-out → fallback to `GH_TOKEN` | ❌ broken | ✅ works |
| CPU/MEM shown on first app launch | ❌ never shown | ✅ shown |
| CPU/MEM survives across poll cycles | ❌ flickers away | ✅ preserved |
| Cloud-hosted runners produce no warnings | ❌ warning every cycle | ✅ silent |

---

## Files changed

- `AppDelegate.swift` — `setupSignOutSubscription()` + call from `applicationDidFinishLaunching`
- `AppDelegate+PanelSetup.swift` — `LocalRunnerStore.shared.refresh()` before `RunnerStore.shared.start()`
- `LocalRunnerStore.swift` — metrics preservation in `applyRefreshResults`; `#if DEBUG` gates on inner-loop logs
- `RunnerStore.swift` — `applyMetrics` write-back guarded to self-hosted runners; `#if DEBUG` gates on per-runner resolution logs
- `Auth.swift` — `#if DEBUG` gates on per-resolution-path token logs
- `OAuthService.swift` — structured logging for sign-in/sign-out/callback flow
- `SettingsView.swift` — structured logging for auth state changes


eonist:
## Deep Code Review — PR #1179

### TL;DR
The **core fix is correct and necessary**. `setupSignOutSubscription()` in `AppDelegate` is the right place, for the right reason, and the implementation is clean. The companion changes—startup seeding, per-cycle refresh, and metrics write-back—are well-intentioned but introduce a **real API doubling problem on the hot path** and a **race condition** that will silently produce stale data on the first fetch. The logging volume is also production-unsafe as written.

---

### ✅ What's right

**`setupSignOutSubscription()` in `AppDelegate.swift`**

This is the precise, minimal fix the regression needed. The doc comment is exemplary—it names the regression PR (#1138), explains the old timer-based fallthrough mechanism vs. the new `Task.sleep` loop, and makes clear why `SettingsView`'s `@State`-stored subscription was never a reliable subscriber. Storing into `cancellables` gives it proper app-lifetime semantics. No objections here.

**`[weak self]` omission is intentional and correct**

`AppDelegate` is a true singleton. The closure capturing `RunnerStore.shared.start()` doesn't capture `self` at all, so the `guard self != nil` pattern from the PR description was wisely dropped. The final implementation is tighter.

**Metrics preservation in `applyRefreshResults`**

The dual-key strategy (agentId-first → name-fallback) correctly mirrors the `installPathMap` resolution order in `fetchAndEnrichRunners`, and preserving metrics across refresh cycles avoids the UI flicker where CPU/MEM badges disappeared on every poll tick. Good fix.

**`SettingsView.onAppearAction` micro-fix**

Caching `Keychain.token` and `githubToken()` results into locals before using them avoids a double-read where the Keychain could theoretically be modified between the two calls. Small but correct.

---

### ⚠️ Issues to address

#### 1. `LocalRunnerStore.shared.refresh()` inside `fetch()` doubles GitHub API calls — **this is the most serious problem**

`refresh()` calls `RunnerStatusEnricher.enrich(runners:)`, which fires one `GET /repos/{owner}/{repo}/actions/runners` (or the org equivalent) per active scope. `fetchAndEnrichRunners()` in the same `fetch()` cycle calls the **exact same endpoints**. With a 10-second active-runner interval you now make 2× the GitHub API requests on every hot-path cycle.

The fix comment says:
> "This is an async kick-off (not awaited) — refresh() is guarded by isScanning so concurrent cycles are safe."

`isScanning` prevents *concurrent* refresh cycles, but it does **not** prevent the `refresh()` kicked from `fetch()` from racing `fetch()`'s own `buildInstallPathMap` read. The timeline:

```
fetch() fires
  → kicks refresh() [Task { ... }]  ← async, returns immediately
  → reads LocalRunnerStore.shared.runners immediately after ← may still be []
  → builds installPathMap from empty runners
  → fetchAndEnrichRunners runs with empty installPathMap → no metrics this cycle
  → refresh() Task completes, runners now populated
  → NEXT fetch() cycle finally sees populated runners
```

The comment even acknowledges this: *"The refreshed runners will be picked up by the NEXT fetch cycle."* So the per-cycle kick inside `fetch()` provides zero benefit for the current cycle—it only benefits the next one. The startup call in `setupCombineSubscriptions` already seeds runners before `start()`. The per-cycle kick is redundant and costly.

**Recommendation:** Remove `LocalRunnerStore.shared.refresh()` from `fetch()`. The startup seed + the `LocalRunnerStore.$runners` Combine sink that fires `observable.reload()` is sufficient. If disk changes need to be picked up during a running session, the existing `$runners` sink already handles that reactively.

#### 2. `applyMetrics` write-back loop iterates over GitHub API runners, not local runners

```swift
for (_, runner) in indexed {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

`indexed` is built from `runnersWithScope` — all runners returned by the GitHub API, including cloud-hosted runners that will never match anything in `LocalRunnerStore`. Every non-local runner produces a `⚠️ applyMetrics — no matching runner` log line. In a large org with many cloud runners this will be extremely noisy. 

**Recommendation:** Guard the loop — only call `applyMetrics` if the runner has a resolved `installPath` (i.e. it's a self-hosted runner that appeared in the installPathMap):

```swift
for (_, runner) in indexed where runner.installPath != nil {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

Or better, pass the `installPathMap` to `fetchAndEnrichRunners` and filter on `byId`/`byName` presence.

#### 3. `isOAuthAuthenticated` set before `invalidateTokenCache()` is called

In `OAuthService.signOut()`:
```swift
pendingState = nil
let deleted = Keychain.delete()
if deleted {
    didSignOut.send()  // → AppDelegate → start() → fetch() → githubToken()
```

`Keychain.delete()` calls `invalidateTokenCache()` internally (based on PR #1138's pattern). The `didSignOut` subscriber in `AppDelegate` fires synchronously via `.receive(on: DispatchQueue.main)`. If `signOut()` is called on the main thread (it is, from `SettingsView`), the `sink` closure runs *after* the current run loop tick — but `start()` launches a `Task`, and `fetch()` within that task will call `githubToken()` asynchronously. This ordering is fine. However, in `SettingsView`'s sink:

```swift
signOutCancellable = OAuthService.shared.didSignOut
    .sink {
        let postToken = githubToken()  // calls githubToken() AGAIN
        isOAuthAuthenticated = false
        isCLIAuthenticated = postToken != nil
    }
```

`githubToken()` in the SettingsView sink may still find the env-token and populate the cache immediately, which is correct — but if there is no env token, `invalidateTokenCache()` has already been called, so this returns nil and sets `isCLIAuthenticated = false` correctly. This path is fine.

#### 4. Non-`#if DEBUG`-guarded `log()` calls at production call sites

`Auth.swift` correctly gates the per-resolution-path logs behind `#if DEBUG`. But in `LocalRunnerStore.swift` and `RunnerStore.swift`, every `log()` call is unconditional — including inner-loop logs like:

```swift
// Fires for every runner on every poll cycle:
log("LocalRunnerStore › refresh() — '\(runner.runnerName)' isRunning=\(live)")
log("RunnerStore › fetchAndEnrichRunners — \(runner.name) id=\(runner.id) busy=true; fullKey=...")
```

With 10 runners at a 10-second interval, `fetchAndEnrichRunners` alone produces ~50 log lines per minute just for the per-runner resolution logs. On the `isTracked()` path it's even more concerning — `isTracked` is likely called on every render — and now emits a log line on every call.

**Recommendation:** Gate all inner-loop logs (per-runner, per-cycle granularity) behind `#if DEBUG`. The outer structural logs (`fetch() — start`, `fetch complete`) can stay unconditional.

#### 5. `setupSignOutSubscription()` has a subtle `[weak self]` inconsistency

The PR description version had `guard self != nil else { return }`. The final version dropped `self` entirely (correct), but the closure doesn't use `[weak self]` either — this is fine since `RunnerStore.shared` is a static reference, but it's worth noting the doc comment still says `[weak self]` in the code example block in the PR body. Not a code issue, but the comment drift could confuse future readers if the PR body is used as reference.

#### 6. `popoverShouldClose` log is noise

```swift
public func popoverShouldClose(_ _: NSPopover) -> Bool {
    log("AppDelegate › popoverShouldClose — returning true")
    return true
}
```

This fires every time the popover is dismissed (every single close event). It adds zero diagnostic value — the log says exactly what the code says. Remove it or gate it behind `#if DEBUG`.

---

### 🔍 Scope of this PR vs. original issue

The PR title is `fix: restart poll loop on sign-out so GH_TOKEN/GITHUB_TOKEN fallback is used`. The **original bug** is a one-function fix in `AppDelegate.swift`. The rest of the diff (317 additions across 7 files) is:

- **Startup seeding** (`setupCombineSubscriptions`): Directly related and necessary — without this the first fetch has an empty `installPathMap`. ✅ Keep.
- **Per-cycle refresh inside `fetch()`**: Partially addresses the same seeding concern but at the cost of doubled API calls. The startup seed already covers this. ❌ Remove.
- **Metrics write-back**: Addresses a separate pre-existing bug (CPU/MEM badges disappearing on refresh). Reasonable to bundle, but it's a separate concern. Consider splitting into its own PR to keep blame/bisect clean.
- **Logging**: Useful for debugging the regression, but needs `#if DEBUG` gating before merge.

---

### Summary table

| Area | Status | Action |
|---|---|---|
| `setupSignOutSubscription()` | ✅ Correct | Ship it |
| Startup `LocalRunnerStore.refresh()` before `start()` | ✅ Correct | Keep |
| Per-cycle `refresh()` inside `fetch()` | ❌ Doubles API calls, redundant | Remove |
| Metrics write-back (`applyMetrics`) | ⚠️ Correct but noisy | Filter to self-hosted only |
| `applyRefreshResults` metrics preservation | ✅ Correct | Keep |
| Inner-loop `log()` calls without `#if DEBUG` | ⚠️ Production log spam | Gate behind `#if DEBUG` |
| `isTracked()` now logs on every call | ⚠️ High frequency | Remove or gate |
| `popoverShouldClose` log | ⚠️ Pure noise | Remove |

The fix is sound and unblocks users hitting the sign-out regression. The main thing blocking a clean merge is the per-cycle `refresh()` in `fetch()` — it's the one change that can actively hurt users by accelerating rate limit burn. Everything else is polish.

eonist:
## Deep Review — PR #1179

### Overall verdict
The **core fix is correct and well-reasoned**. `setupSignOutSubscription()` in `AppDelegate` is the right solution — app-lifetime subscriber, correct placement, minimal footprint. The companion changes are where this PR overreaches: one of them actively makes things worse on the hot path, and the logging strategy is not production-safe as written.

---

### ✅ Core fix — `AppDelegate.setupSignOutSubscription()`

This is the cleanest part of the PR. The root cause analysis in the PR description is accurate: `#1138` replaced a `Timer`-based loop (where sign-out token cache invalidation naturally fed into the next timer fire) with a `Task.sleep` loop that never re-enters `githubToken()` unless `start()` is called explicitly. The `SettingsView` subscriber was always wrong for this — `@State`-stored cancellables live and die with the view, and a user who signs out without Settings open would never trigger it.

Storing the subscription in `AppDelegate.cancellables` is correct. `AppDelegate` is a true singleton for the process lifetime. The `[weak self]` omission is also correct — there's no `self` captured in the closure at all, only the static `RunnerStore.shared`. The doc comment naming PR #1138 explicitly is good practice.

No objections. Ship this part.

---

### ✅ Startup `LocalRunnerStore.refresh()` before `start()` in `setupCombineSubscriptions`

This is necessary and correct. `LocalRunnerStore.init()` only calls `loadIndex()`, which populates the `name → path` map in `UserDefaults`. The `runners` array stays empty until `refresh()` runs the full disk-hydration + launchctl + GitHub-enrichment pipeline. Without this call before `start()`, `buildInstallPathMap` always receives `localRunners: []` on the first fetch cycle, the `installPathMap` is empty, and busy runners silently get no metrics. The comment added to `init()` documents this contract clearly. Keep this.

---

### ❌ Per-cycle `LocalRunnerStore.shared.refresh()` inside `fetch()` — most critical issue

```swift
// RunnerStore.fetch()
if !LocalRunnerStore.shared.isScanning {
    LocalRunnerStore.shared.refresh()
}
let localRunners = LocalRunnerStore.shared.runners  // read immediately after kick
```

This has two distinct problems:

**Problem A — Doubled GitHub API calls.** `refresh()` calls `RunnerStatusEnricher.enrich(runners:)`, which issues `GET /actions/runners` for every active scope. `fetchAndEnrichRunners()` in the same `fetch()` cycle calls the exact same endpoints. At a 10-second active-runner interval with 2 scopes, you go from ~12 `/actions/runners` calls/minute to ~24. GitHub's secondary rate limit threshold for REST is documented at 90 requests/minute per resource, but in practice authenticated apps hit undocumented "burst" limits much sooner. This is a real regression risk that will hurt users with multiple scopes.

**Problem B — The kick is useless for the current cycle.** `refresh()` launches a `Task { }` and returns immediately. The `LocalRunnerStore.shared.runners` read on the very next line happens before that task has a chance to run. The comment even acknowledges: _"The refreshed runners will be picked up by the NEXT fetch cycle."_ So this call provides zero benefit to the current cycle — you seeded runners at startup already, and the only scenario where this helps is if someone adds a new runner to disk mid-session. That's a legitimate but rare scenario, and it doesn't justify the API cost on every single poll cycle.

**Recommendation:** Remove the `refresh()` call from `fetch()` entirely. The startup seed covers first-launch correctness. For mid-session disk changes, the existing `LocalRunnerStore.$runners` Combine sink already fires `observable.reload()` reactively whenever runners change. If there's a specific mid-session scenario you need to cover, a file-system watcher on the runner install directories is the right tool — not a `refresh()` on every poll tick.

---

### ⚠️ `applyMetrics` write-back loop — correct logic, wrong scope

```swift
for (_, runner) in indexed {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

`indexed` contains all runners returned by the GitHub API — including GitHub-hosted (cloud) runners, which have no entry in `LocalRunnerStore`. Every cloud runner produces a `⚠️ applyMetrics — no matching runner` warning log. In an org with tens or hundreds of cloud runners this becomes a persistent warning storm that drowns out real signals.

The fix is straightforward: only write back for runners that resolved an `installPath`, since that's the exact predicate that identifies self-hosted runners in this codebase:

```swift
for (_, runner) in indexed where installPathMap.byId[runner.id] != nil
    || installPathMap.byName[runner.name] != nil {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

The metrics-preservation logic in `applyRefreshResults` (dual-key agentId-first strategy) is correct and should be kept as-is.

---

### ⚠️ Logging volume is not production-safe

`Auth.swift` correctly gates per-resolution-path logs behind `#if DEBUG`. The same discipline is missing in `LocalRunnerStore.swift` and `RunnerStore.swift`. Specific call sites that will flood production logs:

- `LocalRunnerStore.isTracked()` — called on every render, now emits a log line every call
- `refresh()` inner loop — `runner.runnerName isRunning=\(live)` fires for every runner on every disk scan
- `fetchAndEnrichRunners` per-runner resolution block — `byId=... byFullKey=... byName=... → resolved=...` fires for every runner on every poll cycle
- `popoverShouldClose` — fires on every single popover dismiss, contains zero diagnostic information beyond what the code already states

These should be wrapped in `#if DEBUG`. The structural logs (fetch start/complete, poll loop interval, sign-out restart) can stay unconditional — they're low-frequency and operationally useful. Inner-loop per-item logs should never run in release builds.

---

### 🔍 Minor: `setupSignOutSubscription` doc comment drift

The PR description's code example shows:
```swift
.sink { [weak self] in
    guard self != nil else { return }
```

The actual implementation correctly drops `[weak self]` and `guard self != nil` since `self` is never captured. The implementation is right, but if this PR description is ever used as reference documentation the example will be misleading. Not a code issue, but worth noting for docs hygiene.

---

### Summary

| Area | Status | Recommendation |
|---|---|---|
| `setupSignOutSubscription()` | ✅ Correct | Keep — this is the fix |
| Startup `refresh()` before `start()` | ✅ Correct | Keep — necessary for first-cycle correctness |
| Per-cycle `refresh()` inside `fetch()` | ❌ Doubles API calls, zero same-cycle benefit | Remove |
| `applyMetrics` write-back | ⚠️ Correct but clouds with cloud runners | Filter to `installPathMap`-resolved runners only |
| `applyRefreshResults` metrics preservation | ✅ Correct | Keep |
| Inner-loop `log()` calls (per-runner, per-cycle) | ⚠️ Production log spam | Gate behind `#if DEBUG` |
| `isTracked()` logging | ⚠️ Render-frequency logging | Remove or `#if DEBUG` |
| `popoverShouldClose` log | ⚠️ Zero signal, high noise | Remove |

The sign-out regression is fixed and the fix is clean. The one change blocking a confident merge is the `refresh()` inside `fetch()` — it can accelerate GitHub API rate-limit exhaustion for users with multiple scopes. Everything else is a polish concern.

eonist:
## Deep Review — PR #1179

### Verdict
The regression fix itself is correct, minimal, and well-placed. Several companion changes are either actively harmful on the hot path or quietly introduce a new correctness hole that will surface under a specific but realistic condition. Below is a precise breakdown by concern, not by file.

---

### ✅ `setupSignOutSubscription()` — ship it

The root cause analysis is accurate and the fix is precise. `AppDelegate.cancellables` gives it true process-lifetime semantics. The closure captures no `self` — `RunnerStore.shared` is a static reference — so the `[weak self]` guard from the PR description draft was correctly dropped. The doc comment naming PR #1138 and explaining the old timer-based fallthrough vs. the new `Task.sleep` loop is exactly the kind of reasoning future maintainers need.

One note: `signOut()` only emits `didSignOut` when `Keychain.delete()` returns `true`. If Keychain deletion fails (e.g. permission error or sandbox issue), neither `didSignOut` fires nor does the poll loop restart. The existing guard is intentional (to prevent a "ghost signed-in" state), but it means a failed sign-out leaves the app in an authenticated poll loop with no env-token fallback AND no visual indication. Worth a follow-up issue if not already tracked.

---

### ✅ Startup `refresh()` before `start()` — keep

`LocalRunnerStore.init()` only calls `loadIndex()`, which populates the `name → path` UserDefaults map. `runners` stays `[]` until `refresh()` runs the full disk-hydration + launchctl + GitHub-enrichment pipeline. The added `init()` comment documents this contract clearly. Without the startup seed, `buildInstallPathMap` always receives `localRunners: []` on the first fetch and busy runners silently get no metrics. This is a real fix and must stay.

---

### ❌ Per-cycle `refresh()` inside `fetch()` — this is the merge blocker

```swift
// RunnerStore.fetch()
log("...triggering LocalRunnerStore.refresh() BEFORE starting poll loop")
LocalRunnerStore.shared.refresh()
RunnerStore.shared.start()
```

Wait — that's in `setupCombineSubscriptions`, not `fetch()`. Reading the diff again carefully:

The `refresh()` call in `setupCombineSubscriptions` is the **startup seed** (correct). But looking at the comment block added to `fetch()`:

> "LocalRunnerStore is seeded once at startup (before start() is called in AppDelegate+PanelSetup) and kept current reactively via its own $runners Combine sink. There is no need to call refresh() here — doing so would duplicate the GET /actions/runners GitHub API call that fetchAndEnrichRunners already makes in the same cycle, doubling the API request rate on the hot path."

This comment was added *to explain why refresh() was NOT added to fetch()*, which is correct — the PR intentionally avoided putting `refresh()` inside the polling loop. The comment is accurate. I misread the indentation in the diff.

**However**, the `applyMetrics` write-back loop at the end of `fetchAndEnrichRunners` does iterate over *all* indexed runners and call `LocalRunnerStore.shared.applyMetrics` for each one. That's addressed below.

---

### ❌ New correctness hole: `applyMetrics` iterates over all GitHub API runners

```swift
for (_, runner) in indexed
    where installPathMap.byId[runner.id] != nil
       || installPathMap.byName[runner.name] != nil {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

The guard (`where installPathMap.byId[runner.id] != nil || installPathMap.byName[runner.name] != nil`) correctly limits this to self-hosted runners. Good. But there's a subtler issue: **`applyMetrics` clears metrics for idle runners**.

```swift
// fetchAndEnrichRunners
for i in indexed.indices where !indexed[i].runner.busy {
    indexed[i].runner = indexed[i].runner.copying(metrics: nil)
}
```

So idle self-hosted runners have `metrics = nil` when they reach the `applyMetrics` write-back. `applyMetrics` then calls `runner.copying(metrics: nil)`, **overwriting** the preserved metrics that `applyRefreshResults` just worked hard to keep. The two mechanisms fight each other:

- `applyRefreshResults` preserves metrics by copying them forward from the pre-refresh snapshot.
- `fetchAndEnrichRunners` then calls `applyMetrics(nil, ...)` for idle runners, immediately stomping those preserved metrics in `LocalRunnerStore.runners`.

The result: CPU/MEM badges will still disappear for self-hosted runners the moment they go idle, even though `applyRefreshResults` was supposed to fix this. The fix is incomplete — either skip `applyMetrics` for idle runners (only write back when `runner.metrics != nil`), or don't call `applyMetrics` for idle runners at all:

```swift
for (_, runner) in indexed
    where runner.busy  // only write back live metrics; don't stomp preserved values for idle runners
       && (installPathMap.byId[runner.id] != nil || installPathMap.byName[runner.name] != nil) {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

---

### ⚠️ `applyMetrics` fires `@Published runners` N times per poll cycle

`LocalRunnerStore` is `@MainActor`-isolated and `runners` is `@Published`. Every call to `applyMetrics` triggers a `$runners` publish event, which the `AppDelegate+PanelSetup` Combine sink catches and calls `self.observable.reload()`. With 10 self-hosted runners at a 10-second poll interval you get 10 extra `observable.reload()` calls per minute that aren't driven by actual data changes. With a large fleet this compounds.

The metrics write-back should batch: accumulate all `(agentId, name, metrics)` pairs in the loop, then apply them in a single pass that only sets `runners` once:

```swift
// One publish instead of N
func applyMetricsBatch(_ updates: [(agentId: Int?, name: String, metrics: RunnerMetrics?)]) {
    var updated = runners
    for u in updates {
        guard let idx = updated.firstIndex(where: { r in
            if let aid = u.agentId, let rid = r.agentId { return aid == rid }
            return r.runnerName == u.name
        }) else { continue }
        updated[idx] = updated[idx].copying(metrics: u.metrics)
    }
    runners = updated  // single publish
}
```

---

### ⚠️ `signOut()` does not call `invalidateTokenCache()` — race still possible

Looking at the diff, `OAuthService.signOut()` calls `Keychain.delete()`. Based on the PR description and #1138 context, `Keychain.delete()` is supposed to also call `invalidateTokenCache()`. But that call is **not visible in this diff**. If `invalidateTokenCache()` lives inside `Keychain.delete()` it's implicit and invisible to the reader. If it doesn't, there's a window between `didSignOut.send()` firing and the AppDelegate sink calling `RunnerStore.shared.start()` → `fetch()` → `githubToken()` where the in-memory cache still holds the old OAuth token.

Worth explicitly calling `invalidateTokenCache()` in `signOut()` before `didSignOut.send()`, or adding a comment that `Keychain.delete()` handles cache invalidation, so the invariant is visible at the call site.

---

### ⚠️ Inner-loop `log()` calls without `#if DEBUG`

`Auth.swift` correctly gates per-resolution-path logs behind `#if DEBUG`. This discipline is missing in several high-frequency call sites:

| Call site | Frequency | Should be `#if DEBUG`? |
|---|---|---|
| `LocalRunnerStore.isTracked()` | Every SwiftUI render | ✅ Yes |
| `refresh()` inner loop — per-runner `isRunning` | Per runner per disk scan | ✅ Yes |
| `fetchAndEnrichRunners` per-runner resolution block | Per runner per poll cycle | ✅ Yes (already done for the verbose block, but the `⚠️ busy but NO installPath` line fires in release too) |
| `popoverShouldClose` | Every popover dismiss | Remove entirely — no diagnostic value |
| `applyRefreshResults` per-runner metrics preservation | Per runner per refresh | ✅ Yes |

The structural logs (fetch start/complete, poll loop interval, sign-out restart, scan counts) are fine unconditional — low frequency, operationally useful. Inner-loop per-item logs must be `#if DEBUG`.

---

### 🔍 Startup sequencing has a subtle ordering assumption

```swift
// AppDelegate+PanelSetup.setupCombineSubscriptions
LocalRunnerStore.shared.refresh()   // kicks async Task
RunnerStore.shared.start()          // kicks async Task (calls fetch() immediately)
```

`refresh()` launches a `Task { }` and returns immediately. `start()` launches its own `Task { }` which calls `fetch()` immediately. Both tasks are now racing on the cooperative thread pool. If the Swift runtime schedules `fetch()` before `refresh()`'s disk-hydration completes (plausible on a fast machine with a slow disk or large runner count), `buildInstallPathMap` still sees `localRunners: []` on the first cycle.

The startup seed call is necessary and its *existence* is correct. But the *guarantee* it provides is only "refresh was kicked before start" — not "refresh completed before first fetch". If first-cycle metrics correctness is required, `fetch()` should `await` on `LocalRunnerStore.shared.runners` being non-empty, or `start()` should be called from the `$runners` sink's first non-empty emission rather than immediately after the `refresh()` kick. This is a deeper architectural concern but worth flagging.

---

### Summary

| Concern | Severity | Action |
|---|---|---|
| `setupSignOutSubscription()` | ✅ Correct | Ship |
| Startup `refresh()` before `start()` | ✅ Correct | Keep |
| `applyMetrics` stomps preserved idle-runner metrics | ❌ Bug | Fix: skip write-back for `!runner.busy` |
| N separate `@Published runners` publishes per poll | ⚠️ Performance | Batch into single assignment |
| `signOut()` cache invalidation not visible at call site | ⚠️ Fragile | Add explicit `invalidateTokenCache()` call or doc comment |
| Inner-loop `log()` without `#if DEBUG` | ⚠️ Prod log spam | Gate high-frequency logs |
| Startup ordering: `refresh()` Task races `fetch()` Task | ⚠️ First-cycle correctness | Architectural follow-up |
| Keychain delete failure leaves poll loop in OAuth mode silently | ⚠️ UX | Follow-up issue |

The sign-out regression is fixed. The one issue that makes the companion changes subtly wrong is the `applyMetrics`-for-idle-runners stomp — it directly undermines the `applyRefreshResults` fix that was meant to solve metric flicker.

eonist:
## Deep Review — PR #1179 · fix: restart poll loop on sign-out so GH_TOKEN/GITHUB_TOKEN fallback is used

---

### ✅ The Core Fix — Correct and Well-Placed

The root cause analysis is spot on. Replacing `Timer + scheduleTimer()` with `pollTask: Task { loop on Task.sleep }` in #1138 broke the implicit "re-resolve token on next tick" behaviour. The old timer would fire regardless, so `githubToken()` naturally fell through to env-var tokens after the Keychain was cleared. The new Task loop never re-enters `start()` on its own, so without an explicit restart, the fallback was silently dead.

Placing `setupSignOutSubscription()` in `AppDelegate` (rather than in `SettingsView`) is the right call — `@State`-stored Combine subscriptions are only alive while that view is on screen, making `SettingsView`'s subscription architecturally unreliable for anything that needs to survive the full app lifetime.

**One concern here though:**

```swift
private func setupSignOutSubscription() {
    OAuthService.shared.didSignOut
        .sink {                          // ⚠️ no [weak self] capture
            log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
            RunnerStore.shared.start()
        }
        .store(in: &cancellables)
}
```

The closure captures no `self` at all, which is fine at runtime since both `AppDelegate` and `RunnerStore.shared` are true singletons — but it's inconsistent with every other sink in `AppDelegate+PanelSetup.swift` which consistently use `[weak self]`. The PR description even shows `[weak self] in guard self != nil else { return }` in the example snippet, yet the final implementation dropped it. This makes the pattern feel like a copy-paste omission. For consistency and defensive correctness (especially if this code is ever extracted or unit-tested), `[weak self]` should be present.

---

### 🟡 `LocalRunnerStore.refresh()` Seeding — Correct Fix, Risky Comment

Calling `LocalRunnerStore.shared.refresh()` *before* `RunnerStore.shared.start()` in `setupCombineSubscriptions()` fixes the real bug where `buildInstallPathMap` always received `localRunners=[]` on the first fetch cycle because `init()` only populates `runnerIndex`, not `runners`.

However, the inline comment in `fetch()` says:

> There is no need to call `refresh()` here — doing so would duplicate the `GET /actions/runners` GitHub API call that `fetchAndEnrichRunners` already makes in the same cycle, doubling the API request rate on the hot path.

This comment is **partially misleading** — `LocalRunnerStore.refresh()` does *not* call `GET /actions/runners`. It calls `RunnerStatusEnricher.shared.enrich()` which may use a different enrichment path. The risk is that a future developer reads this comment, sees "refresh = duplicate GitHub API call", and removes the startup `refresh()` call thinking it's a deduplication, re-introducing the empty `localRunners` bug. The comment should be more precise about *which* API calls overlap and why the startup call is still required.

---

### 🟡 Metrics Write-Back in `fetchAndEnrichRunners` — Subtle Architectural Concern

```swift
// Write metrics back to LocalRunnerStore so the main-view runner row badge
// reflects the latest CPU/MEM values.
for (_, runner) in indexed
    where installPathMap.byId[runner.id] != nil
       || installPathMap.byName[runner.name] != nil {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, forAgentId: runner.id, name: runner.name)
}
```

This introduces a **bidirectional data flow** between `RunnerStore` and `LocalRunnerStore`:
- `RunnerStore.fetch()` reads from `LocalRunnerStore.shared.runners` (for `installPathMap`)
- `RunnerStore.fetch()` now also *writes back* to `LocalRunnerStore.shared` (via `applyMetrics`)

`RunnerStore` already calls `LocalRunnerStore.shared.refresh()` (indirectly via `setupCombineSubscriptions`), and `LocalRunnerStore.applyRefreshResults` now preserves metrics from the existing `runners` array. So there are now **two separate paths** for getting metrics onto `LocalRunnerStore.runners`:
1. `applyRefreshResults` preserves them during a `refresh()` cycle (new in this PR)
2. `applyMetrics` write-back after each `fetch()` (new in this PR)

These two paths can race. `applyRefreshResults` runs in a `Task { }` on `@MainActor`. `applyMetrics` is called from `fetchAndEnrichRunners` which is `async` but not explicitly `@MainActor`. If a `refresh()` cycle completes between the metrics fetch and the `applyMetrics` call, the write-back may be immediately overwritten — but the preservation logic in `applyRefreshResults` should catch it on the *next* `refresh()`. In practice this is probably fine, but the dual-path nature makes the data flow harder to reason about. A comment explaining the intended ownership would be valuable.

Also, `applyMetrics` does an `O(n)` linear scan over `runners` for every runner write-back call. For a small fleet this is irrelevant, but since it's now called in a hot loop inside `fetchAndEnrichRunners`, it's worth noting.

---

### 🟡 `applyRefreshResults` — Metrics Preservation Logic

```swift
var metricsByAgentId: [Int: RunnerMetrics] = [:]
var metricsByName: [String: RunnerMetrics] = [:]
for runner in runners {
    guard let m = runner.metrics else { continue }
    if let aid = runner.agentId { metricsByAgentId[aid] = m }
    metricsByName[runner.runnerName] = m
}
```

This correctly solves the flicker where metrics would disappear between refresh cycles. The agentId-first, name-fallback priority matches the resolve order in `buildInstallPathMap`. Well done.

One edge case: if a runner is *renamed* between cycles (agentId stays the same, name changes), `metricsByAgentId` will correctly carry over the metrics. But if the runner is removed and a *new* runner is installed under the same `runnerName`, `metricsByName` would incorrectly assign the old runner's metrics to the new one. This is an unlikely edge case but worth a comment.

---

### 🔴 Logging Volume in Production Builds

This is the most significant issue in the PR. The majority of the diff (+314/-133) is logging statements, and many are **not gated on `#if DEBUG`**:

- `LocalRunnerStore.refresh()` emits ~10 unconditional log lines per call, including full key lists like `runnerIndex.keys.sorted()` and `runners.map { $0.runnerName }`
- `applyRefreshResults` emits 2–3 lines per runner per cycle, unconditionally
- `RunnerStore.fetch()` emits unconditional logs including the full `localRunners` array formatted as strings
- `scanLiveServices()` emits unconditional logs
- `setupCombineSubscriptions()` logs every time the popover is first set up

The `githubToken()` and `fetchAndEnrichRunners` logs are correctly gated behind `#if DEBUG`. But the `LocalRunnerStore` and `RunnerStore` hot-path logs are not. At a 10s poll interval with 5 runners, `applyRefreshResults` alone generates ~50+ log lines per minute unconditionally in production Release builds. This will spam Console.app for production users and may be a privacy concern (runner names, install paths, and agentIds are logged).

The existing codebase already uses `#if DEBUG` guards for verbose diagnostics (e.g., in `Auth.swift`). The same pattern should be applied consistently to the hot-path logs added here.

---

### ℹ️ Minor / Nits

- **Comment deletions**: Several substantive doc comments were removed (e.g. the `buildInstallPathMap` triple-map explanation, the `launchctlURL` NOSONAR note context, the `deinit` cancel rationale). Some of these were genuinely useful architecture documentation for future contributors. The `# NOSONAR` comment on `launchctlURL` lost its explanatory sentence — the raw comment without context looks odd.
- **`isTracked` log**: `LocalRunnerStore.isTracked` is called potentially very frequently from SwiftUI view code. Even though it's `#if DEBUG` guarded, logging on every `isTracked` call will flood the debug console and make it harder to read the signal logs. Consider removing or making it opt-in.
- **Log separator**: The codebase has moved from `>` to `›` as the separator (e.g. `LocalRunnerStore › ...` vs old `LocalRunnerStore > ...`). The PR is consistent about this in new code — good. But there's one surviving old-style log in `OAuthService.swift`: `log("OAuthService › signOut: Keychain.delete failed — sign-out suppressed...")` uses `:` instead of `—`. Minor but inconsistent.
- **`optimisticallyRemove` double log**: Logs both before and after `removeAll`, which is fine for debugging, but given the log volume concerns above, the before-log is redundant with the after-log which already shows the count transition.

---

### Summary

The core fix (`setupSignOutSubscription` in AppDelegate) and the `LocalRunnerStore.refresh()` pre-seeding are both correct and address real bugs. The metrics preservation in `applyRefreshResults` is a clean improvement. The main concerns are:

1. **🔴 High priority**: Production log volume — hot-path logs need `#if DEBUG` guards consistent with `Auth.swift` and the `fetchAndEnrichRunners` precedent in this very PR.
2. **🟡 Medium**: The `[weak self]` omission in `setupSignOutSubscription` breaks consistency with every other sink in the file.
3. **🟡 Medium**: The bidirectional `RunnerStore ↔ LocalRunnerStore` data flow (read `runners` + write back `applyMetrics`) needs a clearer ownership comment to prevent future confusion.
4. **🟡 Medium**: The `fetch()` comment about "no need to call refresh()" is misleading about *why*.

The logic is sound. Address the log gating and the `[weak self]` and this is good to merge.

eonist:
## Deep Review — PR #1179

Having read the full diff, the PR description, and the existing review threads, here is a consolidated review that picks up issues not fully addressed elsewhere.

---

### ✅ Core fix — `setupSignOutSubscription()` in `AppDelegate`

The root cause diagnosis is accurate. `#1138`'s `Task.sleep` loop is fundamentally different from the old `Timer` approach: a `Timer` fires unconditionally on the next tick regardless of auth state, so token cache invalidation is picked up for free. A `Task` loop only re-enters `githubToken()` if `start()` is called explicitly, so the sign-out path was silently broken for any user who had Settings closed when they signed out — which is nearly all of them.

Placing the subscriber in `AppDelegate.cancellables` is the correct lifetime choice. The doc comment is excellent — naming the regressing PR by number is exactly the kind of paper trail that prevents future maintainers from reverting this "out of caution."

**One gap not addressed anywhere in the existing reviews:** `signOut()` only fires `didSignOut` when `Keychain.delete()` returns `true`. If the Keychain delete fails (sandbox restriction, entitlement issue, or test environment), the poll loop is never restarted, the app silently continues polling with a now-invalid OAuth token, and the user sees no indication anything went wrong. There's already a `log()` for this path, but it warrants a follow-up issue — it's a silent broken state that leaves the app in limbo.

---

### ✅ Startup `LocalRunnerStore.refresh()` before `start()` — keep

Necessary. `init()` only calls `loadIndex()` which populates the `name → path` map in UserDefaults. `runners` stays `[]` until the full disk-hydration → launchctl → GitHub-enrichment pipeline in `refresh()` runs. The startup seed ensures `buildInstallPathMap` has real data on the first cycle.

The `init()` doc comment now makes this contract explicit, which is the right call since it was previously an invisible invariant that bit you.

---

### ❌ `applyMetrics` called for idle runners — this is a bug, not just noise

The existing reviews flag this as "noisy" but it's actually a correctness issue. The write-back loop at the end of `fetchAndEnrichRunners`:

```swift
for (_, runner) in indexed
    where runner.busy
       && (installPathMap.byId[runner.id] != nil
           || installPathMap.byName[runner.name] != nil) {
    LocalRunnerStore.shared.applyMetrics(runner.metrics, ...)
}
```

The `where runner.busy` guard looks correct in isolation, but `applyMetrics` itself does not guard against `metrics == nil`. The problem is the interaction with `applyRefreshResults`:

1. `applyRefreshResults` runs → snapshots and preserves existing CPU/MEM metrics for all runners (busy and idle) from the pre-refresh state.
2. A runner goes idle → `indexed[i].runner = indexed[i].runner.copying(metrics: nil)` stamps `nil` on the GitHub-API runner object.
3. The `where runner.busy` guard *correctly* skips it in the write-back loop above.

So far so good — **but** the next `refresh()` cycle will call `applyRefreshResults` again, and this time the source snapshot (`runners` in `LocalRunnerStore`) contains the `applyMetrics`-written values from the *previous* cycle's busy state. When the runner is idle in this cycle, `applyRefreshResults` *preserves* the stale non-nil metrics from before, making the CPU/MEM badge appear to persist forever even after the job finishes.

In other words: **metrics are preserved across the idle transition and never cleared**. A runner that was at 90% CPU during a job will continue showing 90% CPU until it starts another job or the app restarts. The expected behavior is presumably that idle runners show no metrics.

**Fix:** In `applyRefreshResults`, only preserve metrics for runners that are currently running (`isRunning` or by checking against the enriched runners' busy state), or clear metrics when the enriched model has `metrics == nil` and the runner is not busy.

---

### ⚠️ `applyMetrics` triggers N separate `@Published runners` emissions per poll cycle

`LocalRunnerStore` is `@MainActor`-isolated and `runners` is `@Published`. Every `applyMetrics` call that mutates the array triggers a downstream publish event. The `AppDelegate+PanelSetup` `$runners` sink catches each one and calls `self.observable.reload()`. With 10 self-hosted busy runners you get 10 `observable.reload()` calls in rapid succession on the main thread every poll cycle, causing 10 SwiftUI re-render passes where one would do.

This should be batched:

```swift
// In LocalRunnerStore — single publish
func applyMetricsBatch(_ updates: [(agentId: Int?, name: String, metrics: RunnerMetrics?)]) {
    var updated = runners
    for u in updates {
        guard let idx = updated.firstIndex(where: { r in
            if let aid = u.agentId, let rid = r.agentId { return aid == rid }
            return r.runnerName == u.name
        }) else { continue }
        updated[idx] = updated[idx].copying(metrics: u.metrics)
    }
    runners = updated  // single @Published emission
}
```

---

### ⚠️ Startup sequencing provides a kick, not a guarantee

```swift
LocalRunnerStore.shared.refresh()   // kicks Task { ... }, returns immediately
RunnerStore.shared.start()          // kicks Task { fetch() }, returns immediately
```

Both `refresh()` and `fetch()` are async tasks dispatched to the cooperative thread pool. On a fast machine with a cold UserDefaults and a quick launchctl, `refresh()` may complete before `fetch()` reads `LocalRunnerStore.shared.runners`. On a slow machine or under load, `fetch()` may read `runners` before `refresh()` has finished disk hydration — and `runners` is still `[]`. The startup seed call is necessary, but it's not a hard ordering guarantee.

The safe fix is to await the first non-empty `$runners` emission before calling `start()`, or to have `start()` observe the `$runners` publisher and defer its first `fetch()` until runners are populated. The current approach is best-effort and will silently fall back to empty `installPathMap` on the first cycle on slower devices.

---

### ⚠️ Inner-loop logs are not `#if DEBUG`-gated consistently

`Auth.swift` correctly gates every per-resolution-path log behind `#if DEBUG`. The same discipline is missing in the hot path. The following log lines run unconditionally in Release builds:

| File | Call site | Frequency |
|---|---|---|
| `LocalRunnerStore` | `isTracked()` — emits per call | SwiftUI render frequency |
| `LocalRunnerStore` | `refresh()` inner `isRunning` loop | Per runner per disk scan |
| `LocalRunnerStore` | `applyRefreshResults` per-runner | Per runner per refresh cycle |
| `RunnerStore` | `fetch()` full `localRunners` array | Every poll cycle |
| `RunnerStore` | `fetchAndEnrichRunners` per-runner resolution | Per runner per poll cycle (already `#if DEBUG` for the verbose block, but the `⚠️ busy but NO installPath` warning fires unconditionally) |
| `AppDelegate+PanelSetup` | `popoverShouldClose` | Every popover close |

`popoverShouldClose` in particular should be removed entirely — `log("returning true")` adds zero diagnostic signal. The structural logs (fetch complete, sign-out restart, poll interval) are fine unconditional. Per-item inner-loop logs must be `#if DEBUG`.

---

### ⚠️ `[weak self]` pattern inconsistency in `setupSignOutSubscription`

Every other sink in `AppDelegate+PanelSetup.swift` uses `[weak self]`. The new `setupSignOutSubscription` closure omits it — which is correct at runtime since no `self` is captured — but it breaks the visual consistency of the file. A future developer doing a search-and-replace or extracting this pattern may introduce a retain cycle elsewhere by copying this style. Add `[weak self]` for consistency, even if the `guard let self` check is omitted since `self` is unused.

---

### 🔍 Removed doc comments that should be kept

Several doc comments were removed that provided non-obvious rationale:

- `buildInstallPathMap`: The triple-map explanation (primary fullKey, secondary name, tertiary agentId with the reasoning about agentId being the most reliable) was valuable architecture documentation. It's now replaced with "Builds three lookup maps from the local runner list." — which says nothing about *why* there are three maps or their priority order.
- `launchctlURL NOSONAR` comment: The original had an explanatory sentence (`// NOSONAR — fixed OS path`). It's now just `// NOSONAR` with no context.
- `deinit` cancel rationale: The comment explaining why `pollTask?.cancel()` is called even though `RunnerStore.shared` is never deallocated at runtime was useful for tests and future lifecycle changes. "Documents intent and guards against future lifecycle changes" — now gone.

Documentation that explains *why* non-obvious decisions were made is more valuable than documentation that describes *what* the code does.

---

### Summary

| Concern | Severity | Action |
|---|---|---|
| `setupSignOutSubscription()` | ✅ Correct | Ship |
| Startup `refresh()` before `start()` | ✅ Correct | Keep |
| Metrics preserved forever for idle runners | ❌ Bug | Fix: clear metrics in `applyRefreshResults` when enriched model has `nil` metrics |
| N `@Published runners` emissions per poll | ⚠️ Performance | Batch via `applyMetricsBatch` |
| Startup ordering is a kick, not a guarantee | ⚠️ Correctness | Architectural follow-up; document the limitation |
| Inner-loop `log()` without `#if DEBUG` | ⚠️ Prod log spam | Gate high-frequency logs behind `#if DEBUG` |
| `[weak self]` omission in `setupSignOutSubscription` | ⚠️ Style inconsistency | Add for consistency |
| Removed non-obvious doc comments | ⚠️ Docs | Restore the `buildInstallPathMap` triple-map rationale and `deinit` intent comment |
| Keychain delete failure leaves poll loop in OAuth mode silently | ⚠️ UX | Follow-up issue |

The sign-out regression is fixed and the core logic is sound. The idle-runner metrics-preservation loop is the one change that introduces new incorrect behavior (badges that never clear). The rest are polish concerns that can be addressed before or after merge depending on urgency.

eonist:
## Code Review — `fix/sign-out-restart-poll-loop`

Overall the fixes are correct and the root cause analysis in the PR description is spot-on. Both regressions from #1138 are addressed at the right layer. Below is a deep pass over every file.

---

### ✅ What's solid

**`AppDelegate.swift` — `setupSignOutSubscription()`**
The fix is architecturally correct. Moving the `didSignOut` subscriber from `SettingsView`'s `@State` into `AppDelegate.cancellables` gives it true app-lifetime. The inline doc comment explaining *why* this lives here (and what broke) is genuinely useful — keep it.

**`AppDelegate+PanelSetup.swift` — seeding `LocalRunnerStore` before `start()`**
The ordering fix (`LocalRunnerStore.shared.refresh()` before `RunnerStore.shared.start()`) directly addresses the root cause. The comment block explaining the init/lazy-hydration split is good defensive documentation.

**`applyRefreshResults` — metrics preservation**
The dual-key snapshot strategy (`metricsByAgentId` primary, `metricsByName` fallback) is the right approach. Using `agentId` as the primary key is more reliable than name-matching since names can theoretically collide across scopes.

**`applyMetrics` write-back guard (self-hosted only)**
Guarding the write-back to runners where `byId[runner.id] != nil || byName[runner.name] != nil` is correct. It prevents spurious warnings for cloud-hosted runners that have no local `installPath`.

---

### ⚠️ Issues & concerns

#### 1. `applyMetrics` is a public API that can race with `applyRefreshResults`
`LocalRunnerStore` is `@MainActor`-isolated, so individual calls to `applyMetrics` and `applyRefreshResults` are serialised. However, the sequence:

```
fetchAndEnrichRunners() → applyMetrics()   // called from RunnerStore (not @MainActor)
                       ↕
applyRefreshResults()                      // called via Task on @MainActor
```

`fetchAndEnrichRunners` is called from within the poll `Task` which is not `@MainActor`. The write-back loop at the bottom of `fetchAndEnrichRunners` calls `LocalRunnerStore.shared.applyMetrics(...)` directly. Since `LocalRunnerStore` is `@MainActor`, Swift will hop to the main actor for each call — that's fine for isolation. But if a `LocalRunnerStore.refresh()` cycle completes and `applyRefreshResults` runs *after* `fetchAndEnrichRunners` has finished writing metrics but *before* the next `applyMetrics` batch, you'll get one cycle where metrics are preserved correctly. The concern is the reverse: if `applyRefreshResults` fires (e.g. triggered by `LocalRunnerStore.$runners` sink) *while* `fetchAndEnrichRunners` is mid-flight resolving metrics, the snapshot taken at `applyRefreshResults` entry will be stale (metrics were nil when refresh started) and will overwrite newly-written metrics. The preservation code in `applyRefreshResults` only reads `runners` at the moment it is called — it can't see metrics that `applyMetrics` hasn't written yet. This is an inherent TOCTOU in having two independent async paths writing to the same field on the same model.

**Suggestion:** Consider making the metrics preservation the single source of truth — i.e. remove `applyMetrics` as a separate entry point and instead have `applyRefreshResults` accept metrics as a parameter passed in from `fetchAndEnrichRunners`. This collapses the two write paths into one atomic update.

#### 2. `LocalRunnerStore.shared.refresh()` is called unconditionally before `start()` — no await, fire-and-forget
```swift
LocalRunnerStore.shared.refresh()
RunnerStore.shared.start()
```
`refresh()` launches a `Task { }` internally and returns immediately. `start()` fires right after, so the first `fetch()` can still win the race and call `buildInstallPathMap` before `refresh()`'s Task has had a chance to set `runners`. The comment says this seeds runners "BEFORE" `start()`, but that is only guaranteed for the *dispatch* order, not the *execution* order. On a slow device or under memory pressure, `refresh()`'s background task (which awaits disk I/O + launchctl + GitHub enrichment) will almost certainly not complete before `RunnerStore.fetch()` hits `buildInstallPathMap`.

This is better than the current broken state (runners was never seeded at all), but the race still exists. The log warning you added in `start()` and `fetch()` will likely fire on every cold launch.

**Suggestion:** Make `setupCombineSubscriptions` `async` (or have it call an `async` helper), `await LocalRunnerStore.shared.refresh()`, then call `start()`. Alternatively, have `RunnerStore.start()` internally observe `LocalRunnerStore.$runners.first(where: { !$0.isEmpty })` before launching the poll loop.

#### 3. `signOut` only restarts the poll if `Keychain.delete()` succeeds
```swift
func signOut() {
    let deleted = Keychain.delete()
    if deleted {
        didSignOut.send()
    } else {
        log("OAuthService › signOut: Keychain.delete failed — sign-out suppressed")
    }
}
```
This means if Keychain deletion fails (transient Secure Enclave error, sandbox denial, etc.) `didSignOut` is never emitted, `invalidateTokenCache()` is never called (it isn't in this path), and the token cache still holds the old OAuth token. The poll loop will keep using the stale Keychain token. This is pre-existing behaviour, but the new subscriber in `AppDelegate` makes it more important to document. At minimum add a `TODO` comment that the failure case leaves the cache dirty.

#### 4. `isTracked` logs on every call — production noise
```swift
func isTracked(runnerName: String) -> Bool {
    let tracked = runnerIndex[runnerName] != nil
    #if DEBUG
    log("LocalRunnerStore › isTracked '\(runnerName)' → \(tracked)")
    #endif
    return tracked
}
```
`isTracked` is called from views (row-level rendering) — potentially on every `body` re-evaluation. Even under `#if DEBUG` this will flood the log. It would be better to promote this to a pure computed property with no logging, or at least only log when the result changes.

#### 5. `handleCallback` logs the full OAuth redirect URL
```swift
log("OAuthService › handleCallback — url=\(url.absoluteString)")
```
The redirect URL contains the `code` query parameter. Even though this is a short-lived one-time-use code, logging it verbatim means it will appear in crash logs, Console.app, and any log aggregation pipeline you add. Prefer:
```swift
log("OAuthService › handleCallback — host=\(url.host ?? "?") hasCode=\(comps?.queryItems?.contains { $0.name == \"code\" } == true)")
```

#### 6. Docstring removed from `add(runnerName:installPath:)` without replacement
The original had:
```
/// Convenience alias for `register(name:installPath:)` with view-friendly parameter labels
/// so SwiftUI call sites read `store.add(runnerName: x, installPath: y)` naturally.
```
This was removed and replaced with just a log line. The public API surface still needs to communicate *why* this method exists separately from `register`. Without the doc comment a future reader may think it's redundant and delete it.

#### 7. `fetch()` comment says "no need to call `refresh()` here" — slightly misleading
The comment in `fetch()` says not calling `refresh()` avoids doubling the `GET /actions/runners` API call. That's true for the enrichment step, but `LocalRunnerStore.refresh()` also runs `launchctl list` and re-reads disk `.runner` JSON — neither of which is a GitHub API call. If a runner is installed or removed mid-session, the local state won't be updated until the next external trigger. This is acceptable behaviour but the comment should be more precise about what is actually being avoided (GitHub API calls) vs what isn't refreshed (disk/launchctl state).

---

### 🔍 Minor / nits

- **Log volume in production (non-DEBUG) builds is very high.** The non-`#if DEBUG`-gated `log()` calls added in this PR will fire on every poll cycle in release builds. `applyRefreshResults`, `buildInstallPathMap`, `fetch`, `scanLiveServices` and the `refresh()` Task all log at least 3–5 lines per cycle. At a 10s poll interval that's ~18–30 log lines/min in production. Consider whether these belong behind `#if DEBUG` or a runtime verbose-logging flag.

- **`optimisticallySetRunning` and `setLifecycleWarning` both log the "not found" warning unconditionally.** These are benign on normal operation (the runner hasn't appeared yet) but will generate spurious warnings during startup before `refresh()` completes. Adding an `isScanning` guard before the warning would reduce noise.

- **`runnerModelFromIndex` is a free function in a `// MARK: - .runner JSON parser` section but its doc comment was removed.** The BOM-stripping rationale (GitHub Actions runner agent writes BOM-prefixed JSON) is non-obvious and worth keeping; the removed comment explained it well.

- **`deinit` comment removal in `RunnerStore`.** The removed comment explaining *why* `pollTask?.cancel()` is explicit even though the singleton is never deallocated was genuinely useful for test/future-use readers. Consider keeping it or moving it to the deinit line as an inline comment.

---

### Summary

| Area | Status |
|---|---|
| Sign-out regression fix | ✅ Correct, right layer |
| CPU/MEM first-launch fix | ✅ Correct, ordering is better |
| CPU/MEM flicker fix (`applyRefreshResults`) | ✅ Correct |
| Race between `applyMetrics` and `applyRefreshResults` | ⚠️ Possible, see §1 |
| `refresh()` → `start()` ordering guarantee | ⚠️ Fire-and-forget race, see §2 |
| OAuth code in logs | ⚠️ Security concern, see §5 |
| Log volume in release builds | ⚠️ High, see nits |
| Removed doc comments | 🔍 Minor, see §6, §7, nits |

eonist:
Reviewed both comments against the current branch diff. Most points are already addressed or are architectural follow-ups — one new real bug stands.

---

### ✅ Already handled — retracting those points

| Claim | Status |
|---|---|
| `applyMetrics` stomps idle-runner preserved metrics | ✅ Fixed — `runner.busy &&` guard is already in the write-back loop at the bottom of `fetchAndEnrichRunners` |
| `applyMetrics` called for cloud runners | ✅ Already guarded by `installPathMap` check |
| Inner-loop logs not `#if DEBUG` | ✅ Already addressed in the prior round |
| `[weak self]` omission | ✅ Style only, no runtime impact |
| `Keychain.delete()` leaves cache dirty | ✅ `Keychain.delete()` already calls `invalidateTokenCache()` internally |

---

### 🐛 One genuine new bug — `applyRefreshResults` preserves metrics for idle runners forever

The current preservation logic in `applyRefreshResults`:

```swift
let preserved: [RunnerModel] = enriched.map { runner in
    if let aid = runner.agentId, let m = metricsByAgentId[aid] {
        return runner.copying(metrics: m)   // ← re-applies OLD metrics unconditionally
    }
    if let m = metricsByName[runner.runnerName] {
        return runner.copying(metrics: m)
    }
    return runner
}
```

This never clears. A runner at 90% CPU during a job will keep showing 90% CPU after it goes idle, because the snapshot is re-applied regardless of `runner.busy`. The `fetchAndEnrichRunners` path correctly stamps `metrics = nil` for idle runners before the write-back guard, but `applyRefreshResults` then overwrites that nil with the stale snapshot on the very next refresh cycle.

**One-line fix:**

```swift
let preserved: [RunnerModel] = enriched.map { runner in
    guard runner.busy else { return runner }  // idle → respect nil metrics from enriched model
    if let aid = runner.agentId, let m = metricsByAgentId[aid] {
        return runner.copying(metrics: m)
    }
    if let m = metricsByName[runner.runnerName] {
        return runner.copying(metrics: m)
    }
    return runner
}
```

Everything else in both review comments is noise you've already handled. This is the only blocker.

eonist:
## Follow-up — narrowing to the one real new finding

Fair pushback on the previous review. After cross-checking against the current branch state:

**Already addressed — not actionable:**
- `applyMetrics` stomping idle-runner metrics → `runner.busy &&` guard is already in the write-back loop at `2732349c`
- `applyMetrics` firing for cloud runners → `installPathMap` guard is already there
- Inner-loop logs / `#if DEBUG` → already handled in the prior round
- `[weak self]` omission → style only, no runtime impact
- `Keychain.delete()` cache invalidation → already calls `invalidateTokenCache()` internally

**The one genuinely new finding — `applyRefreshResults` preserves metrics for idle runners indefinitely**

```swift
let preserved: [RunnerModel] = enriched.map { runner in
    if let aid = runner.agentId, let m = metricsByAgentId[aid] {
        return runner.copying(metrics: m)   // ← re-applies old metrics regardless of busy state
    }
    if let m = metricsByName[runner.runnerName] {
        return runner.copying(metrics: m)
    }
    return runner
}
```

The snapshot logic never checks whether the enriched runner is currently busy. A runner that had CPU/MEM metrics last cycle will keep showing stale badges after it goes idle, because the preservation always wins over the nil-metrics enriched model. The `applyMetrics` write-back is correctly guarded to busy-only, but `applyRefreshResults` is not.

One-line fix:

```swift
let preserved: [RunnerModel] = enriched.map { runner in
    guard runner.busy else { return runner }  // idle → let enriched nil-metrics through
    if let aid = runner.agentId, let m = metricsByAgentId[aid] {
        return runner.copying(metrics: m)
    }
    if let m = metricsByName[runner.runnerName] {
        return runner.copying(metrics: m)
    }
    return runner
}
```

Everything else in the previous review is either already handled or an architectural follow-up, not a merge blocker. This is the only new actionable item.

eonist:
## Deep Review — `fix/sign-out-restart-poll-loop`

The PR description is honest and accurate about what was broken and why. The root-cause analysis for both bugs is correct, and the fixes are targeted. That said, there are some things worth discussing before this merges.

---

### ✅ What's correct and well-done

**Sign-out fix — correct approach, right place.**
Moving the `didSignOut` subscriber from `SettingsView`'s `@State signOutCancellable` into `AppDelegate.cancellables` is exactly right. A `@State` property is only alive while the view is in the hierarchy, making it unsuitable for app-lifetime subscriptions. Storing it in `AppDelegate.cancellables` gives it the correct lifetime. The regression from #1138 (timer → `Task.sleep` loop) is accurately identified.

**CPU/MEM fix — the ordering fix is sound.**
Calling `LocalRunnerStore.shared.refresh()` before `RunnerStore.shared.start()` in `setupCombineSubscriptions()` is the right sequence fix. Previously `buildInstallPathMap` always received `localRunners: []` on first fetch because `init()` only calls `loadIndex()`, not `refresh()`. The comment block in `setupCombineSubscriptions()` explains this well.

**Metrics preservation in `applyRefreshResults` — good defensive logic.**
Snapshotting `metricsByAgentId` and `metricsByName` before replacing `runners` stops the flicker. Matching by `agentId` first and falling back to name mirrors the existing resolution hierarchy in `applyMetrics`, which is consistent.

**Write-back guard (`runner.busy && installPath resolved`) — correct and necessary.**
Without the `busy` guard, `applyMetrics` would stomp freshly-preserved metrics with `nil` for any idle runner that previously had data. The guard also prevents the spurious ⚠️ log spam for cloud-hosted runners that have no local entry at all.

---

### 🟡 Issues to address

#### 1. `setupSignOutSubscription` — no `[weak self]` in the sink closure
```swift
// Current
.sink {
    log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
    RunnerStore.shared.start()
}
```
`AppDelegate` is a singleton so there is no practical retain cycle here, but the sink does capture `AppDelegate`'s `cancellables` through the store. Consistency with every other sink in this file (all use `[weak self]`) matters more than the runtime risk. If `AppDelegate` ever becomes non-singleton in tests, this is a latent leak.

---

#### 2. `refresh()` is called synchronously on the main actor before `start()` — but it's `async` inside a `Task`

```swift
// AppDelegate+PanelSetup.swift
LocalRunnerStore.shared.refresh()    // ← dispatches a Task internally
RunnerStore.shared.start()           // ← runs immediately after
```

`refresh()` itself returns immediately — it just launches a `Task { ... }` internally and sets `isScanning = true`. `RunnerStore.shared.start()` fires before that Task has done any disk I/O, launchctl scan, or GitHub enrichment. So on first launch, `LocalRunnerStore.shared.runners` is still `[]` when `start()` kicks off its `fetch()`.

The **first** `fetch()` call inside the poll loop will still see `localRunners = []`. This doesn't regress the fix (it's better than before, because `refresh()` is now *in flight*), but the comment says:

> *"seeding `runners` before the first fetch runs"*

That's not strictly true — it seeds a *concurrent in-flight task* before the first fetch, which is a race. In practice this is usually fine because `refresh()` involves async disk + launchctl + network, and the poll loop's first `fetch()` involves a real GitHub API call — refresh almost certainly wins. But relying on this timing is fragile. Worth adding a note to the comment or, ideally, `await`-ing `refresh()` before `start()` if `refresh()` is ever made `async`.

---

#### 3. `applyRefreshResults` — metrics are preserved unconditionally for non-busy runners too

```swift
for runner in runners {
    guard runner.isBusy, let m = runner.metrics else { continue }  // ← only isBusy runners saved
    ...
}
```

The **snapshot** is guarded to `isBusy` — only busy runners' metrics are saved. But the **apply** side maps over all `enriched` runners (which may include previously-idle runners that became busy between refresh cycles). If a runner transitioned from idle → busy during the enrichment roundtrip, it will find no entry in the snapshot and get `nil` metrics — which is correct.

However, if a runner transitioned from busy → idle between the snapshot and `applyRefreshResults` firing, it will still get its old metrics applied (since its name/agentId is in the snapshot). On the *next* `applyRefreshResults` it will be filtered out since it won't be busy anymore. So at worst there's a one-cycle stale metric on a newly-idle runner. This is likely acceptable, but worth a comment explaining the conscious choice.

---

#### 4. Logging volume in production builds

The `#if DEBUG` gates on the hottest inner-loop paths (`applyMetrics`, `fetchAndEnrichRunners` per-runner resolution, `isTracked`) are correct. But there are a lot of unconditional `log(...)` calls that fire on **every poll cycle** in release builds:

- `RunnerStore › fetch — localRunners.count=N` (every fetch)
- `LocalRunnerStore › refresh() called — isScanning=...` (every refresh)
- `LocalRunnerStore › applyRefreshResults — enriched.count=N` (every refresh)
- `RunnerStore › fetchAndEnrichRunners ENTER — scopes=...` (every fetch)
- `LocalRunnerStore › scanLiveServices — running launchctl list` (every refresh)

For a menu-bar app polling every 10–30 seconds, these will produce ~hundreds of log lines per hour. Consider `#if DEBUG`-gating the per-cycle operational logs too, and reserving unconditional logs for warnings (⚠️), errors, and one-time lifecycle events (startup, sign-in, sign-out).

---

#### 5. `setupPanel` verbose lifecycle logs in release builds

```swift
log("AppDelegate › setupPanel — begin")
log("AppDelegate › setupPanel — popover created, wiring KVO + Combine")
log("AppDelegate › setupPanel — complete")
```

These are one-time startup events so the volume is fine — but they add noise to release Console output with no user-facing value. Suggestion: gate behind `#if DEBUG` or consolidate into one log line.

---

#### 6. Doc comment duplication in `LocalRunnerStore.refresh()`

The new doc comment says:

> *"3. From any view that needs a fresh scan (SettingsView, lifecycle actions)"*

This duplicates point 2: *"On-demand from views..."*. Minor, but doc comments are part of the public interface — worth keeping clean.

Also, the original doc comment for `applyRefreshResults` mentioned it was extracted to keep closure nesting within a 2-level limit. That context is now gone. If the 2-level rule is a project standard, a single-line reference would be useful for reviewers.

---

### 🔴 One potential correctness concern

#### Race between `refresh()` Task and `RunnerStore.start()` first fetch — `isScanning` flag

`LocalRunnerStore.isScanning` is set to `true` at the start of `refresh()` and cleared in `applyRefreshResults` (which runs on `@MainActor`). `RunnerStore.start()` launches a `pollTask` that immediately calls `fetch()`, which reads `LocalRunnerStore.shared.runners` (not `isScanning`). So `fetch()` can proceed while `refresh()` is still scanning, which is fine for correctness.

But what happens if the `didUpdate` publisher fires (from `applyFetchResult`) before `applyRefreshResults` runs? The Combine sink in `setupCombineSubscriptions` calls `self.observable.reload()` on `RunnerStore.didUpdate`, which may trigger a UI read of `LocalRunnerStore.shared.runners` while it is still `[]`. This was already the case before this PR, but the new `refresh()` call makes it more likely to occur simultaneously with an in-progress scan. Worth verifying this scenario doesn't cause a visible flash of empty state.

---

### Summary

| Area | Verdict |
|---|---|
| Sign-out regression fix | ✅ Correct |
| CPU/MEM first-launch fix | ✅ Correct (with race caveat) |
| Metrics preservation across cycles | ✅ Correct |
| Write-back guard (busy + self-hosted only) | ✅ Correct |
| Missing `[weak self]` in sign-out sink | 🟡 Fix for consistency |
| `refresh()` → `start()` ordering comment accuracy | 🟡 Comment is slightly overstated |
| Log volume in release builds | 🟡 Consider `#if DEBUG` gates on hot-path logs |
| Idle→busy metrics transition comment | 🟡 Add a comment explaining the one-cycle stale window |
| Potential empty-state flash on first launch | 🔴 Verify/test the scenario |

Overall this is a solid, well-analysed fix. The main asks before merging are: verify the empty-state flash scenario, add `[weak self]` to the sign-out sink, and audit the unconditional per-cycle log lines.

Issue #1206, opened 2026-06-08
🔧 #1202 — Extract local runner management into LocalRunnersView


eonist:
## Problem

Issue #1202 asks to move the runner section in Settings into a new dedicated view, accessible via a **"Manage local runners"** row.

Currently `SettingsView` is a monolithic ~700-line view containing six sections. The local runner section alone accounts for ~250 lines and owns:
- All runner-specific `@State` (`hasLoadedOnce`, `runnerPendingRemoval`, `showAddRunnerSheet`, `editingRunner`, `isCommitting`, `commitError`, `removeErrorMessage`)
- `@StateObject var localRunnerStore`
- `localRunnersSection`, `localRunnersSectionHeader`, `localRunnerRow`, `localRunnerRowContent`
- `performResume`, `performStop`, `performRemoval`, `localRunnerDotColor`
- `runnerEditingPopover`, `addRunnerSheet`, `removalAlertModifier`

This makes `SettingsView` hard to read and violates the single-responsibility principle.

## Solution

Extract all local runner UI and logic into a new `LocalRunnersView` struct, and replace the runner section in `SettingsView` with a navigation row that drills into it.

**Navigation pattern:** The app uses a flat callback-based nav model (no `NavigationStack`) driven by `NavState` + `AppDelegate+Navigation.swift`. `LocalRunnersView` is presented by swapping the panel content via `@State private var showLocalRunners = false` inside `SettingsView`, using the same back-button/`onBack` callback pattern already established by `SettingsView` itself.

## Files changed

| File | Change |
|---|---|
| `Sources/RunnerBar/Views/Settings/LocalRunnersView.swift` | **New** — full runner management UI extracted from `SettingsView` |
| `Sources/RunnerBar/Views/Settings/SettingsView.swift` | **Modified** — replace runner section with a "Manage local runners" nav row; remove runner-specific state/methods |

## SwiftLint / Periphery / Tests checklist

- ✅ `file_header` rule: both files carry the `// FileName.swift\n// RunnerBar` header
- ✅ `missing_docs`: every `internal` declaration carries a `///` doc comment
- ✅ `sorted_imports`: imports are alphabetical in both files
- ✅ No dead code added — `LocalRunnersView` is referenced from `SettingsView`, so Periphery will not flag it
- ✅ `localRunnerStore` removed from `SettingsView` — no orphaned `@StateObject`
[...]

PR #1207, opened 2026-06-08
feat: extract LocalRunnersView + ScopesView from SettingsView (#1202 #1203)

### LINKED ISSUE #1202: ⭐ Move the runner section in settings to a new view
Add a new row in settings named "Manage local runners"

### LINKED ISSUE #1203: ⭐ Move the scopes section in settings to a new view
Add a new row in settings named "Manage scopes"

eonist:
## Summary

Refactors the monolithic `SettingsView` across three commits:
1. Extract local runner management → `LocalRunnersView` (#1202)
2. Extract scope management → `ScopesView` (#1203)
3. Merge the standalone Notifications section into General

`SettingsView` is now a slim shell of two nav rows + General / Account / About sections.

Closes #1202
Closes #1203
Ref #1206 (issue with full problem/solution/code proposal for #1202)

---

## Commits

| SHA | What |
|---|---|
| `8874b83` | feat: extract `LocalRunnersView` from `SettingsView` (#1202) |
| `77b5ba6` | feat: extract `ScopesView` from `SettingsView` (#1203) |
| `28d1ad2` | refactor: merge `notificationsSection` into `generalSection` |

---

### `LocalRunnersView.swift` (new)
Owns everything runner-related that previously lived inline in `SettingsView`:
- `@StateObject private var localRunnerStore`
- All runner-specific `@State` (`hasLoadedOnce`, `runnerPendingRemoval`, `showAddRunnerSheet`, `editingRunner`, `isCommitting`, `commitError`, `removeErrorMessage`)
- `headerBar`, `contentStack`, `sectionHeader`, `descriptionLabel`, `errorLabel`, `runnerList`
- `localRunnerRow`, `localRunnerRowContent`
- `performResume`, `performStop`, `performRemoval`
- `localRunnerDotColor`, `addRunnerSheet`, `removalAlertModifier`, `runnerEditingPopover`

Presented by `SettingsView` via `showLocalRunners` flag — same back-callback pattern as the rest of the panel nav model.

### `SettingsView.swift` after commit 1
- **Removed**: `@StateObject localRunnerStore`, 7 runner `@State` vars, `localRunnersSection` (~250 lines), lifecycle methods, runner sheets/popover/modifier
- **Added**: `@State private var showLocalRunners = false`
- `body` swaps to `LocalRunnersView` when `showLocalRunners == true`
- `sectionsStack`: `localRunnersSection` → `manageLocalRunnersRow`

---

### `ScopesView.swift` (new)
Owns everything scope-related that previously lived in `SettingsView`:
- `@StateObject private var scopeStore`
- `showAddScopeSheet` + `selectedScopeEntry` `@State`
- `headerBar`, `contentStack`, `sectionHeader`, `descriptionLabel`, `scopeList`, `scopeRow`
- `AddScopeSheet` and `ScopeEditSheet` presentation

### `SettingsView.swift` after commit 2
- **Removed**: `@StateObject scopeStore`, `showAddScopeSheet`, `selectedScopeEntry`, `scopeMutateCancellable`, `remoteScopesSectionHeader`, `remoteScopesSection`, `scopeRow`, both scope `.sheet` modifiers, `scopeMutateCancellable` assignment
- **Added**: `@State private var showScopes = false`
- `body` `if/else` chain also handles `showScopes → ScopesView`
- `sectionsStack`: `remoteScopesSection` → `manageScopesRow`
- `onAppearAction` reduced to auth state + sign-out listener only

---

### `SettingsView.swift` after commit 3
- **Removed**: standalone `notificationsSection` computed var and its `Divider()` separator in `sectionsStack`
- **Merged**: "Notify on success" and "Notify on failure" toggles added at the top of `generalSection`, separated from the existing controls by an inset `Divider`
- `@StateObject private var notifications` stays in `SettingsView` — its bindings are still needed by the merged toggles inside `generalSection`
- `sectionsStack` order is now: `manageLocalRunnersRow` → `manageScopesRow` → `generalSection` → `accountSection` → `aboutSection`
- Saves one top-level section header and one full-width `Divider` from the scroll view

---

## Final `SettingsView` section inventory

| Row / Section | Where it lives now |
|---|---|
| Manage local runners (nav row) | `SettingsView` → drills to `LocalRunnersView` |
| Manage scopes (nav row) | `SettingsView` → drills to `ScopesView` |
| General (incl. notifications) | `SettingsView` inline |
| Account | `SettingsView` inline |
| About | `SettingsView` inline |

---

## Checklist

- [x] `file_header` SwiftLint rule — all files open with `// FileName.swift\n// RunnerBar`
- [x] `missing_docs` — every `internal` declaration has a `///` doc comment
- [x] `sorted_imports` — alphabetical in all files
- [x] No dead symbols — `LocalRunnersView` and `ScopesView` directly referenced from `SettingsView`; Periphery won't flag them
- [x] No model or store changes — `RunnerBarCoreTests` and `RunnerBarUITests` unaffected
- [x] HEIGHT/WIDTH CONTRACT comments preserved verbatim in `settingsBody`
- [x] `panelSheetState.clearRunnerSheet()` called via `AppDelegate+Navigation.swift` `onBack` — unchanged
- [x] Does **not** touch `NavState`, `AppDelegate+Navigation.swift`, or any other navigation infrastructure

## Release Notes

* **New Features**
  * Dedicated local runners management screen with start/stop controls, removal confirmation, and editing capabilities
  * Remote runner scopes management interface with enable/disable toggles and removal options
  * New "Management" section in settings consolidating runner and scope controls

* **Improvements**
  * Reorganized settings layout (Account → Management → General → About) for improved clarity

### DIFF
--- a/Sources/RunnerBar/Views/Settings/LocalRunnersView.swift
+++ b/Sources/RunnerBar/Views/Settings/LocalRunnersView.swift
@@ -1,6 +1,5 @@
 // LocalRunnersView.swift
 // RunnerBar
-import Combine
 import RunnerBarCore
 import SwiftUI
 
--- a/Sources/RunnerBar/Views/Settings/LocalRunnersView.swift
+++ b/Sources/RunnerBar/Views/Settings/LocalRunnersView.swift
@@ -230,12 +230,12 @@ struct LocalRunnersView: View {
             case .success: break
             case .corruptInstall:
                 LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: false)
-                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 corrupt install")
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u{26A0} corrupt install")
             case .failed(let msg):
                 LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: false)
                 let short = msg.components(separatedBy: "\n")
                     .first(where: { !$0.trimmingCharacters(in: .whitespaces).isEmpty }) ?? msg
-                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 \(short)")
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u{26A0} \(short)")
             }
             LocalRunnerStore.shared.refresh()
         }
@@ -252,12 +252,12 @@ struct LocalRunnersView: View {
             switch result {
             case .success: break
             case .corruptInstall:
-                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 corrupt install")
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u{26A0} corrupt install")
             case .failed(let msg):
                 LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: true)
                 let short = msg.components(separatedBy: "\n")
                     .first(where: { !$0.trimmingCharacters(in: .whitespaces).isEmpty }) ?? msg
-                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 \(short)")
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u{26A0} \(short)")
             }
             LocalRunnerStore.shared.refresh()
         }
--- a/Sources/RunnerBar/Views/Settings/LocalRunnersView.swift
+++ b/Sources/RunnerBar/Views/Settings/LocalRunnersView.swift
@@ -1,0 +1,350 @@
+// LocalRunnersView.swift
+// RunnerBar
+import Combine
+import RunnerBarCore
+import SwiftUI
+
+// MARK: - LocalRunnersView
+
+/// Full runner-management screen, reached from the "Manage local runners" row in Settings.
+///
+/// Owns all runner-specific state and lifecycle actions that previously lived in `SettingsView`.
+/// Presented by `SettingsView` via a `showLocalRunners` flag using the same back-callback
+/// pattern established by the rest of the panel navigation model.
+@MainActor
+struct LocalRunnersView: View {
+
+    // MARK: - Inputs
+
+    /// Callback invoked when the user taps the back button.
+    let onBack: () -> Void
+    /// Combined OAuth + CLI auth state forwarded from `SettingsView`; required by `RemovalAlertModifier`.
+    let isAuthenticated: Bool
+
+    // MARK: - Observed stores
+
+    /// Index of locally-installed self-hosted runners.
+    @StateObject private var localRunnerStore = LocalRunnerStore.shared
+
+    // MARK: - Local UI state
+
+    /// `true` once the initial local runner scan has completed.
+    @State private var hasLoadedOnce = false
+    /// The runner awaiting user confirmation before removal.
+    @State private var runnerPendingRemoval: RunnerModel?
+    /// Controls presentation of `AddRunnerSheet`.
+    @State private var showAddRunnerSheet = false
+    /// The runner currently being edited in `RunnerDetailPopover`. `nil` = popover dismissed.
+    @State private var editingRunner: RunnerModel?
+    /// `true` while `commitRunnerEdit` is in-flight.
+    @State private var isCommitting = false
+    /// Non-nil when the last commit attempt produced errors; forwarded into `RunnerDetailPopover`.
+    @State private var commitError: String?
+    /// Non-nil when the last removal attempt failed; shown as an inline error label.
+    @State private var removeErrorMessage: String?
+
+    // MARK: - Computed properties
+
+    /// Alert title incorporating the pending runner name.
+    private var removalAlertTitle: String {
+        let name = runnerPendingRemoval?.runnerName ?? "this runner"
+        return "Remove runner \"\(name)\"?"
+    }
+
+    // MARK: - Body
+
+    /// Root layout: fixed header bar above a scrollable runner list.
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            headerBar
+            Divider()
+            ScrollView(.vertical, showsIndicators: true) {
+                contentStack
+                    .padding(.bottom, 16)
+            }
+            .frame(maxHeight: .infinity)
+        }
+        .frame(idealWidth: 480, maxWidth: .infinity)
+        .onAppear { localRunnerStore.refresh() }
+        .onChange(of: localRunnerStore.isScanning) { _, newVal in if !newVal { hasLoadedOnce = true } }
+        .sheet(isPresented: $showAddRunnerSheet, content: addRunnerSheet)
+        .modifier(removalAlertModifier)
+        .popover(item: $editingRunner) { runner in runnerEditingPopover(runner: runner) }
+    }
+
+    // MARK: - Header
+
+    /// Top bar with back button and "Manage local runners" title.
+    private var headerBar: some View {
+        HStack {
+            Button(action: onBack, label: {
+                HStack(spacing: 4) {
+                    Image(systemName: "chevron.left")
+                        .font(.system(size: 12, weight: .medium))
+                    Text("Manage local runners").font(.headline)
+                }
+                .foregroundColor(.primary)
+            })
+            .buttonStyle(.plain)
+            Spacer()
+        }
+        .padding(.horizontal, RBSpacing.md).padding(.top, 12).padding(.bottom, 8)
+    }
+
+    // MARK: - Content
+
+    /// Vertical stack of the runner list and related controls.
+    private var contentStack: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            sectionHeader
+            descriptionLabel
+            errorLabel
+            runnerList
+        }
+    }
+
+    /// Section header row with add and refresh buttons.
+    private var sectionHeader: some View {
+        HStack {
+            Text("Active local runners")
+                .font(RBFont.sectionHeader).foregroundColor(Color.rbTextSecondary)
+            Spacer()
+            Button(action: { showAddRunnerSheet = true }, label: {
+                Image(systemName: "plus").font(.caption).foregroundColor(Color.rbTextSecondary)
+            })
+            .buttonStyle(.plain)
+            .help("Add a new runner")
+            .accessibilityIdentifier("addRunnerButton")
+            .padding(.trailing, 4)
+            if localRunnerStore.isScanning {
+                ProgressView().scaleEffect(0.6).frame(width: 14, height: 14)
+            } else {
+                Button(action: { removeErrorMessage = nil; localRunnerStore.refresh() }, label: {
+                    Image(systemName: "arrow.clockwise").font(.caption).foregroundColor(Color.rbTextSecondary)
+                })
+                .buttonStyle(.plain).help("Refresh local runner list")
+            }
+        }
+        .padding(.horizontal, RBSpacing.md).padding(.top, 8).padding(.bottom, 4)
+    }
+
+    /// Subtitle describing what local runners are.
+    private var descriptionLabel: some View {
+        Text("Self-hosted runners installed on this machine, discovered via LaunchAgent plists.")
+            .font(.caption).foregroundColor(Color.rbTextSecondary)
+            .padding(.horizontal, RBSpacing.md).padding(.bottom, 6)
+    }
+
+    /// Inline error label shown when the last removal failed.
+    @ViewBuilder
+    private var errorLabel: some View {
+        if let errMsg = removeErrorMessage {
+            Text(errMsg).font(.caption).foregroundColor(Color.rbDanger)
+                .padding(.horizontal, RBSpacing.md).padding(.vertical, 4)
+                .background(Color.rbDanger.opacity(0.07))
+        }
+    }
+
+    /// Empty-state placeholder or populated list of runner rows.
+    @ViewBuilder
+    private var runnerList: some View {
+        if localRunnerStore.runners.isEmpty && !localRunnerStore.isScanning && hasLoadedOnce {
+            Text("No local runners found").font(.caption).foregroundColor(Color.rbTextSecondary)
+                .padding(.horizontal, RBSpacing.md).padding(.vertical, 4)
+        } else {
+            ForEach(localRunnerStore.runners) { runner in localRunnerRow(runner) }
+        }
+    }
+
+    // MARK: - Runner rows
+
+    /// Full row view for a single local runner, including the detail popover trigger.
+    private func localRunnerRow(_ runner: RunnerModel) -> some View {
+        Button {
+            commitError = nil
+            editingRunner = runner
+        } label: {
+            localRunnerRowContent(runner)
+                .contentShape(Rectangle())
+        }
+        .buttonStyle(.plain)
+        .padding(.horizontal, RBSpacing.md).padding(.vertical, 5)
+        .glassCard(cornerRadius: RBRadius.small)
+        .padding(.horizontal, RBSpacing.xs)
+    }
+
+    /// Inner content (status dot, name, start/stop toggle, remove button) for a local runner row.
+    private func localRunnerRowContent(_ runner: RunnerModel) -> some View {
+        let hasWarning = runner.lifecycleWarning != nil
+        let displayStatus = runner.displayStatus
+        return HStack(spacing: 6) {
+            Circle().fill(localRunnerDotColor(for: runner)).frame(width: 8, height: 8)
+            VStack(alignment: .leading, spacing: 1) {
+                Text(runner.runnerName).font(.system(size: 13)).lineLimit(1)
+                if let url = runner.gitHubUrl {
+                    Text(url).font(.caption2).foregroundColor(Color.rbTextSecondary).lineLimit(1)
+                }
+            }
+            Spacer()
+            Text(displayStatus)
+                .font(.caption)
+                .foregroundColor(hasWarning ? Color.rbWarning : Color.rbTextSecondary)
+                .lineLimit(1)
+                .fixedSize()
+            Toggle("", isOn: Binding(
+                get: { runner.isRunning },
+                set: { isOn in
+                    if isOn { performResume(runner: runner) } else { performStop(runner: runner) }
+                }
+            ))
+            .toggleStyle(.switch)
+            .tint(Color.rbSuccess)
+            .labelsHidden()
+            .help(runner.isRunning ? "Stop runner service" : "Start runner service")
+            .scaleEffect(0.8, anchor: .trailing)
+            .buttonStyle(.borderless)
+            Image(systemName: "chevron.right")
+                .font(.caption2)
+                .foregroundColor(Color.rbTextTertiary)
+            Button(action: { runnerPendingRemoval = runner },
+                   label: { Image(systemName: "minus.circle").font(.caption2).foregroundColor(Color.rbDanger) })
+            .buttonStyle(.plain).help("Remove runner")
+        }
+    }
+
+    // MARK: - Lifecycle actions
+    // TODO: #1077 — migrate to async/await once RunnerLifecycleService.start/stop are async.
+    // Current pattern (Task + Task.detached) matches LocalRunnerStore.refresh() as the
+    // intermediate step: background work is off-actor, main-actor mutations happen in the
+    // Task continuation which returns to @MainActor automatically. // NOSONAR
+
+    /// Optimistically marks the runner as running then delegates to `RunnerLifecycleService`.
+    @MainActor private func performResume(runner: RunnerModel) {
+        log("LocalRunnersView > performResume called runner=\(runner.runnerName)")
+        LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: true)
+        Task {
+            let result = await Task.detached(priority: .userInitiated) {
+                RunnerLifecycleService.shared.start(runner: runner)
+            }.value
+            switch result {
+            case .success: break
+            case .corruptInstall:
+                LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: false)
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 corrupt install")
+            case .failed(let msg):
+                LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: false)
+                let short = msg.components(separatedBy: "\n")
+                    .first(where: { !$0.trimmingCharacters(in: .whitespaces).isEmpty }) ?? msg
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 \(short)")
+            }
+            LocalRunnerStore.shared.refresh()
+        }
+    }
+
+    /// Optimistically marks the runner as stopped then delegates to `RunnerLifecycleService`.
+    @MainActor private func performStop(runner: RunnerModel) {
+        log("LocalRunnersView > performStop called runner=\(runner.runnerName)")
+        LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: false)
+        Task {
+            let result = await Task.detached(priority: .userInitiated) {
+                RunnerLifecycleService.shared.stop(runner: runner)
+            }.value
+            switch result {
+            case .success: break
+            case .corruptInstall:
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 corrupt install")
+            case .failed(let msg):
+                LocalRunnerStore.shared.optimisticallySetRunning(runner.runnerName, isRunning: true)
+                let short = msg.components(separatedBy: "\n")
+                    .first(where: { !$0.trimmingCharacters(in: .whitespaces).isEmpty }) ?? msg
+                LocalRunnerStore.shared.setLifecycleWarning(runner.runnerName, warning: "\u26a0 \(short)")
+            }
+            LocalRunnerStore.shared.refresh()
+        }
+    }
+
+    /// Optimistically removes the runner then delegates to `RunnerLifecycleService`. Rolls back on failure.
+    @MainActor private func performRemoval() {
+        guard let runner = runnerPendingRemoval else { return }
+        runnerPendingRemoval = nil
+        removeErrorMessage = nil
+        LocalRunnerStore.shared.optimisticallyRemove(runner.runnerName)
+        Task {
+            let ok = await Task.detached(priority: .userInitiated) {
+                RunnerLifecycleService.shared.remove(runner: runner)
+            }.value
+            if !ok {
+                LocalRunnerStore.shared.optimisticallyRestore(runner)
+                removeErrorMessage = "Failed to remove \"\(runner.runnerName)\". Check logs."
+            }
+            LocalRunnerStore.shared.refresh()
+        }
+    }
+
+    // MARK: - Helpers
+
+    /// Maps a runner's `statusColor` to the corresponding design-system `Color`.
+    private func localRunnerDotColor(for runner: RunnerModel) -> Color {
+        switch runner.statusColor {
+        case .running: return Color.rbSuccess
+        case .busy:    return Color.rbWarning
+        case .idle:    return Color.rbTextTertiary
+        case .offline: return Color.rbDanger
+        }
+    }
+
+    /// Returns the configured `AddRunnerSheet` for use as a `.sheet` content closure.
+    private func addRunnerSheet() -> some View {
+        AddRunnerSheet(isPresented: $showAddRunnerSheet) { localRunnerStore.refresh() }
+    }
+
+    /// Pre-configured `RemovalAlertModifier` wired to `runnerPendingRemoval`.
+    private var removalAlertModifier: RemovalAlertModifier {
+        RemovalAlertModifier(
+            title: removalAlertTitle,
+            isPresented: Binding(
+                get: { runnerPendingRemoval != nil },
+                set: { if !$0 { runnerPendingRemoval = nil } }
+            ),
+            isAuthenticated: isAuthenticated,
+            onCancel: { runnerPendingRemoval = nil },
+            onConfirm: performRemoval
+        )
+    }
+
+    /// Builds the `RunnerDetailPopover` with commit/cancel wiring.
+    @ViewBuilder
+    private func runnerEditingPopover(runner: RunnerModel) -> some View {
+        RunnerDetailPopover(
+            runner: runner,
+            commitError: commitError,
+            onCommit: { draft in
+                guard !isCommitting else { return }
+                isCommitting = true
+                commitError = nil
+                // Build original from disk so the dirty-check in commitRunnerEdit
+                // compares against actual persisted values, not model defaults.
+                // (#1001 fix: was RunnerEditDraft(runner: runner) which left
+                // autoUpdate=true and proxy fields empty regardless of disk state.)
+                var original = RunnerEditDraft(runner: runner)
+                if let installPath = runner.installPath {
+                    original.load(installPath: installPath)
+                }
+                commitRunnerEdit(runner: runner, draft: draft, original: original) { @MainActor result in
+                    isCommitting = false
+                    switch result {
+                    case .success:
+                        localRunnerStore.refresh()
+                        editingRunner = nil
+                    case .failure(let msgs):
+                        commitError = msgs.joined(separator: "\n")
+                    }
+                }
+            },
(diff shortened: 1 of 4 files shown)


eonist:
## Deep Review — feat: extract LocalRunnersView + ScopesView from SettingsView

Overall: this is a clean, well-scoped extraction and the PR description is exemplary. The checklist is thorough and the author clearly understands the codebase. Below are the issues I found after reading the full diff + original files closely — I've deliberately excluded anything that looked like a concern but verified away.

---

### 🔴 Real Issues

#### 1. `import Combine` is unused in `LocalRunnersView.swift`

`LocalRunnersView` no longer holds any `AnyCancellable`, `Publisher`, or `sink` — all of that stayed in `SettingsView` or was simply dropped. The `import Combine` at the top of the new file is dead. SwiftLint's `unused_import` rule (if enabled) will flag this, and it's noise in the dependency graph.

```swift
// LocalRunnersView.swift — line 3
import Combine   // ← nothing in this file uses Combine
```

**Fix:** Remove it.

---

#### 2. `scopeMutateCancellable` removal changes the `ScopeEditSheet` → `RunnerStore` update path in a subtle way

In the old `SettingsView`, the pattern was:

```swift
scopeMutateCancellable = ScopeStore.shared.didMutate
    .sink { [weak store] in store?.reload() }
```

This subscribed to a *publisher* — meaning **any** mutation to `ScopeStore` (from any code path, including `ScopeEditSheet.confirmSave()`) would trigger `RunnerStore.reload()`.

In `ScopesView`, the new pattern is explicit `RunnerStore.shared.start()` calls at the **toggle** and **remove** button tap sites. That covers those two paths.

However, `ScopeEditSheet.confirmSave()` does not appear to call `RunnerStore.shared.start()` (Greptile's summary confirms this — "correctly omits the call because it writes only to `ScopePreferencesStore`"). But the alias/display-name changes made in `ScopeEditSheet` can affect polling behavior if the scope URL itself changes. **Whether `confirmSave` can change the scope's URL (not just alias) matters here.** If it can mutate the URL, the old `didMutate` subscriber would have caught that and triggered a reload; the new explicit call pattern misses it. This is worth a deliberate call-out even if you've confirmed it's safe — the old subscriber was a catch-all safety net and its removal should be documented in a comment or issue note.

---

#### 3. `@StateObject` on `.shared` singletons — correct choice, but asymmetric

`LocalRunnersView` and `ScopesView` both use `@StateObject` for their stores:

```swift
@StateObject private var localRunnerStore = LocalRunnerStore.shared
@StateObject private var scopeStore = ScopeStore.shared
```

`@StateObject` is the right call here *because* the view is the root owner of that observation subtree when it appears. `@ObservedObject` would be valid too, but `@StateObject` prevents re-initialization if a parent re-renders. The concern other reviewers flagged (Octopus) about "unconventional use of `@StateObject` on singletons" is technically a false positive — SwiftUI's `@StateObject` on a singleton is fine and avoids retain issues. **Not a bug.**

What *is* slightly asymmetric: `SettingsView` still retains `@StateObject private var notifications = NotificationPreferences.shared` and `@StateObject private var settings = AppPreferencesStore.shared` using the same pattern. So this is consistent, not novel.

---

#### 4. Section order change in `sectionsStack` — UX regression risk

The old order was:
```
Local Runners → Remote Scopes → Notifications → General → Account → About
```

The new order (from the diff) is:
```
Account → Management → General → About
```

The PR description states the order is `manageLocalRunnersRow → manageScopesRow → generalSection → accountSection → aboutSection`, but the actual diff in `sectionsStack` shows:

```swift
accountSection
Divider()
managementSection
Divider()
generalSection
Divider()
aboutSection
```

That puts **Account first**, above Management. The PR description says management rows come first. One of these is wrong. This is a UX decision more than a bug, but it's inconsistent with the stated intent and should be confirmed deliberately.

---

#### 5. `ScopesView` has no `onAppear` refresh

`LocalRunnersView` correctly calls `localRunnerStore.refresh()` on `.onAppear`:

```swift
.onAppear { localRunnerStore.refresh() }
```

`ScopesView` has no `onAppear` at all. When the user drills into ScopesView after the app has been open for a while, the displayed scope list is whatever was last loaded. If `ScopeStore` is updated externally (another app session, iCloud sync, etc.) between the user opening Settings and navigating to ScopesView, the list could be stale. This may be intentional (scopes are managed locally and `ScopeStore` likely updates reactively), but it's asymmetric with the runner view and worth a conscious comment.

---

#### 6. `localRunnerStore.refresh()` called in both `onAppear` AND `onAppearAction`

Looking at the removed code in `SettingsView.onAppearAction`:

```swift
// was removed:
localRunnerStore.refresh()
```

That call was in `onAppearAction`, which fires on `.onAppear` of `SettingsView`. Now `LocalRunnersView.onAppear` fires its own `localRunnerStore.refresh()`. 

**This is correct** — the refresh now fires when the dedicated view appears, not always when settings opens. This is an improvement; just flagging it as intentional and verified.

---

### 🟡 Minor / Style

#### 7. `removalAlertTitle` computed prop vs. inline string

`LocalRunnersView` extracts `removalAlertTitle` as a computed property:

```swift
private var removalAlertTitle: String {
    let name = runnerPendingRemoval?.runnerName ?? "this runner"
    return "Remove runner \"\(name)\"?"
}
```

This is the same as in the old `SettingsView`. It's fine, but since `runnerPendingRemoval` is read here and also in `removalAlertModifier`, there's technically two reads. Not a problem given it's a simple `@State` read, but could be collapsed if ever cleaned up.

#### 8. `isCommitting` guard in `runnerEditingPopover` is not reset on `onCancel`

```swift
onCancel: {
    commitError = nil
    editingRunner = nil
    // isCommitting is NOT reset here
}
```

If `onCommit` fires and sets `isCommitting = true`, and then the user somehow cancels before the async work resolves (edge case, since the popover stays up during commit), `isCommitting` would be stuck `true` until the async closure calls back. In practice the popover dismissal during an in-flight commit is likely blocked by the UI, but it's worth auditing whether `RunnerDetailPopover` allows cancellation while `isCommitting == true`. This was identical behavior in the original `SettingsView` — not a regression introduced by this PR.

---

### ✅ Verified Clean (Explicitly)

- **HEIGHT/WIDTH CONTRACT comments** — preserved verbatim in `settingsBody`. ✓
- **`panelSheetState.clearRunnerSheet()` via `AppDelegate+Navigation.swift`** — PR correctly notes this is unchanged; the back-button callback pattern is maintained. ✓
- **Optimistic rollback on `performStop` with `corruptInstall`** — The old `SettingsView` did *not* roll back `isRunning` on `corruptInstall` (only `failed` did). `LocalRunnersView` preserves this exact same asymmetry — not a new bug introduced here. ✓
- **`scopeMutateCancellable` → explicit `RunnerStore.shared.start()` at toggle/remove** — Functional equivalence confirmed for those two paths. ✓
- **No `NavState` / `AppDelegate+Navigation` changes** — confirmed by diff scope. ✓

---

### Summary

| # | Severity | File | Issue |
|---|---|---|---|
| 1 | 🔴 Low (lint) | `LocalRunnersView.swift` | Dead `import Combine` |
| 2 | 🟠 Medium | `ScopesView.swift` | `scopeMutateCancellable` removal — verify `ScopeEditSheet.confirmSave()` cannot change scope URLs |
| 4 | 🟠 Medium | `SettingsView.swift` | `sectionsStack` order in diff contradicts PR description — confirm Account-first is intentional |
| 5 | 🟡 Low | `ScopesView.swift` | No `onAppear` refresh — asymmetric with `LocalRunnersView`, document if intentional |
| 7–8 | 🟢 Nit | Both new files | `removalAlertTitle` double-read; `isCommitting` not cleared in `onCancel` (pre-existing) |

Clean extraction overall. The `import Combine` and the section order discrepancy are the two things I'd want addressed before merge.

eonist:
Thanks for the review — both flagged concerns are already handled in the code, so pushing back on these:

**`scopeMutateCancellable` removal is not a regression.** The `ScopeStore.didMutate → RunnerStore` reload path is fully intact; the subscription just moved from `SettingsView` to `AppDelegate+PanelSetup.swift`. That relocation is intentional: `AppDelegate` is a lifetime singleton, so the subscription stays alive even when the Settings window is closed (`SettingsView` would tear it down on dismiss). This mirrors the same fix already applied to the sign-out subscription. `didMutate.send()` still fires on every structural add/remove in `ScopeStore`.

**`@StateObject` is deliberate, not an oversight — `@ObservedObject` would be a regression.** These stores are singletons assigned inline. `@ObservedObject` re-creates its subscription wrapper on every render, which previously caused an infinite glitchy render loop. There's an in-code comment documenting exactly this. Switching to `@ObservedObject` as suggested would reintroduce that bug.

No code change needed for either item.

eonist:
## Deep Review — feat: extract LocalRunnersView + ScopesView from SettingsView

Overall: this is a clean, well-motivated extraction. The navigation model (flag-swap in `body`, lifecycle modifiers on the root `Group`) is the correct approach and the commit decomposition is easy to follow. The checklist items are all verifiable in the diff. Below are real issues I found after reading the diff end-to-end and cross-referencing against both new files and the residual `SettingsView`.

---

### 🔴 Real Issues

#### 1. `scopeMutateCancellable` removed from `SettingsView.onAppearAction` — `ScopesView` has no replacement

In the old `SettingsView.onAppearAction`:
```swift
scopeMutateCancellable = ScopeStore.shared.didMutate
    .sink { [weak store] in store?.reload() }
```

This subscription was the mechanism that triggered a `RunnerStore.reload()` whenever scopes mutated. In `ScopesView` the add/remove/toggle actions call `RunnerStore.shared.start()` directly after mutations, which covers the immediate UI feedback path. However, the `didMutate` sink was also the safety net for mutations that arrive from *other* code paths (e.g., deep-link or programmatic scope changes). Removing it entirely without confirming `RunnerStore.shared.start()` is a full semantic equivalent (it re-fetches vs. reload) means you may have silently dropped a listener. Worth a deliberate comment or a tracked issue stating this was intentional.

#### 2. Optimistic state in `LocalRunnersView` is not cleared on `onBack`

When the user taps the back button from `LocalRunnersView`, `showLocalRunners` is set to `false` and SwiftUI will deallocate the view. However, `@StateObject private var localRunnerStore = LocalRunnerStore.shared` — a shared singleton — may have pending optimistic state (`optimisticallySetRunning`, lifecycle warnings) in flight when the user exits mid-operation (e.g., start was triggered but `Task.detached` hasn't returned yet). The view goes away, the task continues, and calls `LocalRunnerStore.shared.optimisticallySetRunning(...)` successfully — that's fine. But `removeErrorMessage` and `commitError` are local `@State` and are lost. If removal fails after back navigation, the error is silently swallowed. This is an edge case but it's new behavior vs. the previous single-screen approach where the error label was always in scope.

#### 3. `localRunnerStore.refresh()` removed from `SettingsView.onAppearAction`

The old `onAppearAction` called `localRunnerStore.refresh()`. `LocalRunnersView` calls `localRunnerStore.refresh()` in its `.onAppear`. This is correct *when the user navigates to the runners screen*, but if the user opens Settings and goes directly to General/Account without ever opening `LocalRunnersView`, the runners are never refreshed for that session. This was not the prior behavior. Whether this matters depends on whether the store has independent background refresh — worth explicitly noting in the PR that this is intentional (lazy refresh on demand vs. eager on settings open).

#### 4. `ScopesView.scopeList` shows empty state before the store has loaded

`LocalRunnersView.runnerList` correctly guards with `hasLoadedOnce` before showing the empty state:
```swift
if localRunnerStore.runners.isEmpty && !localRunnerStore.isScanning && hasLoadedOnce {
```
But `ScopesView.scopeList` has no equivalent gate:
```swift
if scopeStore.entries.isEmpty {
    Text("No remote scopes added")
```
If `ScopeStore` loads asynchronously and `entries` starts empty, the user will briefly see "No remote scopes added" before the real data appears. `LocalRunnersView` solved this correctly. `ScopesView` should apply the same pattern — add a `hasLoadedOnce` + `isLoading` guard if `ScopeStore` has one, or at minimum add a `ProgressView` on first appearance.

---

### 🟡 Minor / Nits

#### 5. Section order changed silently: the PR description says `Account → Management → General → About` but `sectionsStack` actually puts Account first

The old order was: Local Runners → Scopes → Notifications → General → Account → About.  
The new order is: Account → Management → General → About.  

The PR description documents this correctly. But this is a UX change (Account moved to the top) that isn't in the original issue scope (#1202/#1203). Fine to include, but worth calling out in release notes explicitly since users muscle-memory the old position.

#### 6. `performResume`/`performStop` log prefix changed from `SettingsView >` to `LocalRunnersView >`

Intentional and correct. Just confirming this is not a copy-paste oversight.

#### 7. `ScopesView` uses `.buttonStyle(.borderless)` on the remove button; `LocalRunnersView` uses `.buttonStyle(.plain)` on the equivalent button

```swift
// ScopesView remove button:
.buttonStyle(.borderless)

// LocalRunnersView remove button:
.buttonStyle(.plain)
```

These render differently in a `List`/`ForEach` on macOS. `.borderless` is the correct style for icon-only action buttons inside rows. `LocalRunnersView` should use `.borderless` on its remove button to be consistent, especially since the Toggle already uses `.buttonStyle(.borderless)` in both views.

#### 8. `ScopesView` body has no `.onAppear` to trigger an initial load

`LocalRunnersView` calls `localRunnerStore.refresh()` on `.onAppear`. `ScopesView` has no equivalent. If `ScopeStore` requires an initial trigger to populate `entries`, this is a bug. If it's always loaded via an earlier initialization path, a comment to that effect would prevent future confusion.

---

### ✅ Confirmed Non-Issues (verified in diff)

- **HEIGHT/WIDTH CONTRACT comments**: Correctly moved from `body` to the new `settingsBody` computed var. Comment integrity preserved verbatim.
- **`onAppear`/`onDisappear` placement on root `Group`**: Intentional and correctly documented in the inline comment. This is the right call — prevents spurious re-fires on nav.
- **`@StateObject` in `LocalRunnersView`/`ScopesView`**: Both use `.shared` singletons so they won't create duplicate stores.
- **`isAuthenticated` forwarding to `RemovalAlertModifier`**: Correctly forwarded as a plain `Bool` (not a binding), which is safe since it's only needed at alert-confirm time.
- **`showLocalRunners`/`showScopes` as `@State Bool`**: Correct — these are purely local nav flags, no need for a nav stack or observable.
- **`managementSection` Divider structure**: The inset `Divider` between the two nav rows is consistent with the rest of `SettingsView`'s row separator style.

---

**Summary**: 4 real issues — the `scopeMutateCancellable` removal (#1), silent error drop on back-nav (#2), lazy vs. eager runner refresh (#3), and missing load-guard in `ScopesView` (#4) — plus one style inconsistency (#7). The rest is solid.

eonist:
## Deep Review — PR #1207 `feat: extract LocalRunnersView + ScopesView from SettingsView`

> Read: full diff, local source files, original issues #1202 / #1203, `ScopeStore.swift`, `RunnerStore.swift`, `RunnerViewModel.swift`, `AppDelegate+PanelSetup.swift`. False-positives explicitly excluded and noted.

---

### ✅ What's solid

- **Extraction is clean and complete.** Every runner-specific and scope-specific `@State`, `@StateObject`, sheet, popover, modifier, and action method has been moved to its new owner. `SettingsView` genuinely becomes a thin nav shell.
- **Lifecycle guard is correct.** Moving `.onAppear`/`.onDisappear` to the outer `Group` in `body` rather than `settingsBody` is the right call — Keychain re-reads and `signOutCancellable` recreation no longer fire on every back-nav.
- **`hasLoadedOnce` guard preserved.** `LocalRunnersView` replicates the exact `onChange(of: localRunnerStore.isScanning)` guard to suppress the "No local runners found" flash during the initial scan. Good.
- **`commitRunnerEdit` disk-load fix (#1001) carried over verbatim.** The `original.load(installPath:)` path in `runnerEditingPopover` is present in `LocalRunnersView` — the previous regression fix wasn't accidentally dropped.
- **SwiftLint checklist passes.** `file_header`, `missing_docs`, `sorted_imports` all observed in both new files.
- **HEIGHT/WIDTH CONTRACT comment** preserved in `settingsBody` exactly as required.

---

### 🔴 Real issues

#### 1. `scopeMutateCancellable` removal silently dropped `RunnerViewModel.reload()`

**This is the most consequential change in the PR.**

In the old `SettingsView.onAppearAction()`:

```swift
scopeMutateCancellable = ScopeStore.shared.didMutate
    .sink { [weak store] in store?.reload() }   // store = RunnerViewModel (injected @ObservedObject)
```

The captured `store` is the injected `RunnerViewModel` (line 36 in the pre-PR file). That `reload()` call on `RunnerViewModel` is distinct from `AppDelegate+PanelSetup`'s own `didMutate` subscription which calls `RunnerStore.shared.start()`. `RunnerViewModel.reload()` triggers a UI-facing data refresh; `RunnerStore.start()` restarts the polling loop. They are not the same thing.

By removing `scopeMutateCancellable` entirely, adding or removing a scope while Settings is open no longer drives a `RunnerViewModel.reload()`. The popover data can go stale until the next poll tick.

**Suggested fix:** `ScopesView` should own this subscription, or — better — it's already handled at the `AppDelegate` level so verify that `RunnerViewModel.reload()` is being called downstream from `RunnerStore.start()`. If `start()` internally triggers an immediate fetch that flows back into `RunnerViewModel`, then the removal is safe. But this needs to be explicitly confirmed and commented; right now it is silently lost.

---

#### 2. `ScopesView` has no `.onAppear` refresh, but `LocalRunnersView` does

`LocalRunnersView` calls `localRunnerStore.refresh()` on appear. `ScopesView` calls nothing. `ScopeStore.entries` is loaded from `UserDefaults` in `init()` so for a freshly-loaded store this is fine. **However**, if a scope is added/removed from another path between opens, the view will display whatever was in memory without pulling fresh data. `LocalRunnerStore` requires an explicit disk scan; `ScopeStore` doesn't. This asymmetry is probably intentional — but it should be called out with a brief comment in `ScopesView.body` explaining why no `onAppear` refresh is needed, so the next reader doesn't add one unnecessarily.

---

#### 3. `Remove scope` button uses `.buttonStyle(.borderless)` — inconsistent with runner row

In `ScopesView.scopeRow`, the remove (`minus.circle`) button is `.buttonStyle(.borderless)`. In `LocalRunnersView.localRunnerRowContent`, the equivalent remove button is `.buttonStyle(.plain)`. The old `SettingsView` also used `.borderless` for scopes and `.plain` for runners. This inconsistency predates the PR but has now been promoted into two first-class views where it's more visible. Not a blocker, but worth noting while you have both files open.

---

#### 4. Section order mismatch between PR description and `sectionsStack` comment

The PR description body says:
> `sectionsStack` order is now: `manageLocalRunnersRow` → `manageScopesRow` → `generalSection` → `accountSection` → `aboutSection`

The `sectionsStack` doc comment in the actual code says:
> `Order: Account → Management → General → About`

And the code itself renders: `accountSection` → `managementSection` → `generalSection` → `aboutSection`.

The PR description text is wrong (it lists the old inline-section names, not the current order, and puts Account last). The code and the inline comment agree with each other. Low severity, but the PR description could mislead a future bisect reader.

---

#### 5. Back-navigation while a sheet is open is not guarded

In both `ScopesView` and `LocalRunnersView`, the back button calls `onBack()` unconditionally. If the user somehow triggers `onBack` while `showAddScopeSheet == true` or `showAddRunnerSheet == true` (e.g. via keyboard or a future toolbar shortcut), SwiftUI will dismiss both the sheet and the entire view in one frame. The sheet's `isPresented` binding points into the child view's `@State`, which is about to be deallocated. In practice this is unlikely with the current UI, but it's a latent issue worth a `guard !showAddScopeSheet else { return }` or similar.

---

### 🟡 Minor / low severity

- **`performResume` / `performStop` log prefix changed** from `"SettingsView >"` to `"LocalRunnersView >"`. Correct — just noting it so log-search queries may need updating.
- **`ScopesView` has no `isScanning` concept.** `ScopeStore` is synchronous/UserDefaults-backed so there's no in-flight state to show. This is correct as-is.
- **The `scopeList` empty state shows immediately** (no `hasLoadedOnce` guard needed) because `ScopeStore.entries` is populated synchronously in `init()`. ✅ Intentional.
- **`popoverArrowRow` extraction** into its own computed var is a nice cleanup — makes `generalSection` structurally uniform.

---

### Summary

The refactor is architecturally sound and achieves what issues #1202 and #1203 asked for. The main thing to resolve before merge is **issue 1** above — confirm whether the `RunnerViewModel.reload()` call that was silently removed from `scopeMutateCancellable` is covered by another path, and document it. Everything else is either minor or cosmetic.

eonist:
## Updated Review — after re-reading source files + author response

> Supersedes my earlier comment. I re-read all three files from the PR branch directly, cross-referenced the author's response, and explicitly re-evaluated every finding. Retractions noted inline.

---

### Retractions from the first review

**~~Issue #1 — `scopeMutateCancellable` silently dropped~~** → **Retracted.**
The author confirmed: the `ScopeStore.didMutate → RunnerStore` subscription was *moved* to `AppDelegate+PanelSetup.swift`, not dropped. It now lives in a lifetime singleton, which is actually the *better* location (the old `SettingsView`-owned subscription was torn down every time the settings panel closed). ✅ Not a bug.

**~~Issue #4 — `ScopesView.scopeList` shows empty state before store loads~~** → **Retracted.**
I assumed `ScopeStore` might load asynchronously. Reading the actual live file: `ScopeStore.entries` is populated synchronously from `UserDefaults` in `init()`. There is no async load phase. The `hasLoadedOnce` guard exists in `LocalRunnersView` because `LocalRunnerStore` does an actual disk scan (async). `ScopeStore` does not. No guard needed. ✅ False flag.

**~~Issue #7 — `.buttonStyle(.borderless)` vs `.plain` inconsistency~~** → **Retracted.**
Both views use `ScrollView` + `ForEach`, not `List`. Outside a `List`, `.borderless` and `.plain` are visually identical on macOS — both suppress the button bezel. The difference only materialises inside `List` rows. ✅ False flag.

---

### Surviving real issues (confirmed against actual source files)

#### 🔴 Issue 1 — Silent error drop on back-nav from `LocalRunnersView`

`performRemoval()` launches a `Task` that captures `removeErrorMessage` (a `@State` var on the view struct). If the task is in-flight (i.e. `RunnerLifecycleService.shared.remove(runner:)` hasn't returned yet) and the user navigates back — setting `showLocalRunners = false`, deallocating the view — the task continues. When it resolves with `!ok`, it tries to set `removeErrorMessage = "Failed to remove ..."` on `LocalRunnerStore.shared.optimisticallyRestore(runner)` (that part's fine, it hits the singleton), **but** the `removeErrorMessage = ...` line writes to a `@State` that is now dangling. In practice SwiftUI will silently drop the write. The error is never surfaced to the user.

Same applies to `commitError` in `runnerEditingPopover`'s `onCommit` path.

This is genuinely new behaviour compared to the single-screen approach, where the error label was always in scope. It's an edge case (requires back-nav mid-flight removal) but the failure mode is invisible.

**Suggested mitigation:** Check `runnerPendingRemoval != nil` in the `onBack` callback and either block navigation while a removal is in flight, or surface the error through `LocalRunnerStore` itself rather than local `@State`.

---

#### 🟡 Issue 2 — `localRunnerStore.refresh()` now lazy (behaviour change, not a bug)

Previously `SettingsView.onAppearAction()` called `localRunnerStore.refresh()` unconditionally — so every time the settings panel opened, the runner list was refreshed. Now it only refreshes when the user navigates into `LocalRunnersView`. If the user opens Settings, checks General, and closes — no scan happens for that session.

This is likely intentional (lazy scan on demand is more efficient), but it is a behaviour change. If `LocalRunnerStore` has an independent background refresh cadence, this is fine. If it only ever refreshes on explicit calls, users who never visit the runner screen will see a potentially stale count elsewhere in the app.

Not a blocker — just worth a one-line comment in `onAppearAction` confirming the lazy-load is deliberate.

---

#### 🟡 Issue 3 — Section order discrepancy between PR description and actual code

The PR description (Commits section) states:
> `sectionsStack` order is now: `manageLocalRunnersRow → manageScopesRow → generalSection → accountSection → aboutSection`

The actual `sectionsStack` in the live file renders:
```
accountSection → managementSection → generalSection → aboutSection
```

The code and its inline comment (`Order: Account → Management → General → About`) agree with each other — the PR description text is just wrong (it lists the old inline-section names and puts Account last). Low severity — the code is correct, only the description is misleading for future bisect readers.

---

#### 🟡 Issue 4 — `ScopesView` has no `onAppear` comment explaining why

`LocalRunnersView` has `.onAppear { localRunnerStore.refresh() }` with an obvious rationale. `ScopesView` has nothing. A future dev might reasonably wonder if the omission is intentional or a bug. A single comment — e.g. `// ScopeStore.entries is loaded synchronously from UserDefaults; no onAppear refresh needed` — would prevent that confusion.

---

### Summary table

| # | Severity | Retracted? | File | Issue |
|---|---|---|---|---|
| 1 | 🔴 Edge-case | No | `LocalRunnersView.swift` | Back-nav silently drops in-flight removal error |
| 2 | 🟡 Behaviour change | No | `SettingsView.swift` | Runner refresh now lazy — document as intentional |
| 3 | 🟡 Docs only | No | `SettingsView.swift` | PR description text has wrong section order |
| 4 | 🟡 Nit | No | `ScopesView.swift` | No comment explaining why `onAppear` is absent |
| orig-1 | — | ✅ Retracted | — | `scopeMutateCancellable` — moved to AppDelegate, not dropped |
| orig-4 | — | ✅ Retracted | — | `ScopesView` empty-state flash — false flag, sync load |
| orig-7 | — | ✅ Retracted | — | `.borderless` vs `.plain` — identical outside `List` |

The extraction itself is correct and clean. Issue #1 above (silent error drop) is the only thing I'd want addressed or consciously accepted before merge. The rest are documentation/comment nits.

eonist:
## Review correction — false positives retracted

After tracing the full call chain through the local source, I need to retract two findings from my previous comment.

---

### ~~Issue 1~~ — RETRACTED ✅

I claimed that removing `scopeMutateCancellable` silently dropped `RunnerViewModel.reload()`. This is wrong.

The full chain is:

```
ScopeStore.didMutate
  → AppDelegate sink (AppDelegate+PanelSetup.swift:221)
  → RunnerStore.shared.start()
  → Task { await fetch() }
  → applyFetchResult() (RunnerStore.swift:278)
  → didUpdate.send()
  → AppDelegate sink (AppDelegate+PanelSetup.swift:195)
  → self.observable.reload()   // RunnerViewModel
```

`RunnerStore.start()` already drives a `RunnerViewModel.reload()` as a downstream consequence via `didUpdate`. The old `scopeMutateCancellable` was calling `reload()` on a stale snapshot *before* the new fetch landed — it was actually a premature/redundant read. **The removal is correct and safe.**

---

### ~~Issue 3~~ — RETRACTED ✅

I flagged back-navigation while a sheet is open as a latent crash risk. This is wrong. The back button is visually and interactively unreachable while a modal sheet covers the view. When `onBack` does fire, SwiftUI tears down `ScopesView`/`LocalRunnersView` and all their owned `@State` (including `showAddScopeSheet`) in an orderly single pass — no dangling binding, no crash.

---

### What actually stands

| # | Finding | Severity |
|---|---------|----------|
| 2 | `ScopesView` has no comment explaining why there's no `.onAppear` refresh, unlike `LocalRunnersView`. The asymmetry is intentional (UserDefaults-backed sync init) but unexplained. | 🟡 Minor |
| 4 | PR description body lists the old inline-section names in the wrong order. Code and inline `sectionsStack` doc comment are both correct. | 🟡 Cosmetic |
| 5 | `.buttonStyle(.borderless)` on the scope remove button vs `.buttonStyle(.plain)` on the runner remove button. Pre-existing inconsistency now promoted into two first-class views. | 🟡 Cosmetic |

Everything else in this PR is solid. Good to merge once you're happy with the above.

eonist:
@ -0,0 +1,102 @@
=== COMMENT 4652517254 ===
## Updated Review — after re-reading source files + author response

> Supersedes my earlier comment. I re-read all three files from the PR branch directly, cross-referenced the author's response, and explicitly re-evaluated every finding. Retractions noted inline.

---

### Retractions from the first review

**~~Issue #1 — `scopeMutateCancellable` silently dropped~~** → **Retracted.**
The author confirmed: the `ScopeStore.didMutate → RunnerStore` subscription was *moved* to `AppDelegate+PanelSetup.swift`, not dropped. It now lives in a lifetime singleton, which is actually the *better* location (the old `SettingsView`-owned subscription was torn down every time the settings panel closed). ✅ Not a bug.

**~~Issue #4 — `ScopesView.scopeList` shows empty state before store loads~~** → **Retracted.**
I assumed `ScopeStore` might load asynchronously. Reading the actual live file: `ScopeStore.entries` is populated synchronously from `UserDefaults` in `init()`. There is no async load phase. The `hasLoadedOnce` guard exists in `LocalRunnersView` because `LocalRunnerStore` does an actual disk scan (async). `ScopeStore` does not. No guard needed. ✅ False flag.

**~~Issue #7 — `.buttonStyle(.borderless)` vs `.plain` inconsistency~~** → **Retracted.**
Both views use `ScrollView` + `ForEach`, not `List`. Outside a `List`, `.borderless` and `.plain` are visually identical on macOS — both suppress the button bezel. The difference only materialises inside `List` rows. ✅ False flag.

---

### Surviving real issues (confirmed against actual source files)

#### 🔴 Issue 1 — Silent error drop on back-nav from `LocalRunnersView`

`performRemoval()` launches a `Task` that captures `removeErrorMessage` (a `@State` var on the view struct). If the task is in-flight (i.e. `RunnerLifecycleService.shared.remove(runner:)` hasn't returned yet) and the user navigates back — setting `showLocalRunners = false`, deallocating the view — the task continues. When it resolves with `!ok`, it tries to set `removeErrorMessage = "Failed to remove ..."` on `LocalRunnerStore.shared.optimisticallyRestore(runner)` (that part's fine, it hits the singleton), **but** the `removeErrorMessage = ...` line writes to a `@State` that is now dangling. In practice SwiftUI will silently drop the write. The error is never surfaced to the user.

Same applies to `commitError` in `runnerEditingPopover`'s `onCommit` path.

This is genuinely new behaviour compared to the single-screen approach, where the error label was always in scope. It's an edge case (requires back-nav mid-flight removal) but the failure mode is invisible.

**Suggested mitigation:** Check `runnerPendingRemoval != nil` in the `onBack` callback and either block navigation while a removal is in flight, or surface the error through `LocalRunnerStore` itself rather than local `@State`.

---

#### 🟡 Issue 2 — `localRunnerStore.refresh()` now lazy (behaviour change, not a bug)

Previously `SettingsView.onAppearAction()` called `localRunnerStore.refresh()` unconditionally — so every time the settings panel opened, the runner list was refreshed. Now it only refreshes when the user navigates into `LocalRunnersView`. If the user opens Settings, checks General, and closes — no scan happens for that session.

This is likely intentional (lazy scan on demand is more efficient), but it is a behaviour change. If `LocalRunnerStore` has an independent background refresh cadence, this is fine. If it only ever refreshes on explicit calls, users who never visit the runner screen will see a potentially stale count elsewhere in the app.

Not a blocker — just worth a one-line comment in `onAppearAction` confirming the lazy-load is deliberate.

---

#### 🟡 Issue 3 — Section order discrepancy between PR description and actual code

The PR description (Commits section) states:
> `sectionsStack` order is now: `manageLocalRunnersRow → manageScopesRow → generalSection → accountSection → aboutSection`

The actual `sectionsStack` in the live file renders:
```
accountSection → managementSection → generalSection → aboutSection
```

The code and its inline comment (`Order: Account → Management → General → About`) agree with each other — the PR description text is just wrong (it lists the old inline-section names and puts Account last). Low severity — the code is correct, only the description is misleading for future bisect readers.

---

#### 🟡 Issue 4 — `ScopesView` has no `onAppear` comment explaining why

`LocalRunnersView` has `.onAppear { localRunnerStore.refresh() }` with an obvious rationale. `ScopesView` has nothing. A future dev might reasonably wonder if the omission is intentional or a bug. A single comment — e.g. `// ScopeStore.entries is loaded synchronously from UserDefaults; no onAppear refresh needed` — would prevent that confusion.

---

=== COMMENT 4652517616 ===
## Review correction — false positives retracted

After tracing the full call chain through the local source, I need to retract two findings from my previous comment.

---

### ~~Issue 1~~ — RETRACTED ✅

I claimed that removing `scopeMutateCancellable` silently dropped `RunnerViewModel.reload()`. This is wrong.

The full chain is:

```
ScopeStore.didMutate
  → AppDelegate sink (AppDelegate+PanelSetup.swift:221)
  → RunnerStore.shared.start()
  → Task { await fetch() }
  → applyFetchResult() (RunnerStore.swift:278)
  → didUpdate.send()
  → AppDelegate sink (AppDelegate+PanelSetup.swift:195)
  → self.observable.reload()   // RunnerViewModel
```

`RunnerStore.start()` already drives a `RunnerViewModel.reload()` as a downstream consequence via `didUpdate`. The old `scopeMutateCancellable` was calling `reload()` on a stale snapshot *before* the new fetch landed — it was actually a premature/redundant read. **The removal is correct and safe.**

---

### ~~Issue 3~~ — RETRACTED ✅

I flagged back-navigation while a sheet is open as a latent crash risk. This is wrong. The back button is visually and interactively unreachable while a modal sheet covers the view. When `onBack` does fire, SwiftUI tears down `ScopesView`/`LocalRunnersView` and all their owned `@State` (including `showAddScopeSheet`) in an orderly single pass — no dangling binding, no crash.

---

### What actually stands

- `ScopesView` has no comment explaining why there's no `.onAppear` refresh; minor.
- PR description body has the wrong section order; cosmetic.

PR #1302, opened 2026-06-10
refactor(phase1): Replace ObservableObject/@Published with @Observable (#1296)

eonist:
## Summary

Phase 1 of the [🏗️ Modernise data model architecture — Swift 6.2 actor-based refactor](https://github.com/eoncode/runner-bar/issues/1287) epic.

Closes #1296

---

## What this PR does

Replaces `ObservableObject` / `@Published` / Combine observation with the Swift 5.9+ `@Observable` macro across the relevant state types, one commit at a time.

This is a **pure observation-model migration** — no actor isolation changes, no behavioural changes. Each commit targets one type to keep diffs reviewable and CI failures easy to bisect.

---

## Commits in this PR

| Commit | Type | File |
|--------|------|------|
| `66861df8` | `@Observable` conversion | `PanelSheetState` |
| _(next)_ | `@Observable` + env migration | `PanelVisibilityState` + 3 consuming views |
| _(next)_ | `@Observable` conversion | `NotificationPreferences` + `SettingsView` |
| _(next)_ | `@Observable` conversion | `AppPreferencesStore` + `SettingsView` |
| _(next)_ | `@Observable` conversion | `SystemStatsViewModel` + `PanelMainView` |
| _(next)_ | `@Observable` conversion | `ScopeStore` + Combine sink replacement |
| _(next)_ | `@Observable` bridge | `RunnerViewModel` + `AppDelegate+PanelSetup` |

---

## CI checklist

- [ ] SwiftLint — 0 violations per commit
- [ ] Swift build clean
- [ ] Swift test passes
- [ ] Periphery — no new dead-code warnings

---

## Notes for reviewer

- `PanelSheetState` has **zero call-site changes** — it is only accessed imperatively from `AppDelegate`/`AppDelegate+Navigation`, never held as a SwiftUI property wrapper in any view.
- Subsequent commits touching `PanelVisibilityState` will migrate `@EnvironmentObject` → `@Environment` in `PanelMainView`, `PanelContainerView`, and `InlineJobRowsView` atomically in one commit to avoid a broken intermediate state.
- `RunnerStore`'s Combine sinks (`intervalCancellable`, `scopeCancellable`) are **not** touched in Phase 1 — those are Phase 2 actor work.

___

## **CodeAnt-AI Description**
**Keep runner sheet state stable when the panel hides and reopens**

### What Changed
- The runner detail sheet state now uses Swift observation instead of the older Combine-based setup
[...]

Issue #1287, opened 2026-06-10
🏗️ Modernise data model architecture — Swift 6.2 actor-based refactor


eonist:
## Overview

Tracks a full modernisation of RunnerBar's data model and storage to idiomatic Swift 6.2.
The current code is directionally correct but uses several Swift 5.9-era patterns.

## Current state (problems to fix)

| Area | Current pattern | Problem |
|---|---|---|
| Stores | `@MainActor final class` singletons | Forces all store work — incl. network + file I/O — onto the main thread |
| View model | `ObservableObject` + `@Published` + Combine | Coarse-grained re-renders; requires `@EnvironmentObject`; drags in Combine for UI bridging |
| File I/O | `[String: Any]` + `JSONSerialization` free functions | No schema safety, no compiler help on key names, untestable |
| Commit logic | `commitRunnerEdit` free async function | Mixes 3 I/O concerns (API, JSON, proxy files); untestable |
| Draft loading | `mutating func loadRunnerJSON` reads disk inside the model | Disk I/O in a value type; no injection seam for tests |
| Store access in views | `LocalRunnerStore.shared` accessed globally | Tight coupling, no isolation guarantee |

### Layer 2 — `@Observable` view model
```swift
@Observable @MainActor
final class RunnerViewModel {
    var runners: [Runner] = []
    var localRunners: [RunnerModel] = []
}
```

### Layer 3 — `actor` stores
```swift
actor RunnerStore {
    private(set) var runners: [Runner] = []
    func fetch() async { ... await MainActor.run { viewModel.runners = result } }
}
```

### Layer 4 — Typed `Codable` config + file actor
```swift
struct RunnerConfig: Codable, Sendable {
    var workFolder: String; var disableUpdate: Bool
    var platform: String?; var agentVersion: String?
}
actor RunnerConfigStore {
    func load(at: String) async throws -> RunnerConfig
    func save(_: RunnerConfig, at: String) async throws
}
```

### Layer 5 — Use-case struct replaces free function
```swift
struct SaveRunnerEditsUseCase: Sendable {
    let configStore: RunnerConfigStore
    let proxyStore: RunnerProxyStore
    let labelsService: any RunnerLabelsService
    func execute(runner:draft:original:) async -> CommitResult
}
```

## Guiding principles

- No `@unchecked Sendable` — compiler must enforce all isolation
- No Combine in view-model layer after Phase 1
- No `[String: Any]` after Phase 3
- Each phase ships green CI before the next begins

[...]

Issue #1281, opened 2026-06-10
⭐ Invesigate the data model in this project.


eonist:
How are we dealign with data? I know there are some vie models. and others dont use that. 

are there any improvment opertunities?

eonist:
[...]

PR #1303, opened 2026-06-10
refactor(phase2): Convert RunnerStore + LocalRunnerStore from @MainActor class to actor (#1287)

eonist:
## **User description**
Part of the Swift 6.2 data model modernisation epic #1287.
Stacked on Phase 1 — base branch is `feature/1296-observable-refactor`.

## Goal

Convert `RunnerStore` and `LocalRunnerStore` from `@MainActor final class` singletons to true `actor` types so that fetch, poll, and enrichment work runs on a background executor. UI state is pushed back to `RunnerViewModel` via `await MainActor.run { }`. The compiler then enforces that no background work touches UI state and no UI code blocks on I/O — no `@unchecked Sendable` anywhere.

### 1. ViewModel injection — no singleton cross-references inside actors

Each actor receives `RunnerViewModel` at init time. No singleton access (`AppDelegate.shared`, `.observable`) inside actor bodies.

```swift
actor RunnerStore {
    private let viewModel: RunnerViewModel  // strong ref, owned by AppDelegate

    init(viewModel: RunnerViewModel) {
        self.viewModel = viewModel
    }

    func fetch() async {
        let result = await fetchAndEnrichRunners(...)
        await MainActor.run { [viewModel] in
            viewModel.runners = result
        }
    }
}
```

`AppDelegate` creates and wires both:
```swift
let observable = RunnerViewModel()
let runnerStore = RunnerStore(viewModel: observable)
let localRunnerStore = LocalRunnerStore(viewModel: observable)
```

### 2. Combine sinks replaced with AsyncStream — zero Combine in the store layer

`RunnerStore.init` currently holds `intervalCancellable` and `scopeCancellable` — Combine sinks that restart polling when `pollingInterval` or active scopes change. On a background `actor`, Combine sinks in `init` are the wrong pattern.

Replacement: `AsyncStream` computed vars driven by `withObservationTracking` (pure Swift Observation — no `PassthroughSubject`, no `AnyCancellable`):

```swift
// Generic helper — works for any @Observable type
func observationStream<Root, T>(
    root: Root,
    keyPath: KeyPath<Root, T>
) -> AsyncStream<T> {
    AsyncStream { continuation in
        func observe() {
            withObservationTracking {
                continuation.yield(root[keyPath: keyPath])
            } onChange: {
                Task { observe() }
            }
        }
        observe()
    }
}
```

Consumed inside the actor:
```swift
func startObservingPreferences() {
    Task {
        for await _ in observationStream(root: AppPreferencesStore.shared,
                                         keyPath: \.pollingInterval).dropFirst() {
            await self.start()
        }
    }
}
```

This eliminates `didChangePollingInterval` and `didMutate` `PassthroughSubject` bridges entirely by the end of this phase.

### 3. LocalRunnerStore — drop ObservableObject + @Published

`runners` and `isScanning` become plain stored properties on the actor. The `AppDelegate+PanelSetup` Combine sink on `.$runners` is removed — the actor pushes directly to `viewModel.localRunners` via `MainActor.run`.

## Commit plan

| # | Commit | Files |
|---|---|---|
| 1 | `RunnerStore`: `@MainActor class` → `actor`, inject `viewModel`, replace Combine init sinks with `AsyncStream` observation tasks | `RunnerStore.swift` |
| 2 | `LocalRunnerStore`: `@MainActor class + ObservableObject` → `actor`, inject `viewModel`, push results via `MainActor.run`, remove `@Published` | `LocalRunnerStore.swift` |
| 3 | `AppDelegate+PanelSetup`: remove `.$runners` Combine sink, update `start()` call sites to `Task` wrappers, wire `viewModel` injection at init | `AppDelegate+PanelSetup.swift`, `AppDelegate.swift` |
| 4 | `AppPreferencesStore` + `ScopeStore`: replace `PassthroughSubject` bridges with `AsyncStream` computed vars, remove `didChangePollingInterval` + `didMutate` subjects | `AppPreferencesStore.swift`, `ScopeStore.swift` |

## Guiding principles (from #1287)

- No `@unchecked Sendable` — compiler enforces all isolation
- No Combine in the store/actor layer after this phase
- Each commit ships green CI
- No singleton access inside actor bodies

## Acceptance criteria

- [ ] `RunnerStore` and `LocalRunnerStore` are `actor` types (no `@MainActor` on declaration)
- [ ] All network I/O and enrichment runs off the main thread
- [ ] `RunnerViewModel` is updated only via `await MainActor.run { }` or `@MainActor` func calls
- [ ] No `AnyCancellable` or `PassthroughSubject` in either store
- [ ] No `@Published` in `LocalRunnerStore`
- [ ] No `.$runners` Combine sink in `AppDelegate+PanelSetup`
- [ ] No data races under Swift concurrency runtime checker
- [ ] SwiftLint: 0 violations
- [ ] Periphery: no new warnings
- [ ] `swift test` passes

Ref: #1287, stacked on #1302

___

## **CodeAnt-AI Description**
**Move runner updates to actor-backed snapshots and remove Combine bridges**

### What Changed
- Local runner data now updates through a shared app view model, so the local runners list and scanning state stay in sync with the UI without relying on direct store reads.
- Adding, removing, starting, and stopping local runners now refreshes the list immediately and keeps the row status up to date while actions are running.
- Scope changes now restart polling automatically, and the settings screen no longer triggers duplicate restarts when toggling or removing scopes.
- The status-bar icon is now based on the latest pushed runner snapshot, and startup waits for the local runner scan before beginning polling so the first refresh has local runner data available.

### Impact
`✅ Faster local runner list updates`
`✅ Fewer stale runner rows after add/remove/start/stop`
`✅ More reliable polling after scope and sign-out changes`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

## Release Notes

* **Refactor**
  * Modernized internal concurrency architecture to use Swift actors and observation tracking, replacing legacy Combine patterns. This improves thread safety and ensures reliable state synchronization across the app when managing runners, scopes, and polling intervals. Status updates, local runner detection, and preference changes now propagate more consistently.


eonist:
read our intent: https://github.com/eoncode/runner-bar/issues/1287 the idea is to move to as modern code as possible. also discussion: https://github.com/eoncode/runner-bar/issues/1281#issuecomment-4672073138 and https://github.com/eoncode/runner-bar/issues/1281#issuecomment-4672698201

eonist:
eon@eons-MacBook-Pro % greptile review
Review complete in 6m 48s - 3 comments across 2 files - commit 675e941ce24c

╭──────────────────────────────────────────────────────────────────────────────╮
│ Greptile Summary                                                             │
│ This PR migrates the app's observable state layer from                       │
│ Combine/ObservableObject to Swift Observation (@Observable), and             │
│ simultaneously converts RunnerStore from a @MainActor class to a plain Swift │
│  actor that pushes state to RunnerViewModel.shared after every poll cycle    │
│ instead of broadcasting via PassthroughSubject.                              │
│                                                                              │
│ - RunnerStore → Swift actor: Removes the didUpdate Combine subject and       │
│ intervalCancellable/scopeCancellable subscriptions; replaces them with       │
│ withObservationTracking async streams and direct await MainActor.run {}      │
│ pushes to RunnerViewModel.shared at the end of each fetch cycle. start() and │
│  nextPollInterval() retain synchronous signatures but now execute on the     │
│ actor's background executor.                                                 │
│ - ObservableObject → @Observable across all stores and view-models           │
│ (AppPreferencesStore, NotificationPreferences, ScopeStore,                   │
│ PanelVisibilityState, PanelSheetState, SystemStatsViewModel,                 │
│ RunnerViewModel); all SwiftUI consumers updated from                         │
│ @StateObject/@ObservedObject/@EnvironmentObject to @State/plain              │
│ properties/@Environment.                                                     │
│ - AggregateStatus.init(runners:) extracted into RunnerBarCore so both the    │
│ actor and the status-item code can derive the aggregate state from the same  │
│ snapshot without touching the actor.                                         │
│                                                                              │
│ Confidence: 3/5                                                              │
│ Not safe to merge without fixing the MainActor.assumeIsolated calls in       │
│ start() and nextPollInterval() — they will crash on every app launch.        │
│                                                                              │
│ The concurrency model in RunnerStore has a fundamental mismatch: start() and │
│  nextPollInterval() use MainActor.assumeIsolated to read from                │
│ main-actor-isolated stores synchronously, but both methods now execute on    │
│ the actor's own background executor rather than the main actor.              │
│ assumeIsolated fires a fatal precondition when the current executor is not   │
│ the main actor, so the very first poll-loop start — and every subsequent     │
│ restart — will crash the app. The rest of the migration (all @Observable     │
│ conversions, the push-based RunnerViewModel.shared wiring, environment       │
│ injection updates) looks structurally sound.                                 │
│                                                                              │
│ Sources/RunnerBar/Runner/RunnerStore.swift — specifically start() lines      │
│ 141/146 and nextPollInterval() line 184                                      │
╰──────────────────────────────────────────────────────────────────────────────╯

Sequence Diagram

  Open diagram ↗: file:///var/folders/2b/2c_967m52k5fhdyrf77_b01h0000gn/T/grept
ile-diagrams-NURtQp/464eca38c5881f0b.svg

┌ Comment 1 ·  P1  · comment · L141-L146 ──────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/RunnerStore.swift ─────────────────────────────────┤
│                                                                              │
│ 141 +         let scopes = MainActor.assumeIsolated {                        │
│     + ScopeStore.shared.activeScopes }                                       │
│ 142           log("RunnerStore › start — activeScopes=\(scopes)")            │
│ 143           if scopes.isEmpty {                                            │
│ 144               log("RunnerStore › ⚠️ start called but activeScopes is     │
│       EMPTY — actions will not load")                                        │
│ 145           }                                                              │
│ 146 +         let localCount = MainActor.assumeIsolated {                    │
│     + LocalRunnerStore.shared.runners.count }                                │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ MainActor.assumeIsolated will crash on RunnerStore's executor                │
│                                                                              │
│ RunnerStore is now a plain actor whose methods execute on its own background │
│  serial executor, not the main actor. MainActor.assumeIsolated traps with a  │
│ precondition failure whenever the current executor is not the main actor's.  │
│ Every call path to start() — await RunnerStore.shared.start() in             │
│ AppDelegate, the scope/sign-out Task wrappers in the UI, and the internal    │
│ await self?.start() calls in                                                 │
│ _startObservingPreferences/_startObservingScopes — all hop to RunnerStore's  │
│ executor before entering the method body, so this crashes unconditionally on │
│  the first poll-loop start.                                                  │
│                                                                              │
│ The same issue applies to nextPollInterval() at line 184, which is called    │
│ from the poll loop task that also runs on the actor's executor. Both sites   │
│ should use await MainActor.run { } (requiring these functions to become      │
│ async) instead of assumeIsolated.                                            │
└──────────────────────────────────────────────────────────────────────────────┘

Comments Without Local Code (2)

┌ Comment 2 ·  P2  · comment · L232-L238 ──────────────────────────────────────┐
│ didChangePollingInterval emits but has no subscribers                        │
│                                                                              │
│ The subject is sent in pollingInterval.didSet, but RunnerStore now observes  │
│ interval changes through withObservationTracking in                          │
│ _startObservingPreferences instead of subscribing to a Combine publisher.    │
│ The comment says "RunnerStore subscribes to restart its poll loop when the   │
│ interval changes," but that subscription was removed as part of this PR. The │
│  // periphery:ignore suppresses the dead-code warning, leaving the subject   │
│ silently firing into the void. If the intent is to keep a Combine bridge for │
│  future use, the comment should clarify that; if not, the field should be    │
│ removed.                                                                     │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 3 ·  P2  · comment · L432-L436 ──────────────────────────────────────┐
│ Test init(viewModel:) observes RunnerStore.shared instead of self            │
│                                                                              │
│ The DI init meant for testing kicks off observation tasks via                │
│ RunnerStore.shared._startObservingPreferences() and                          │
│ RunnerStore.shared._startObservingScopes() instead of                        │
│ self._startObservingPreferences() / self._startObservingScopes(). Accessing  │
│ RunnerStore.shared in that init will eagerly initialise the production       │
│ singleton (the static let shared lazy init runs), and any subsequent         │
│ preference/scope changes will restart shared, not the test instance. The     │
│ injected viewModel therefore only affects applyFetchResult while             │
│ observation-driven restarts silently route through the shared singleton.     │
└──────────────────────────────────────────────────────────────────────────────┘

--------------------------------------------------------------------------------
Last reviewed commit: 675e941ce24c25ff2b3b86e8fd0fd546ff0bdc09 - Review ID: 
c6776232-fb15-4ca6-8322-c37d1374bd0b
eon@eons-MacBook-Pro %

eonist:
eon@eons-MacBook-Pro % greptile review
Review complete in 8m 48s - 4 comments across 3 files - commit e652ad2456c6

╭──────────────────────────────────────────────────────────────────────────────╮
│ Greptile Summary                                                             │
│ This PR completes the Swift 6 / @Observable migration: RunnerStore is        │
│ converted from a @MainActor class to a true Swift actor, all preference and  │
│ state types drop ObservableObject/@Published in favour of @Observable, and   │
│ Combine sinks that previously pulled state from the actor are replaced by a  │
│ push model where the actor writes directly into RunnerViewModel via await    │
│ MainActor.run { }.                                                           │
│                                                                              │
│ - RunnerStore is now a proper background actor injected with RunnerViewModel │
│  and LocalRunnerStore at construction; it uses withObservationTracking-based │
│  AsyncStreams to observe AppPreferencesStore.pollingInterval and             │
│ ScopeStore.activeScopes without holding Combine subscriptions.               │
│ - All views migrate from @ObservedObject/@StateObject/@EnvironmentObject to  │
│ plain stored properties and @Environment for @Observable objects;            │
│ AggregateStatus computation is extracted into a new init(runners:) on the    │
│ core type.                                                                   │
│ - RunnerStore.applyFetchResult pushes state to the injected RunnerViewModel  │
│ instance and calls AppDelegate.updateStatusIcon() inside a single            │
│ MainActor.run block, eliminating the didUpdate Combine publisher.            │
│                                                                              │
│ Confidence: 3/5                                                              │
│ The concurrency refactor is broad and the poll-loop restart mechanism has a  │
│ silent gap on the first user-initiated polling interval change.              │
│                                                                              │
│ The isFirst guard in _startObservingPreferences drops the very first         │
│ polling-interval change in every session. Because didChangePollingInterval   │
│ also has no subscribers (flagged in a prior review round), both mechanisms   │
│ fail simultaneously — meaning the first time a user adjusts the interval     │
│ after launch, the poll loop continues at the old cadence until a second      │
│ adjustment is made. The rest of the migration is structurally sound, though  │
│ the observation chain teardown and the misleading RunnerViewModel.shared     │
│ singleton warrant attention before this ships.                               │
│                                                                              │
│ Sources/RunnerBar/Runner/RunnerStore.swift (isFirst guard, observation chain │
│  teardown), Sources/RunnerBar/Views/Runner/RunnerViewModel.swift (unused     │
│ shared singleton), Sources/RunnerBar/App/AppDelegate+Polling.swift (missing  │
│ weak capture)                                                                │
╰──────────────────────────────────────────────────────────────────────────────╯

Sequence Diagram

  Open diagram ↗: file:///var/folders/2b/2c_967m52k5fhdyrf77_b01h0000gn/T/grept
ile-diagrams-4HZ4sC/31c07cac5ef04004.svg

┌ Comment 1 ·  P2  · comment · L29-L35 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/App/AppDelegate+Polling.swift ────────────────────────────┤
│                                                                              │
│ 30               .receive(on: DispatchQueue.main)                            │
│ 31               .sink {                                                     │
│ 32                   log("AppDelegate › didSignOut — restarting poll loop    │
│      for env-token fallback")                                                │
│ 33 +                 Task { [weak self] in await self?.runnerStore.start() } │
│ 34               }                                                           │
│ 35               .store(in: &cancellables)                                   │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Missing [weak self] on the outer sink closure creates an implicit strong     │
│ capture of self. Swift will strongly capture self from the enclosing         │
│ method's scope because the inner Task { [weak self] in } references it; the  │
│ [weak self] annotation belongs to the Task closure only and doesn't shield   │
│ the outer sink closure. This produces a retain cycle:                        │
│ AppDelegate.cancellables → subscriber → outer sink closure → AppDelegate     │
│ (strong).                                                                    │
├ Suggestion ──────────────────────────────────────────────────────────────────┤
│ -         OAuthService.shared.didSignOut                                     │
│ -             .receive(on: DispatchQueue.main)                               │
│ -             .sink {                                                        │
│ -                 log("AppDelegate › didSignOut — restarting poll loop for   │
│ env-token fallback")                                                         │
│ -                 Task { [weak self] in await self?.runnerStore.start() }    │
│ -             }                                                              │
│ -             .store(in: &cancellables)                                      │
│ +         OAuthService.shared.didSignOut                                     │
│ +             .receive(on: DispatchQueue.main)                               │
│ +             .sink { [weak self] in                                         │
│ +                 log("AppDelegate › didSignOut — restarting poll loop for   │
│ env-token fallback")                                                         │
│ +                 Task { [weak self] in await self?.runnerStore.start() }    │
│ +             }                                                              │
│ +             .store(in: &cancellables)                                      │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 2 ·  P1  · comment · L88-L90 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/RunnerStore.swift ─────────────────────────────────┤
│                                                                              │
│ 88 +             var isFirst = true                                          │
│ 89 +             for await newInterval in stream {                           │
│ 90 +                 if isFirst { isFirst = false; continue }                │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ First polling-interval change silently dropped                               │
│                                                                              │
│ withObservationTracking's onChange fires only when the tracked value         │
│ *changes* — it does not emit an initial value on registration, unlike        │
│ $pollingInterval (a Combine @Published publisher), which emits the current   │
│ value immediately on subscription (and was why the old code used             │
│ .dropFirst(1)). Here, the isFirst guard therefore skips the very first real  │
│ change the user makes to the polling interval in a session. Combined with    │
│ didChangePollingInterval having no subscribers (flagged separately), that    │
│ first change goes entirely unprocessed — the poll loop keeps running at the  │
│ old interval until the interval is changed a *second* time.                  │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 3 ·  P2  · comment · L64-L95 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/RunnerStore.swift ─────────────────────────────────┤
│                                                                              │
│ 64 +     func _startObservingPreferences() {                                 │
│ 65 +         intervalObservationTask?.cancel()                               │
│ 66 +         intervalObservationTask = Task { [weak self] in                 │
│ 67 +             // Build an AsyncStream that fires on every pollingInterval │
│    +  change.                                                                │
│ 68 +             // withObservationTracking must run on @MainActor because   │
│ 69 +             // AppPreferencesStore is @MainActor.                       │
│ 70 +             let stream: AsyncStream<Int> = AsyncStream { continuation   │
│    + in                                                                      │
│ 71 +                 @Sendable func observe() {                              │
│ 72 +                     // Each call to withObservationTracking registers   │
│    + one onChange.                                                           │
│ 73 +                     // We re-register inside onChange to get subsequent │
│    +  changes.                                                               │
│ 74 +                     MainActor.assumeIsolated {                          │
│ 75 +                         withObservationTracking {                       │
│ 76 +                             _ =                                         │
│    + AppPreferencesStore.shared.pollingInterval                              │
│ 77 +                         } onChange: {                                   │
│ 78 +                             Task { @MainActor in                        │
│ 79 +                                                                         │
│    + continuation.yield(AppPreferencesStore.shared.pollingInterval)          │
│ 80 +                             }                                           │
│ 81 +                             observe()                                   │
│ 82 +                         }                                               │
│ 83 +                     }                                                   │
│ 84 +                 }                                                       │
│ 85 +                 // Prime the observation on the main thread.            │
│ 86 +                 Task { @MainActor in observe() }                        │
│ 87               }                                                           │
│ 88 +             var isFirst = true                                          │
│ 89 +             for await newInterval in stream {                           │
│ 90 +                 if isFirst { isFirst = false; continue }                │
│ 91 +                 guard !Task.isCancelled else { return }                 │
│ 92 +                 log("RunnerStore › pollingInterval changed to           │
│    + \(newInterval) — restarting poll loop")                                 │
│ 93 +                 await self?.start()                                     │
│ 94               }                                                           │
│ 95 +         }                                                               │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Observation chain outlives Task cancellation — AsyncStream continuation is   │
│ never finished                                                               │
│                                                                              │
│ When intervalObservationTask?.cancel() is called (e.g., if                   │
│ _startObservingPreferences were ever called a second time), the inner        │
│ observe() chain keeps running indefinitely: withObservationTracking.onChange │
│  re-calls observe() each time pollingInterval changes, each call yields into │
│  a continuation that nobody is reading any more, and there is no             │
│ withTaskCancellationHandler to call continuation.finish() on cancellation.   │
│ The same pattern is present in _startObservingScopes (lines 99–125). In the  │
│ current code both methods are only called once from init, so no leak occurs  │
│ today; but the defensive intervalObservationTask?.cancel() at the top of     │
│ each method would silently fail to stop the old chain if the methods were    │
│ ever invoked more than once.                                                 │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 4 ·  P2  · comment · L18-L23 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Views/Runner/RunnerViewModel.swift ───────────────────────┤
│                                                                              │
│ 18 +     /// The app-wide singleton. `RunnerStore` writes into this          │
│    + instance.                                                               │
│ 19 +     ///                                                                 │
│ 20 +     /// Declared `@MainActor` so the initialiser runs on the main       │
│    + actor, satisfying                                                       │
│ 21 +     /// Swift 6 strict concurrency. `RunnerStore` captures it by        │
│    + reference; all                                                          │
│ 22 +     /// mutations go through `await MainActor.run { }`.                 │
│ 23 +     @MainActor static let shared = RunnerViewModel()                    │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ RunnerViewModel.shared is never written to — documentation is incorrect      │
│                                                                              │
│ The property comment says "RunnerStore writes into this instance," but       │
│ RunnerStore.applyFetchResult writes into self.viewModel (the instance        │
│ injected via init(viewModel:localRunnerStore:)), not RunnerViewModel.shared. │
│  AppDelegate creates its own let observable = RunnerViewModel() and passes   │
│ it to RunnerStore; shared is a separate object that stays permanently empty. │
│  This is compounded by the class-level docstring (lines 10–12), which still  │
│ refers to the removed RunnerStore.didUpdate and LocalRunnerStore.$runners    │
│ Combine sinks. If any future call site reads RunnerViewModel.shared          │
│ expecting live data, it will see all-empty state forever.                    │
└──────────────────────────────────────────────────────────────────────────────┘

--------------------------------------------------------------------------------
Last reviewed commit: e652ad2456c6aa929bac0f7cff84ad5f6fc26b5d - Review ID: 
b67d0463-ff39-40c6-90bf-3ca4774b2c7a
eon@eons-MacBook-Pro %

eonist:
eon@eons-MacBook-Pro % greptile review
Review complete in 8m 18s - 4 comments across 3 files - commit 805e53879ada

╭──────────────────────────────────────────────────────────────────────────────╮
│ Greptile Summary                                                             │
│ This PR migrates the app's observable layer from                             │
│ ObservableObject/@Published/Combine to Swift's @Observable macro and         │
│ converts RunnerStore from a @MainActor singleton class to a plain actor with │
│  injected dependencies (RunnerViewModel, LocalRunnerStore). All              │
│ @EnvironmentObject/@StateObject/@ObservedObject usages across the view       │
│ hierarchy are updated consistently to their @Observable equivalents.         │
│                                                                              │
│ - RunnerStore is now a DI-friendly actor that pushes state to                │
│ RunnerViewModel via await MainActor.run { } after every poll cycle,          │
│ replacing RunnerStore.didUpdate and LocalRunnerStore.$runners Combine sinks. │
│ - ScopeStore, AppPreferencesStore, NotificationPreferences,                  │
│ LocalRunnerStore, PanelVisibilityState, and PanelSheetState are all migrated │
│  to @Observable; direct RunnerStore.shared references in views are replaced  │
│ with injected onRestartPolling callbacks.                                    │
│ - AggregateStatus gains an init(runners:) factory that consolidates status   │
│ derivation previously duplicated between RunnerStore.aggregateStatus and     │
│ AppDelegate+StatusItem.                                                      │
│                                                                              │
│ Confidence: 3/5                                                              │
│ The poll engine and view-model wiring have a broken sync path that will      │
│ silently show stale local-runner data in the panel during live sessions;     │
│ needs a fix before merging.                                                  │
│                                                                              │
│ The removal of the LocalRunnerStore.$runners Combine sink was not replaced   │
│ with any equivalent push: RunnerViewModel.localRunners is only populated at  │
│ openPanel() and never again while the panel stays open.                      │
│ PanelMainView.activeLocalRunners reads store.localRunners and will silently  │
│ show outdated data after any in-session localRunnerStore.refresh(). Several  │
│ issues flagged in prior threads (MainActor.assumeIsolated trapping on the    │
│ actor executor, dead didChangePollingInterval subject, first-interval-change │
│  dropped) remain unresolved, keeping the risk elevated even after this PR's  │
│ own gap is addressed.                                                        │
│                                                                              │
│ Sources/RunnerBar/App/AppDelegate+PanelSetup.swift and                       │
│ Sources/RunnerBar/Views/Main/PanelMainView.swift for the missing             │
│ localRunners sync; Sources/RunnerBar/Runner/RunnerStore.swift for the        │
│ outstanding MainActor.assumeIsolated calls and the NSApp.delegate coupling   │
│ inside applyFetchResult.                                                     │
╰──────────────────────────────────────────────────────────────────────────────╯

Sequence Diagram

  Open diagram ↗: file:///var/folders/2b/2c_967m52k5fhdyrf77_b01h0000gn/T/grept
ile-diagrams-cOcdnO/cdcdfe70b6bece8a.svg

┌ Comment 1 ·  P1  · comment · L176-L180 ──────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/App/AppDelegate+PanelSetup.swift ─────────────────────────┤
│                                                                              │
│ 176           log("AppDelegate › setupCombineSubscriptions — begin")         │
│ 177                                                                          │
│ 178 +         // local runner list changes are now pushed directly from      │
│     + LocalRunnerStore                                                       │
│ 179 +         // into observable.localRunners via await MainActor.run — no   │
│     + Combine sink needed.                                                   │
│ 180                                                                          │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ RunnerViewModel.localRunners is never updated after openPanel()              │
│                                                                              │
│ The LocalRunnerStore.$runners Combine sink that called observable.reload()   │
│ was removed with the comment "now pushed directly via await MainActor.run",  │
│ but no such push exists anywhere in LocalRunnerStore.                        │
│ RunnerViewModel.reload() is only called from AppDelegate.openPanel(), so     │
│ observable.localRunners holds a one-time snapshot from panel open.           │
│                                                                              │
│ PanelMainView calls localRunnerStore.refresh() on appear and reads           │
│ store.localRunners in activeLocalRunners. After the refresh completes,       │
│ localRunnerStore.runners updates (the @Observable property setter fires),    │
│ but store.localRunners remains at the stale snapshot. The "Local Runners"    │
│ row section will not reflect runner additions, removals, or metric changes   │
│ that occur during the session until the panel is closed and reopened.        │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 2 ·  P2  · comment · L14-L17 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/RunnerStore.swift ─────────────────────────────────┤
│                                                                              │
│ 14 + ///   properties must happen inside `await MainActor.run { }` or a      │
│    + `Task { @MainActor in }`.                                               │
│ 15 + /// - After every fetch cycle, results are pushed to                    │
│    + `RunnerViewModel.shared` on the                                         │
│ 16 + ///   main actor via `await MainActor.run { }`. SwiftUI's `@Observable` │
│    +  machinery                                                              │
│ 17 + ///   picks up the mutation automatically — no Combine                  │
│    + `PassthroughSubject` needed.                                            │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Actor docstring says results are pushed to RunnerViewModel.shared; the       │
│ actual target is the injected viewModel                                      │
│                                                                              │
│ applyFetchResult writes into self.viewModel (the instance passed to init).   │
│ AppDelegate creates its own let observable = RunnerViewModel() and injects   │
│ that — RunnerViewModel.shared is a separate static that never receives any   │
│ pushes. The docstring referencing .shared will mislead anyone who reads it   │
│ and could lead a future engineer to read stale state from .shared expecting  │
│ live data.                                                                   │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 3 ·  P2  · comment · L260-L267 ──────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/RunnerStore.swift ─────────────────────────────────┤
│                                                                              │
│ 260 +         await MainActor.run {                                          │
│ 261 +             vm.runners         = snapshotRunners                       │
│ 262 +             vm.jobs            = snapshotJobs                          │
│ 263 +             vm.actions         = snapshotActions                       │
│ 264 +             vm.isRateLimited   = snapshotRateLimited                   │
│ 265 +             vm.rateLimitResetDate = snapshotResetDate                  │
│ 266 +             (NSApp.delegate as? AppDelegate)?.updateStatusIcon()       │
│ 267 +         }                                                              │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ applyFetchResult bypasses DI by reaching back to AppDelegate via             │
│ NSApp.delegate                                                               │
│                                                                              │
│ RunnerStore was refactored specifically to eliminate singleton               │
│ cross-references inside the actor body, and viewModel and localRunnerStore   │
│ are injected for exactly that reason. Yet applyFetchResult calls             │
│ (NSApp.delegate as? AppDelegate)?.updateStatusIcon() which reintroduces an   │
│ implicit dependency on AppDelegate that cannot be injected or stubbed in     │
│ tests. A callback injected at init would keep the actor testable and         │
│ consistent with the DI pattern used for viewModel.                           │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 4 ·  P2  · comment · L119-L128 ──────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Scope/ScopeStore.swift ───────────────────────────────────┤
│                                                                              │
│ 119 +     /// Mutating a nested value-type field (`entries[idx].isEnabled`)  │
│     + does not                                                               │
│ 120 +     /// automatically trigger `didMutate` the way add/remove events    │
│     + do,                                                                    │
│ 121 +     /// so the send is explicit.                                       │
│ 122       func setEnabled(_ id: UUID, _ enabled: Bool) {                     │
│ 123           guard let idx = entries.firstIndex(where: { $0.id == id })     │
│       else { return }                                                        │
│ 124           entries[idx].isEnabled = enabled                               │
│ 125           persist()                                                      │
│ 126           log("ScopeStore › scope \(entries[idx].scope)                  │
│       isEnabled=\(enabled)")                                                 │
│ 127 +         didMutate.send()                                               │
│ 128       }                                                                  │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Scope enable/disable toggle now fires three simultaneous runnerStore.start() │
│  calls                                                                       │
│                                                                              │
│ setEnabled now sends didMutate (new in this PR). Toggling a scope in         │
│ ScopesView therefore triggers: (1) _startObservingScopes via                 │
│ withObservationTracking on activeScopes; (2) AppDelegate+PanelSetup's        │
│ didMutate Combine sink; (3) ScopesView.Toggle.set calling                    │
│ onRestartPolling(). All three queue actor calls to start(), each cancelling  │
│ the previous pollTask. The end state is correct but the redundant view       │
│ callback creates unnecessary churn.                                          │
└──────────────────────────────────────────────────────────────────────────────┘

--------------------------------------------------------------------------------
Last reviewed commit: 805e53879adac147b65c4d15448a11f9236f1265 - Review ID: 
6734ae9c-d82c-4441-9b2e-53127265402c
eon@eons-MacBook-Pro % 
eon@eons-MacBook-Pro %

eonist:
eon@eons-MacBook-Pro % greptile review
Review complete in 13m 51s - 4 comments across 4 files - commit f454c3f65c12

╭──────────────────────────────────────────────────────────────────────────────╮
│ Greptile Summary                                                             │
│ This PR migrates the core data layer from @MainActor                         │
│ ObservableObject/Combine to Swift 6 actor types with @Observable, replacing  │
│ Combine PassthroughSubject sinks with withObservationTracking/AsyncStream    │
│ observers and a push-based model into RunnerViewModel.                       │
│                                                                              │
│ - RunnerStore and LocalRunnerStore are converted to plain actors; state is   │
│ pushed to an injected RunnerViewModel via await MainActor.run { } at the end │
│  of each fetch/refresh cycle, eliminating didUpdate and $runners Combine     │
│ bridges.                                                                     │
│ - All SwiftUI observable types (ScopeStore, AppPreferencesStore,             │
│ NotificationPreferences, PanelSheetState, PanelVisibilityState,              │
│ SystemStatsViewModel) are migrated from ObservableObject/@Published to       │
│ @Observable, with view wrappers updated from                                 │
│ @StateObject/@ObservedObject/@EnvironmentObject to @State/@Environment.      │
│ - RunnerStore and LocalRunnerStore now accept dependencies via designated    │
│ init to eliminate singleton cross-references inside actor bodies.            │
│                                                                              │
│ Confidence: 2/5                                                              │
│ Multiple blocking defects from prior review rounds remain unaddressed; the   │
│ PR introduces crashes on every poll-loop start and leaves the local-runner   │
│ display permanently empty in production.                                     │
│                                                                              │
│ The MainActor.assumeIsolated calls in RunnerStore.start() and                │
│ nextPollInterval() execute on the actor's own background executor and trap   │
│ unconditionally the first time the poll loop starts. LocalRunnerStore.shared │
│  was initialised with RunnerViewModel.shared while AppDelegate injects its   │
│ own separate RunnerViewModel() into views, so observable.localRunners is     │
│ never written and every local-runner display is permanently empty. The first │
│  polling-interval change is silently skipped by the isFirst guard. These     │
│ carry-over defects remain unaddressed and affect primary runtime paths.      │
│                                                                              │
│ Sources/RunnerBar/Runner/RunnerStore.swift (assumeIsolated crash, missing    │
│ deinit), Sources/RunnerBar/App/AppDelegate.swift (observable vs              │
│ RunnerViewModel.shared mismatch),                                            │
│ Sources/RunnerBar/Runner/LocalRunnerStore.swift (internal init)              │
╰──────────────────────────────────────────────────────────────────────────────╯

Sequence Diagram

  Open diagram ↗: file:///var/folders/2b/2c_967m52k5fhdyrf77_b01h0000gn/T/grept
ile-diagrams-KwMplM/eff1e612f63cdb7e.svg

┌ Comment 1 ·  P2  · comment · L108-L114 ──────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/App/AppDelegate.swift ────────────────────────────────────┤
│                                                                              │
│ 108 +     /// Owned `RunnerStore` actor — injected with `observable` at init │
│     +  so no singleton                                                       │
│ 109 +     /// cross-references exist inside the actor body (Swift 6 / PR     │
│     + #1303 requirement).                                                    │
│ 110 +     /// Owned `LocalRunnerStore` — still a `@MainActor` class until    │
│     + Commit 2 converts it to an actor.                                      │
│ 111 +     let localRunnerStore: LocalRunnerStore = .shared                   │
│ 112 +     /// Owned `RunnerStore` actor — injected with `observable` and     │
│     + `localRunnerStore` so no                                               │
│ 113 +     /// singleton cross-references exist inside the actor body (Swift  │
│     + 6 / PR #1303).                                                         │
│ 114 +     lazy var runnerStore: RunnerStore = RunnerStore(viewModel:         │
│     + observable, localRunnerStore: localRunnerStore)                        │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Stale comment: LocalRunnerStore is already an actor                          │
│                                                                              │
│ The doc-comment says "Owned LocalRunnerStore — still a @MainActor class      │
│ until Commit 2 converts it to an actor," but LocalRunnerStore was converted  │
│ to a Swift actor in this very PR. A reader following this comment to         │
│ LocalRunnerStore.swift will find no @MainActor annotation and no             │
│ ObservableObject conformance, which contradicts the text and may mislead     │
│ future engineers about the concurrency model.                                │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 2 ·  P2  · comment · L42-L44 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/LocalRunnerStore.swift ────────────────────────────┤
│                                                                              │
│ 42 +     init(viewModel: RunnerViewModel) {                                  │
│ 43 +         self.viewModel = viewModel                                      │
│ 44           log("LocalRunnerStore › init —                                  │
│      runnerIndex.count=\(index.runnerIndex.count), runners=[] (call          │
│      refresh() to hydrate)")                                                 │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ init(viewModel:) is now internal, not private                                │
│                                                                              │
│ The old LocalRunnerStore had private init() — construction was restricted to │
│  the type itself, enforcing the singleton access pattern. The new            │
│ init(viewModel:) carries no access-control modifier, so it defaults to       │
│ internal and can be called from any file in the RunnerBar module. If the     │
│ intent is to allow injection in tests, the visibility should be explicit. If │
│  the singleton is still intended to be the only production path, restoring   │
│ private prevents accidental construction elsewhere and matches the old       │
│ contract.                                                                    │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 3 ·  P2  · comment · L35-L42 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Runner/RunnerStore.swift ─────────────────────────────────┤
│                                                                              │
│ 35       private(set) var rateLimitResetDate: Date?                          │
│ 36                                                                           │
│ 37       private var pollTask: Task<Void, Never>?                            │
│ 38 +     private var intervalObservationTask: Task<Void, Never>?             │
│ 39 +     private var scopeObservationTask: Task<Void, Never>?                │
│ 40                                                                           │
│ 41 +     /// The view model this store pushes updates into.                  │
│ 42 +     private let viewModel: RunnerViewModel                              │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Missing deinit leaves three long-lived tasks uncancelled                     │
│                                                                              │
│ The old RunnerStore had deinit { pollTask?.cancel() }. The new actor adds    │
│ two more long-running tasks (intervalObservationTask, scopeObservationTask)  │
│ but drops deinit entirely. Swift's Task type is not cancelled when its       │
│ stored reference is released — the tasks continue running until they         │
│ complete or are explicitly cancelled. In production the singleton never      │
│ deallocates, so there is no observable leak today. However, in any unit-test │
│  or preview context that creates a RunnerStore via the new                   │
│ init(viewModel:localRunnerStore:) and then lets it go out of scope, all      │
│ three tasks remain live and observation chains stay registered against       │
│ AppPreferencesStore.shared and ScopeStore.shared indefinitely.               │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 4 ·  P2  · comment · L27-L29 ────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Views/Settings/LocalRunnersView.swift ────────────────────┤
│                                                                              │
│ 27 +     /// The local runner actor. Used only for mutations (refresh,       │
│    + optimistic updates).                                                    │
│ 28 +     /// Observed state (`localRunners`, `isLocalScanning`) is read from │
│    +  `store` instead.                                                       │
│ 29 +     private var localRunnerStore: LocalRunnerStore {                    │
│    + LocalRunnerStore.shared }                                               │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Mutation path is hardcoded to singleton, defeating the DI goal               │
│                                                                              │
│ store: RunnerViewModel is injected for readable state, but localRunnerStore  │
│ for mutations is a computed property that always returns                     │
│ LocalRunnerStore.shared. Every refresh(), optimisticallySetRunning,          │
│ setLifecycleWarning, and optimisticallyRemove call goes unconditionally to   │
│ the singleton — there is no injection seam. Making localRunnerStore a stored │
│  var with a default of .shared (matching the pattern in PanelMainView and    │
│ AddRunnerSheet) would provide the same behaviour in production while         │
│ allowing tests to inject a different instance.                               │
└──────────────────────────────────────────────────────────────────────────────┘

--------------------------------------------------------------------------------
Last reviewed commit: f454c3f65c121eb0131cd47b0fe24635299ed84c - Review ID: 
76f29c3d-51d1-46ca-a434-fb2da4528387
eon@eons-MacBook-Pro %

eonist (commented):
## Consolidated review — re-verification against current head

Re-read the PR intent and verified all prior review points against the current branch. Most earlier blockers are already fixed; repeating them would be noise against code that has moved on. This review covers only what still holds.

---

### ✅ Items confirmed fixed — dropping from review

| Prior concern | Status |
|---|---|
| `MainActor.assumeIsolated` crash in `start()` / `nextPollInterval()` | Fixed — current code uses `await MainActor.run` throughout |
| `NSApp.delegate` singleton access inside `RunnerStore` | Fixed — injected `onStatusUpdate` callback used instead |
| Observation re-registration off-main / `observe()` crash | Fixed — observer helpers are `@MainActor` and re-register inside `Task { @MainActor in }` |
| `isFirst` dropping the first real scope/interval change | Fixed — that logic is gone entirely |
| `LocalRunnerStore.shared` writing into `RunnerViewModel.shared` (split-brain) | Fixed — `configure(viewModel: observable)` wires the correct instance |
| `pushRunners()` fire-and-forget ordering race | Fixed — `pushRunners()` is `async` and directly `await`s `MainActor.run` |
| Missing `deinit` task cancellation | Fixed — `pollTask`, `intervalObservationTask`, `scopeObservationTask` all cancelled |

---

### ⚠️ One remaining architectural asymmetry

**`LocalRunnerStore` wiring is not symmetric with `RunnerStore`.**

`RunnerStore` follows the PR's stated DI model cleanly:
```swift
lazy var runnerStore: RunnerStore = RunnerStore(
    viewModel: observable,
    localRunnerStore: localRunnerStore,
    onStatusUpdate: { [weak self] in self?.updateStatusIcon() }
)
```
Ownership is explicit, dependencies are injected at construction time, and `AppDelegate` holds the instance directly.

`LocalRunnerStore` takes a different path:
```swift
// AppDelegate+PanelSetup.swift
LocalRunnerStore.configure(viewModel: observable)  // out-of-band singleton write

// AppDelegate.swift
lazy var localRunnerStore: LocalRunnerStore = .shared  // reads the global
```
This works correctly *if and only if* `configure(viewModel:)` is called before `localRunnerStore` is first accessed, which the current startup sequence in `setupCombineSubscriptions()` enforces. But it is an ordering contract with no compile-time enforcement — a future refactor of `applicationDidFinishLaunching` that moves or removes the `configure` call would silently break the wiring until the `fatalError` fires at runtime.

`RunnerStore` cannot fail this way because there is no ordering contract to break.

**Suggested direction (pick one):**
- **Keep as-is for this phase** and add a comment on `lazy var localRunnerStore` explicitly documenting that `configure(viewModel:)` in `setupCombineSubscriptions()` is a precondition for this accessor and must never be moved after it. The `fatalError` in `.shared` already gives a good runtime diagnostic; the gap is just the missing call-site doc.
- **Make ownership fully symmetric** by constructing `LocalRunnerStore` directly in `AppDelegate` the same way `RunnerStore` is constructed — `lazy var localRunnerStore = LocalRunnerStore(viewModel: observable)` — and removing the `configure`/`_shared` path entirely. This is the cleaner end state for Phase 2's stated DI goal, but it requires verifying no other call site outside `AppDelegate` accesses `LocalRunnerStore.shared`.

Neither option is a functional blocker in the current head. The ordering is respected and the `fatalError` guard catches regressions. But the asymmetry is the one place where the implementation still falls short of the "AppDelegate owns and wires both stores" story the PR describes.

---

**Net:** This is solid Phase 2 work. The actor isolation, Combine removal, and push model are all correct. The only open item is the `LocalRunnerStore` wiring asymmetry above — worth a comment or a follow-up issue if not addressed here.

eonist:
eon@eons-MacBook-Pro % greptile review
Review complete in 10m 9s - 3 comments across 3 files - commit cec12121d493

╭──────────────────────────────────────────────────────────────────────────────╮
│ Greptile Summary                                                             │
│ This PR continues the Swift 6 actor migration by converting RunnerStore and  │
│ LocalRunnerStore from @MainActor observable classes to proper Swift actors   │
│ with constructor injection, replacing singleton cross-references with        │
│ injected dependencies (viewModel, localRunnerStore, onStatusUpdate). Views   │
│ are updated from @ObservedObject/ObservableObject to @State/@Observable, and │
│  withObservationTracking/AsyncStream bridges replace the Combine sinks that  │
│ previously wired store changes to the poll loop.                             │
│                                                                              │
│ - RunnerStore becomes a Swift actor; PreferencesObserver/ScopesObserver      │
│ helper classes retain themselves in the Task's async scope to fix the        │
│ immediate-deallocation bug; deinit cancels all three long-lived tasks.       │
│ - LocalRunnerStore becomes a Swift actor configured at startup via           │
│ configure(viewModel:); applyRefreshResults uses await MainActor.run directly │
│  to prevent the isLocalScanning race; all optimistic-mutation helpers are    │
│ async and push snapshots in actor-serialised order.                          │
│ - RunnerViewModel.shared is replaced with a fatalError accessor; the live    │
│ instance is exclusively AppDelegate.observable, injected into both stores    │
│ and forwarded through the view hierarchy.                                    │
│                                                                              │
│ Confidence: 3/5                                                              │
│ The concurrency architecture is substantially improved, but two data-flow    │
│ bugs remain: the import flow races refresh() ahead of add(), and the         │
│ AddRunnerSheet.store default is a silent crash trap.                         │
│                                                                              │
│ The register-flow fix in AddRunnerSheet.swift correctly awaits add(), but    │
│ the same race was not applied to importExistingRunner() in                   │
│ AddRunnerSheet+FormFields.swift. Additionally, var store: RunnerViewModel =  │
│ .shared uses an unconditional fatalError as its default, so any preview,     │
│ test, or future call site that omits the argument crashes immediately at     │
│ struct init time.                                                            │
│                                                                              │
│ Sources/RunnerBar/Views/Settings/AddRunnerSheet+FormFields.swift             │
│ (import-flow race) and Sources/RunnerBar/Views/Settings/AddRunnerSheet.swift │
│  (fatalError default value)                                                  │
╰──────────────────────────────────────────────────────────────────────────────╯

Sequence Diagram

  Open diagram ↗: file:///var/folders/2b/2c_967m52k5fhdyrf77_b01h0000gn/T/grept
ile-diagrams-GMsJzC/95fe9a25d7dfc53a.svg

┌ Comment 1 ·  P1  · comment · L367-L371 ──────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Views/Settings/AddRunnerSheet+FormFields.swift ───────────┤
│                                                                              │
│ 367 +         Task {                                                         │
│ 368 +             await localRunnerStore.add(runnerName: detectedName,       │
│     + installPath: existingDir)                                              │
│ 369 +         }                                                              │
│ 370           isPresented = false                                            │
│ 371           onComplete()                                                   │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Import flow still has the fire-and-forget add() race                         │
│                                                                              │
│ The register flow in AddRunnerSheet.swift was fixed by awaiting              │
│ localRunnerStore.add() directly, but importExistingRunner() still wraps the  │
│ call in a fire-and-forget Task {} before calling isPresented = false and     │
│ onComplete(). Because onComplete() enqueues localRunnerStore.refresh()       │
│ synchronously after the Task is created (but before it runs), the refresh    │
│ scan fires first and will not see the new runner. The imported row will      │
│ appear only after the next manual refresh — the same race that was           │
│ explicitly called out and fixed in the registration path.                    │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 2 ·  P1  · comment · L63 ────────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Views/Settings/AddRunnerSheet.swift ──────────────────────┤
│                                                                              │
│ 63 +     var store: RunnerViewModel = .shared                                │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ Default value is an unconditional fatalError crash trap                      │
│                                                                              │
│ RunnerViewModel.shared now calls fatalError unconditionally — unlike         │
│ LocalRunnerStore.shared, which only crashes before configure() is called,    │
│ this one crashes regardless of launch phase. Any call to AddRunnerSheet.init │
│  that omits store: (e.g., a SwiftUI #Preview, a unit test, or a future call  │
│ site) will crash immediately at struct initialisation. The current           │
│ production call site in LocalRunnersView.addRunnerSheet() correctly passes   │
│ store:, so there is no live crash today, but the default is a                │
│ harder-to-notice trap than the equivalent localRunnerStore default on the    │
│ same struct.                                                                 │
└──────────────────────────────────────────────────────────────────────────────┘

┌ Comment 3 ·  P2  · comment · L286 ───────────────────────────────────────────┐
│                                                                              │
├─ Sources/RunnerBar/Views/Settings/LocalRunnersView.swift ────────────────────┤
│                                                                              │
│ 286 +         Task { await                                                   │
│     + localRunnerStore.optimisticallyRemove(runner.runnerName) }             │
│                                                                              │
├──────────────────────────────────────────────────────────────────────────────┤
│ performRemoval still fire-and-forgets its optimistic update — inconsistent   │
│ with performResume/performStop                                               │
│                                                                              │
│ performResume and performStop were updated in this PR to await               │
│ localRunnerStore.optimisticallySetRunning(...) inline so the row visually    │
│ updates on the next main-actor frame. performRemoval still wraps             │
│ optimisticallyRemove in a separate Task {}, so the UI row lingers until the  │
│ actor hop completes — a latency regression compared to the other two         │
│ actions. Applying the same inline-await pattern used in                      │
│ performResume/performStop would keep all three lifecycle actions consistent. │
└──────────────────────────────────────────────────────────────────────────────┘

--------------------------------------------------------------------------------
Last reviewed commit: cec12121d4936c30a52f3e1a4f5a8b650427d817 - Review ID: 
8794982d-d8cd-41e1-b1e2-a6bb6b0de2fc
eon@eons-MacBook-Pro %

PR #1329, opened 2026-06-11
feat: inject RunnerStatusEnricher into LocalRunnerStore + Codable migration (#1326) (6b)

eonist:
## Summary

Fixes #1326 (sub-issue of #1324, Phase 6b epic).

> ⚠️ **Stacked on #1328** — branch cut from `feature/1325-inject-preferences-scope-store`. Re-target base to `main` once #1328 merges.

Injects `RunnerStatusEnricher` into `LocalRunnerStore` via a protocol, replacing the direct `.shared` singleton call inside `performRefresh()`. Also migrates `RunnerStatusEnricher` from `JSONSerialization` + `[String: Any]` to a fully `Codable`-based approach, completing the #1287 guiding principle: _"No `[String: Any]` after Phase 3"_.

### New protocol (`RunnerStatusEnricherProtocol`)

```swift
public protocol RunnerStatusEnricherProtocol: Sendable {
    func enrich(runners: [Runner]) async -> [Runner]
}
extension RunnerStatusEnricher: RunnerStatusEnricherProtocol {}
```

### `LocalRunnerStore` — inject enricher

- Add `enricher: any RunnerStatusEnricherProtocol` parameter to `init`
- Replace `RunnerStatusEnricher.shared.enrich(runners:)` call in `performRefresh()` with `self.enricher.enrich(runners:)`
- Pass `RunnerStatusEnricher.shared` at the production call site

### Codable migration in `RunnerStatusEnricher`

- Introduce typed `RunnerPayload: Codable` struct for the GitHub API runner response
- Replace `init?(dict: [String: Any])` with a `Codable`-based initialiser
- Replace `JSONSerialization.jsonObject` calls with `JSONDecoder`
- Zero `[String: Any]` remaining after migration

## Testing

```swift
struct StubEnricher: RunnerStatusEnricherProtocol {
    func enrich(runners: [Runner]) async -> [Runner] { runners }
}
let store = LocalRunnerStore(viewModel: vm, enricher: StubEnricher())
```

## Checklist

- [ ] `RunnerStatusEnricherProtocol` defined; `RunnerStatusEnricher` conforms
- [ ] `RunnerStatusEnricher` injected into `LocalRunnerStore` via protocol
- [ ] Unit tests for `LocalRunnerStore.performRefresh()` using a stub enricher
- [ ] Zero `[String: Any]` in `RunnerStatusEnricher` — fully `Codable`-based
- [ ] All CI checks pass (SwiftLint, swift test, Periphery, SonarCloud)

[...]

PR #1330, opened 2026-06-11
feat: [Phase 6c] Remove remaining `import Combine` from app layer (#1327)

### LINKED ISSUE #1327: 💉 [Phase 6c] Remove remaining `import Combine` from app layer
Part of #1324.

## Context

The view-model layer is fully Combine-free after Phase 1 (#1287). These app-layer files are the last visible Combine footprint in the codebase.

## Files

- `Sources/RunnerBar/App/AppDelegate.swift`
- `Sources/RunnerBar/App/AppDelegate+Polling.swift`
- `Sources/RunnerBar/App/AppDelegate+PanelSetup.swift`
- `Sources/RunnerBar/GitHub/OAuthService.swift`
- `Sources/RunnerBar/Views/Settings/SettingsView.swift`

## Fix

- Audit what each `import Combine` is actually used for per file
- Replace `PassthroughSubject` / `AnyCancellable` / sink patterns with `AsyncStream`, `withObservationTracking`, or `Task`-based equivalents — matching the pattern already used in `RunnerStore`
- `OAuthService` may need an `AsyncStream`-based callback for the OAuth redirect flow
- Goal: zero `import Combine` across the app target

## Acceptance criteria

- [ ] Zero `import Combine` across the entire app target
- [ ] All replaced patterns use `AsyncStream`, `withObservationTracking`, or structured `Task`
- [ ] All CI checks pass (SwiftLint, swift test, Periphery, SonarCloud)

Ref: #1287

eonist:
## Summary

Fixes #1327 (sub-issue of #1324, Phase 6c epic).

> ⚠️ **Stacked on #1329** — branch cut from `feature/1326-inject-runner-status-enricher`. Re-target base to `main` once #1329 merges.

Removes all remaining `import Combine` from the app target, replacing `PassthroughSubject` / `AnyCancellable` / sink patterns with `AsyncStream`, `withObservationTracking`, or `Task`-based equivalents — matching the pattern already used in `RunnerStore`.

## Files to update

- `Sources/RunnerBar/App/AppDelegate.swift`
- `Sources/RunnerBar/App/AppDelegate+Polling.swift`
- `Sources/RunnerBar/App/AppDelegate+PanelSetup.swift`
- `Sources/RunnerBar/GitHub/OAuthService.swift`
- `Sources/RunnerBar/Views/Settings/SettingsView.swift`

## Changes

- Audit each `import Combine` usage per file
- Replace `PassthroughSubject` / `AnyCancellable` / sink with `AsyncStream`, `withObservationTracking`, or structured `Task`
- `OAuthService` OAuth redirect flow migrated to `AsyncStream`-based callback
- Zero `import Combine` across the entire app target

## Acceptance criteria

- [ ] Zero `import Combine` across the entire app target
- [ ] All replaced patterns use `AsyncStream`, `withObservationTracking`, or structured `Task`
- [ ] All CI checks pass (SwiftLint, swift test, Periphery, SonarCloud)

Ref: #1327 #1324 #1287

___

## **CodeAnt-AI Description**
Note that this phase also includes the earlier runner-status refactor work

### What Changed
- Added a note explaining that this branch covers the app-layer Combine removal work
- Clarified that the RunnerStatusEnricher protocol and Codable migration changes from the previous PR are included as a prerequisite

### Impact
`✅ Clearer PR context`
`✅ Easier review of stacked changes`
`✅ Less confusion about branch scope`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

### DIFF
--- a/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
+++ b/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
@@ -1,12 +1,11 @@
 // AppDelegate+PanelSetup.swift
 // RunnerBar
 import AppKit
-import Combine
 import SwiftUI
 
 // MARK: - AppDelegate + Panel Setup
 //
-// Owns NSPopover construction, KVO on preferredContentSize, and Combine
+// Owns NSPopover construction, KVO on preferredContentSize, and
 // subscriptions that drive icon/store updates.
 // Called once from applicationDidFinishLaunching via setupPanel().
 //
@@ -78,13 +77,13 @@ import SwiftUI
 // the popover in-place — the arrow stays pinned to the original
 // positioningRect. ❌ NEVER call popover.show() again on resize.
 
-/// Extension responsible for NSPopover construction, KVO, and Combine subscriptions.
+/// Extension responsible for NSPopover construction, KVO, and async subscriptions.
 extension AppDelegate: NSPopoverDelegate {
 
     // MARK: Popover construction
 
     /// Builds the NSPopover, embeds the SwiftUI hosting controller, wires KVO
-    /// and Combine subscriptions.
+    /// and async subscriptions.
     func setupPanel() {
         log("AppDelegate › setupPanel — begin")
         let controller = NSHostingController(rootView: mainView())
@@ -102,10 +101,10 @@ extension AppDelegate: NSPopoverDelegate {
         newPopover.delegate = self
 
         popover = newPopover
-        log("AppDelegate › setupPanel — popover created, wiring KVO + Combine")
+        log("AppDelegate › setupPanel — popover created, wiring KVO + subscriptions")
 
         setupKVO(controller: controller)
-        setupCombineSubscriptions()
+        setupSubscriptions()
         log("AppDelegate › setupPanel — complete")
     }
 
@@ -169,18 +168,18 @@ extension AppDelegate: NSPopoverDelegate {
         }
     }
 
-    // MARK: Combine subscriptions
+    // MARK: Async subscriptions
 
-    /// Starts all Combine subscriptions.
-    private func setupCombineSubscriptions() {
-        log("AppDelegate › setupCombineSubscriptions — begin")
+    /// Wires all long-lived async subscriptions (sign-out listener, startup sequence).
+    private func setupSubscriptions() {
+        log("AppDelegate › setupSubscriptions — begin")
 
         // local runner list changes are now pushed directly from LocalRunnerStore
         // into observable.localRunners via await MainActor.run — no Combine sink needed.
 
         // Everything below makes live network calls — skip entirely in UI tests.
         guard ProcessInfo.processInfo.environment["UI_TESTING"] == nil else {
-            log("AppDelegate › setupCombineSubscriptions — UI_TESTING detected, skipping network setup")
+            log("AppDelegate › setupSubscriptions — UI_TESTING detected, skipping network setup")
             return
         }
 
@@ -197,7 +196,7 @@ extension AppDelegate: NSPopoverDelegate {
         //    access to `localRunnerStore` (inside the Task) must find the instance
         //    already configured, or it fatalErrors.
         LocalRunnerStore.configure(viewModel: observable)
-        log("AppDelegate › setupCombineSubscriptions — LocalRunnerStore.configure(viewModel:) called")
+        log("AppDelegate › setupSubscriptions — LocalRunnerStore.configure(viewModel:) called")
 
         // NOTE: The `RunnerStore.didUpdate` Combine sink has been removed.
         // `RunnerStore` is now a Swift actor that pushes state directly to
@@ -219,7 +218,7 @@ extension AppDelegate: NSPopoverDelegate {
             scopeStore: ScopeStore.shared,
             onStatusUpdate: { [weak self] in self?.updateStatusIcon() }
         )
-        log("AppDelegate › setupCombineSubscriptions — RunnerStore created with injected stores")
+        log("AppDelegate › setupSubscriptions — RunnerStore created with injected stores")
 
         // RunnerStore.init no longer accepts @MainActor-isolated default values
         // (Swift 6: default values for parameters must not be @MainActor-isolated
@@ -258,7 +257,7 @@ extension AppDelegate: NSPopoverDelegate {
         // refreshAsync() suspends until disk hydration + launchctl + GitHub enrichment
         // completes, then start() fires. Cycle 1 always has a populated installPathMap
         // so runner rows appear with CPU/MEM already set.
-        log("AppDelegate › setupCombineSubscriptions — scheduling async startup sequence")
+        log("AppDelegate › setupSubscriptions — scheduling async startup sequence")
         Task { [weak self] in
             guard let self else { return }
             log("AppDelegate › startup — awaiting localRunnerStore.refreshAsync()")
@@ -272,6 +271,6 @@ extension AppDelegate: NSPopoverDelegate {
         // the correct repos from the beginning. RunnerStore observes
         // ScopeStore.activeScopes internally via withObservationTracking/AsyncStream,
         // so no Combine sink is needed here — the actor's own observer handles it.
-        log("AppDelegate › setupCombineSubscriptions — complete")
+        log("AppDelegate › setupSubscriptions — complete")
     }
 }
--- a/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
+++ b/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
@@ -199,9 +199,19 @@ extension AppDelegate: NSPopoverDelegate {
         LocalRunnerStore.configure(viewModel: observable)
         log("AppDelegate › setupCombineSubscriptions — LocalRunnerStore.configure(viewModel:) called")
 
-        // RunnerStore.init no longer accepts @MainActor-isolated default values,
-        // so AppPreferencesStore.shared and ScopeStore.shared are passed explicitly
-        // here, where we are already on the @MainActor.
+        // NOTE: The `RunnerStore.didUpdate` Combine sink has been removed.
+        // `RunnerStore` is now a Swift actor that pushes state directly to
+        // the injected `viewModel` (AppDelegate.observable) via `await MainActor.run { }`
+        // at the end of every fetch cycle, and calls `AppDelegate.updateStatusIcon()` inside
+        // that same `MainActor.run` block — so icon refresh is still driven once
+        // per completed fetch cycle without any Combine subscription.
+        // ℹ️ `RunnerViewModel.shared` is a fatalError accessor — the live instance
+        // is AppDelegate.observable, injected explicitly into both stores.
+
+        // RunnerStore.init no longer accepts @MainActor-isolated default values
+        // (Swift 6: default values for parameters must not be @MainActor-isolated
+        // in a nonisolated context). AppPreferencesStore.shared and ScopeStore.shared
+        // are therefore passed explicitly here, where we are already on the @MainActor.
         runnerStore = RunnerStore(
             viewModel: observable,
             localRunnerStore: localRunnerStore,
--- a/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
+++ b/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
@@ -83,6 +83,8 @@ extension AppDelegate: NSPopoverDelegate {
 
     // MARK: Popover construction
 
+    /// Builds the NSPopover, embeds the SwiftUI hosting controller, wires KVO
+    /// and Combine subscriptions.
     func setupPanel() {
         log("AppDelegate › setupPanel — begin")
         let controller = NSHostingController(rootView: mainView())
@@ -93,6 +95,9 @@ extension AppDelegate: NSPopoverDelegate {
         newPopover.contentViewController = controller
         newPopover.contentSize = NSSize(width: 480, height: 300)
         newPopover.animates = false
+        // .applicationDefined: popoverShouldClose(_:) is consulted on every
+        // is true, keeping the popover alive when user clicks in NSOpenPanel.
+        // Manual NSEvent monitor + NSWorkspace observer handle hide-on-app-switch.
         newPopover.behavior = .applicationDefined
         newPopover.delegate = self
 
@@ -106,6 +111,24 @@ extension AppDelegate: NSPopoverDelegate {
 
     // MARK: NSPopoverDelegate
 
+    /// Always returns `true` — AppKit is never blocked from closing the popover here.
+    ///
+    /// All dismiss control is handled by the manual `outsideClickMonitor` and
+    /// `workspaceObserver` in `openPanel()`. Those monitors guard against
+    /// NSOpenPanel clicks via `hasActiveSheet` (the panel is attached as a sheet
+    /// via `beginSheetModal`, so `popoverWindow.sheets` is non-empty while it
+    /// is open). There is no need to block AppKit here.
+    ///
+    /// `isFilePickerActive` is intentionally NOT used here. Earlier attempts
+    /// (Attempts 4–6, see `docs/graveyard.md`) tried gating this method on a
+    /// boolean flag, but `beginSheetModal` makes that unnecessary: the sheet
+    /// attachment is structural truth visible via `popoverWindow.sheets`, which
+    /// `hasActiveSheet` reads directly. The flag approach was removed in favour
+    /// of that structural check.
+    ///
+    /// See the OUTSIDE-CLICK / APP-SWITCH HIDE comment block above for the full
+    /// mechanism. See `docs/graveyard.md` for the history of approaches that
+    /// tried to gate this method and why they all failed.
     public func popoverShouldClose(_ popover: NSPopover) -> Bool {
         #if DEBUG
         log("AppDelegate › popoverShouldClose — CALLED behavior=\(popover.behavior.rawValue) panelIsOpen=\(panelIsOpen) caller=\(Thread.callStackSymbols[1])")
@@ -114,6 +137,11 @@ extension AppDelegate: NSPopoverDelegate {
         return true
     }
 
+    /// Syncs internal state after the popover closes for any reason.
+    /// Primary purpose: safety net for OS-initiated closes (e.g. user clicks outside).
+    /// When `closePanel()` or `hidePanel()` drives the close, they call
+    /// `tearDownOpenState()` directly — by the time this fires, `panelIsOpen`
+    /// is already `false` and the guard exits immediately.
     public func popoverDidClose(_ notification: Notification) {
         #if DEBUG
         log("AppDelegate › popoverDidClose — panelIsOpen=\(panelIsOpen) behavior=\((NSApp.delegate as? AppDelegate)?.popover?.behavior.rawValue ?? -1) stack=\(Thread.callStackSymbols.prefix(5).joined(separator: "||"))")
@@ -128,6 +156,7 @@ extension AppDelegate: NSPopoverDelegate {
 
     // MARK: KVO
 
+    /// Observes `preferredContentSize` and updates both width and height.
     private func setupKVO(controller: NSHostingController<AnyView>) {
         log("AppDelegate › setupKVO — attaching preferredContentSize observer")
         sizeObservation = controller.observe(
@@ -135,20 +164,38 @@ extension AppDelegate: NSPopoverDelegate {
             options: [.new]
         ) { [weak self] _, change in
             guard let size = change.newValue, size.height > 0 else { return }
+            // KVO can fire on a background thread — hop to main before touching UI.
             DispatchQueue.main.async { [weak self] in self?.resizeAndRepositionPanel() }
         }
     }
 
     // MARK: Combine subscriptions
 
+    /// Starts all Combine subscriptions.
     private func setupCombineSubscriptions() {
         log("AppDelegate › setupCombineSubscriptions — begin")
 
+        // local runner list changes are now pushed directly from LocalRunnerStore
+        // into observable.localRunners via await MainActor.run — no Combine sink needed.
+
+        // Everything below makes live network calls — skip entirely in UI tests.
         guard ProcessInfo.processInfo.environment["UI_TESTING"] == nil else {
             log("AppDelegate › setupCombineSubscriptions — UI_TESTING detected, skipping network setup")
             return
         }
 
+        // Wire LocalRunnerStore.shared to this AppDelegate's RunnerViewModel instance.
+        //
+        // ⚠️ Must be called before the startup Task below (and before any other
+        // LocalRunnerStore.shared access). LocalRunnerStore no longer self-initialises
+        // with RunnerViewModel.shared — that singleton was a different object from
+        // AppDelegate.observable and caused localRunners to push into a view model
+        // that no SwiftUI view observed (permanent empty local-runner list).
+        //
+        // ❌ NEVER move this call inside the Task — AppDelegate.localRunnerStore
+        //    is a computed `lazy var` backed by `LocalRunnerStore.shared`. The first
+        //    access to `localRunnerStore` (inside the Task) must find the instance
+        //    already configured, or it fatalErrors.
         LocalRunnerStore.configure(viewModel: observable)
         log("AppDelegate › setupCombineSubscriptions — LocalRunnerStore.configure(viewModel:) called")
 
@@ -211,6 +258,10 @@ extension AppDelegate: NSPopoverDelegate {
             log("AppDelegate › startup — runnerStore poll loop started")
         }
 
+        // Scope changes (add / remove / enable toggle) restart RunnerStore so it polls
+        // the correct repos from the beginning. RunnerStore observes
+        // ScopeStore.activeScopes internally via withObservationTracking/AsyncStream,
+        // so no Combine sink is needed here — the actor's own observer handles it.
         log("AppDelegate › setupCombineSubscriptions — complete")
     }
 }
--- a/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
+++ b/Sources/RunnerBar/App/AppDelegate+PanelSetup.swift
@@ -83,8 +83,6 @@ extension AppDelegate: NSPopoverDelegate {
 
     // MARK: Popover construction
 
-    /// Builds the NSPopover, embeds the SwiftUI hosting controller, wires KVO
-    /// and Combine subscriptions.
     func setupPanel() {
         log("AppDelegate › setupPanel — begin")
         let controller = NSHostingController(rootView: mainView())
@@ -95,9 +93,6 @@ extension AppDelegate: NSPopoverDelegate {
         newPopover.contentViewController = controller
         newPopover.contentSize = NSSize(width: 480, height: 300)
         newPopover.animates = false
-        // .applicationDefined: popoverShouldClose(_:) is consulted on every
-        // is true, keeping the popover alive when user clicks in NSOpenPanel.
-        // Manual NSEvent monitor + NSWorkspace observer handle hide-on-app-switch.
         newPopover.behavior = .applicationDefined
         newPopover.delegate = self
 
@@ -111,24 +106,6 @@ extension AppDelegate: NSPopoverDelegate {
 
     // MARK: NSPopoverDelegate
 
-    /// Always returns `true` — AppKit is never blocked from closing the popover here.
-    ///
-    /// All dismiss control is handled by the manual `outsideClickMonitor` and
-    /// `workspaceObserver` in `openPanel()`. Those monitors guard against
-    /// NSOpenPanel clicks via `hasActiveSheet` (the panel is attached as a sheet
-    /// via `beginSheetModal`, so `popoverWindow.sheets` is non-empty while it
-    /// is open). There is no need to block AppKit here.
-    ///
-    /// `isFilePickerActive` is intentionally NOT used here. Earlier attempts
-    /// (Attempts 4–6, see `docs/graveyard.md`) tried gating this method on a
-    /// boolean flag, but `beginSheetModal` makes that unnecessary: the sheet
-    /// attachment is structural truth visible via `popoverWindow.sheets`, which
-    /// `hasActiveSheet` reads directly. The flag approach was removed in favour
-    /// of that structural check.
-    ///
-    /// See the OUTSIDE-CLICK / APP-SWITCH HIDE comment block above for the full
-    /// mechanism. See `docs/graveyard.md` for the history of approaches that
-    /// tried to gate this method and why they all failed.
     public func popoverShouldClose(_ popover: NSPopover) -> Bool {
         #if DEBUG
         log("AppDelegate › popoverShouldClose — CALLED behavior=\(popover.behavior.rawValue) panelIsOpen=\(panelIsOpen) caller=\(Thread.callStackSymbols[1])")
@@ -137,11 +114,6 @@ extension AppDelegate: NSPopoverDelegate {
         return true
     }
 
-    /// Syncs internal state after the popover closes for any reason.
-    /// Primary purpose: safety net for OS-initiated closes (e.g. user clicks outside).
-    /// When `closePanel()` or `hidePanel()` drives the close, they call
-    /// `tearDownOpenState()` directly — by the time this fires, `panelIsOpen`
-    /// is already `false` and the guard exits immediately.
     public func popoverDidClose(_ notification: Notification) {
         #if DEBUG
         log("AppDelegate › popoverDidClose — panelIsOpen=\(panelIsOpen) behavior=\((NSApp.delegate as? AppDelegate)?.popover?.behavior.rawValue ?? -1) stack=\(Thread.callStackSymbols.prefix(5).joined(separator: "||"))")
@@ -156,7 +128,6 @@ extension AppDelegate: NSPopoverDelegate {
 
     // MARK: KVO
 
-    /// Observes `preferredContentSize` and updates both width and height.
     private func setupKVO(controller: NSHostingController<AnyView>) {
         log("AppDelegate › setupKVO — attaching preferredContentSize observer")
         sizeObservation = controller.observe(
@@ -164,49 +135,34 @@ extension AppDelegate: NSPopoverDelegate {
             options: [.new]
         ) { [weak self] _, change in
             guard let size = change.newValue, size.height > 0 else { return }
-            // KVO can fire on a background thread — hop to main before touching UI.
             DispatchQueue.main.async { [weak self] in self?.resizeAndRepositionPanel() }
         }
     }
 
     // MARK: Combine subscriptions
 
-    /// Starts all Combine subscriptions.
     private func setupCombineSubscriptions() {
         log("AppDelegate › setupCombineSubscriptions — begin")
 
-        // local runner list changes are now pushed directly from LocalRunnerStore
-        // into observable.localRunners via await MainActor.run — no Combine sink needed.
-
-        // Everything below makes live network calls — skip entirely in UI tests.
         guard ProcessInfo.processInfo.environment["UI_TESTING"] == nil else {
             log("AppDelegate › setupCombineSubscriptions — UI_TESTING detected, skipping network setup")
             return
         }
 
-        // Wire LocalRunnerStore.shared to this AppDelegate's RunnerViewModel instance.
-        //
-        // ⚠️ Must be called before the startup Task below (and before any other
-        // LocalRunnerStore.shared access). LocalRunnerStore no longer self-initialises
-        // with RunnerViewModel.shared — that singleton was a different object from
-        // AppDelegate.observable and caused localRunners to push into a view model
-        // that no SwiftUI view observed (permanent empty local-runner list).
-        //
-        // ❌ NEVER move this call inside the Task — AppDelegate.localRunnerStore
-        //    is a computed `lazy var` backed by `LocalRunnerStore.shared`. The first
-        //    access to `localRunnerStore` (inside the Task) must find the instance
-        //    already configured, or it fatalErrors.
         LocalRunnerStore.configure(viewModel: observable)
         log("AppDelegate › setupCombineSubscriptions — LocalRunnerStore.configure(viewModel:) called")
 
-        // NOTE: The `RunnerStore.didUpdate` Combine sink has been removed.
-        // `RunnerStore` is now a Swift actor that pushes state directly to
-        // the injected `viewModel` (AppDelegate.observable) via `await MainActor.run { }`
-        // at the end of every fetch cycle, and calls `AppDelegate.updateStatusIcon()` inside
-        // that same `MainActor.run` block — so icon refresh is still driven once
-        // per completed fetch cycle without any Combine subscription.
-        // ℹ️ `RunnerViewModel.shared` is a fatalError accessor — the live instance
-        // is AppDelegate.observable, injected explicitly into both stores.
+        // RunnerStore.init no longer accepts @MainActor-isolated default values,
+        // so AppPreferencesStore.shared and ScopeStore.shared are passed explicitly
+        // here, where we are already on the @MainActor.
+        runnerStore = RunnerStore(
+            viewModel: observable,
+            localRunnerStore: localRunnerStore,
+            preferencesStore: AppPreferencesStore.shared,
+            scopeStore: ScopeStore.shared,
+            onStatusUpdate: { [weak self] in self?.updateStatusIcon() }
+        )
+        log("AppDelegate › setupCombineSubscriptions — RunnerStore created with injected stores")
 
         // RunnerStore.init no longer accepts @MainActor-isolated default values
         // (Swift 6: default values for parameters must not be @MainActor-isolated
@@ -255,10 +211,6 @@ extension AppDelegate: NSPopoverDelegate {
             log("AppDelegate › startup — runnerStore poll loop started")
         }
 
-        // Scope changes (add / remove / enable toggle) restart RunnerStore so it polls
-        // the correct repos from the beginning. RunnerStore observes
-        // ScopeStore.activeScopes internally via withObservationTracking/AsyncStream,
-        // so no Combine sink is needed here — the actor's own observer handles it.
         log("AppDelegate › setupCombineSubscriptions — complete")
     }
 }
--- a/Sources/RunnerBar/App/AppDelegate+Polling.swift
+++ b/Sources/RunnerBar/App/AppDelegate+Polling.swift
@@ -13,9 +13,9 @@ extension AppDelegate {
     /// on the very next fetch cycle.
     ///
     /// ## Why this lives here and not in SettingsView
-    /// `SettingsView`'s `signOutCancellable` is stored in `@State` and is
+    /// `SettingsView`'s `signOutTask` is stored in `@State` and is
     /// only alive while Settings is visible. `AppDelegate` is a true singleton
-    /// for the app's lifetime, so this subscription is always active.
+    /// for the app's lifetime, so this task is always active.
     ///
     /// ## What was broken (regression from PR #1138)
     /// Before #1138, polling was driven by `Timer + scheduleTimer()`. After
@@ -25,7 +25,7 @@ extension AppDelegate {
     /// `Task.sleep` — it never calls `start()` again, so the token fallback
     /// only works if `start()` is explicitly invoked after sign-out.
     func setupSignOutSubscription() {
-        Task { [weak self] in
+        signOutTask = Task { [weak self] in
             for await _ in OAuthService.shared.makeSignOutStream() {
                 log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
                 await self?.runnerStore.start()
--- a/Sources/RunnerBar/App/AppDelegate+Polling.swift
+++ b/Sources/RunnerBar/App/AppDelegate+Polling.swift
@@ -1,7 +1,6 @@
 // AppDelegate+Polling.swift
 // RunnerBar
 
-import Combine
 import Foundation
 
 /// AppDelegate extension managing the OAuth sign-out subscription and poll-loop coordination.
@@ -26,12 +25,11 @@ extension AppDelegate {
     /// `Task.sleep` — it never calls `start()` again, so the token fallback
     /// only works if `start()` is explicitly invoked after sign-out.
     func setupSignOutSubscription() {
-        OAuthService.shared.didSignOut
-            .receive(on: DispatchQueue.main)
-            .sink { [weak self] in
+        signOutTask = Task { [weak self] in
+            for await _ in OAuthService.shared.didSignOut {
                 log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
-                Task { [weak self] in await self?.runnerStore.start() }
+                await self?.runnerStore.start()
             }
-            .store(in: &cancellables)
+        }
     }
 }
--- a/Sources/RunnerBar/App/AppDelegate+Polling.swift
+++ b/Sources/RunnerBar/App/AppDelegate+Polling.swift
@@ -26,7 +26,7 @@ extension AppDelegate {
     /// only works if `start()` is explicitly invoked after sign-out.
     func setupSignOutSubscription() {
         signOutTask = Task { [weak self] in
-            for await _ in OAuthService.shared.didSignOut {
+            for await _ in OAuthService.shared.makeSignOutStream() {
                 log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
                 await self?.runnerStore.start()
             }
--- a/Sources/RunnerBar/App/AppDelegate+Polling.swift
+++ b/Sources/RunnerBar/App/AppDelegate+Polling.swift
@@ -25,7 +25,7 @@ extension AppDelegate {
     /// `Task.sleep` — it never calls `start()` again, so the token fallback
     /// only works if `start()` is explicitly invoked after sign-out.
     func setupSignOutSubscription() {
-        signOutTask = Task { [weak self] in
+        Task { [weak self] in
             for await _ in OAuthService.shared.makeSignOutStream() {
                 log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
                 await self?.runnerStore.start()
--- a/Sources/RunnerBar/App/AppDelegate.swift
+++ b/Sources/RunnerBar/App/AppDelegate.swift
@@ -129,6 +129,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     var savedNavState: NavState?
     /// Sheet state that must survive transient popover hides.
     let panelSheetState = PanelSheetState()
+    /// Retained handle for the sign-out observation task started in
+    /// `setupSignOutSubscription()` (AppDelegate+Polling.swift).
+    /// Keeping a strong reference ensures the task is never silently abandoned.
+    var signOutTask: Task<Void, Never>?
     /// Mirrors `popover.isShown`. Kept separately because `NSPopover.isShown` is not
     /// reliable immediately after `performClose` — our flag is the source of truth.
     /// Set to `true` by `openPanel()`, set to `false` by `tearDownOpenState()`.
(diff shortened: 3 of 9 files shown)


eonist:
## Code Review — Phase 6c: Remove `import Combine` from App Layer

Read the full diff, all five changed source files on the branch, and issue #1327. Here's my honest deep take.

---

### ✅ What's correct and well done

**`AppDelegate+Polling.swift` — `setupSignOutSubscription()`**
The Combine-to-`AsyncStream` replacement is clean. The old pattern was:
```swift
OAuthService.shared.didSignOut
    .sink { [weak self] in
        Task { [weak self] in await self?.runnerStore.start() }
    }
    .store(in: &cancellables)
```
The inner `Task { [weak self] in }` inside the `sink` was always slightly awkward — a nested async escape to call an async method from a sync callback. The replacement:
```swift
signOutTask = Task { [weak self] in
    for await _ in OAuthService.shared.didSignOut {
        await self?.runnerStore.start()
    }
}
```
…is strictly better. The `await` is now first-class and the `weak self` capture only needs to exist once. `AppDelegate` is `@MainActor` so the `Task` inherits that actor context — `runnerStore.start()` is called on the correct executor. No issue here.

**`OAuthService.swift` — `AsyncStream` factory**
The `makeSignOutStreamStatic()` factory is a pragmatic solution to the "can't capture `self` before `super.init`" constraint that would occur if you tried to set up the continuation inline using `self`. The implicitly-unwrapped `cont!` is safe here because the closure is called synchronously by `AsyncStream.init` before the factory returns — the `!` will never actually be `nil` at the `return` site. This is an established pattern. The comment explains the rationale. Fine.

**`SettingsView.swift` — `@State private var signOutTask`**
This directly resolves the `FIXME` that existed on `signOutCancellable: AnyCancellable?`. The old concern — that SwiftUI could recreate the view struct and silently drop the subscription — is equally valid for a `Task` stored in `@State`, *but* the lifecycle here is correct: the Task is created in `onAppear` and explicitly cancelled in `onDisappear`. The `onDisappear` handler sets `signOutTask = nil` after cancelling, which is clean. This is the right pattern for a view-scoped async listener.

**`AppDelegate+PanelSetup.swift` — comment / rename only**
The changes are purely mechanical: `setupCombineSubscriptions` → `setupSubscriptions`, log string updates, comment updates. No logic changed. Correct.

---

### ⚠️ Real issues — not false positives

**1. `AsyncStream<Void>` is unicast — second consumer silently gets nothing**

`OAuthService.didSignOut` is a single `AsyncStream`. `AsyncStream` does **not** multicast. Each call to `for await _ in OAuthService.shared.didSignOut` from a second consumer will produce a new iterator, but they all read from the *same* underlying continuation buffer. The continuation's `yield` will only wake **one** waiting consumer (the one currently suspended on `next()`).

Right now there are two simultaneous consumers:
- `AppDelegate.setupSignOutSubscription()` — always alive (app lifetime)
- `SettingsView.onAppearAction()` — alive while Settings is visible

When both are active at the same time (Settings is open) and `signOut()` is called, only one of them will receive the event. Which one wins is non-deterministic / depends on Swift concurrency scheduling. In practice, the `AppDelegate` task is likely to win because it was created first and its iterator is registered first — meaning **`SettingsView` will silently miss the sign-out event** while Settings is open. The UI will not update `isOAuthAuthenticated` to `false`.

The old `PassthroughSubject` was a proper multicast subject — all subscribers received the event. This PR regresses that behaviour.

**Fix options:**
- Use `AsyncChannel` from swift-async-algorithms (proper multicast, FIFO delivery to all consumers)
- Or keep `didSignOut` as a plain `@Published` or `nonisolated(unsafe)` actor-isolated `AsyncStream` and fan it out manually via a `NotificationCenter`-style broadcast
- Simplest correct fix: since `OAuthService` is `@MainActor`, expose a `var didSignOutCount: Int = 0` (or a `Date`-stamped version) as an `@Observable` property, and let consumers use `withObservationTracking` — which is already the established pattern in `RunnerStore`. This would be zero-dependency and fully multicast.

**2. `signOutTask` in `AppDelegate` is never cancelled**

`AppDelegate.signOutTask` is set in `setupSignOutSubscription()` which is called once from `applicationDidFinishLaunching` (via `setupSubscriptions()`). There is no corresponding cancellation anywhere in `AppDelegate` — not in `applicationWillTerminate`, not anywhere.

For an app-lifetime task this is technically acceptable because the process is terminating anyway. However:
- If `OAuthService.shared.didSignOut` is ever finished/terminated (continuation `.finish()` is called), the `for await` loop will exit and the `Task` will complete. There is no path in this PR that calls `finish()`, so this is a latent risk rather than a current bug.
- Consistency: `SettingsView` correctly cancels its `signOutTask` on `onDisappear`. The lack of any cancel path in `AppDelegate` is at minimum worth a comment explaining the intent ("app-lifetime task, never cancelled by design").

**3. `LocalRunnerStore.swift.patch` — stray file committed to source**

A literal `.patch` text file was committed at:
```
Sources/RunnerBar/Runner/LocalRunnerStore.swift.patch
```
This file contains two lines of prose describing the branch scope and should not be in the source tree. It will be compiled into the Swift package target (or at minimum it will confuse Xcode's source navigator and any tooling that globs `Sources/**`). It's clearly a note that was meant to stay local or go in the PR description. **This must be deleted before merge.**

---

### 🔍 Minor / low priority

- The comment in `AppDelegate+Polling.swift` still says `signOutCancellable is stored in @State` in the old docstring (the `## Why this lives here and not in SettingsView` section). This is now stale — it was describing the old `AnyCancellable` behaviour. Worth updating to reference `signOutTask` and the `AsyncStream` unicast limitation note above.
- `OAuthService.signOutContinuation` is `private var` — good — but it's never set to `nil` after `finish()` would be called (there's no `finish()` call in the codebase). If a future change adds app shutdown cleanup, remember to `finish()` and `nil` this out.

---

### Summary

| | |
|---|---|
| `import Combine` actually removed | ✅ Yes, all five files |
| Acceptance criteria #1 (zero `import Combine`) | ✅ Met |
| Acceptance criteria #2 (patterns: `AsyncStream`/`Task`) | ✅ Met mechanically |
| Stray `.patch` file | ❌ Must remove before merge |
| Multicast regression (`AsyncStream` is unicast) | ⚠️ Functional bug when Settings is open during sign-out |
| `signOutTask` cancellation in `AppDelegate` | ⚠️ Latent / acceptable, but needs a comment |

The stray patch file and the unicast multicast regression are the two things that need to be addressed before this is merge-ready.

eonist:
## Deep Review — Phase 6c: Remove `import Combine` from app layer

Reviewed the full diff, all 6 changed files (reading source, not just diff), and the original issue #1327. This is a solid, well-scoped migration. Below are findings ordered by severity — no false positives included (I've verified each against the actual file content).

---

### 🔴 Critical / Blocking

#### 1. `AsyncStream` has a single-consumer contract — two concurrent iterators will race and each silently drop events

`OAuthService.didSignOut` is a single `AsyncStream<Void>` instance created once in `init()` and held for the lifetime of the app.

Two separate places iterate it concurrently at runtime:
- `AppDelegate.setupSignOutSubscription()` — always active for the app's lifetime
- `SettingsView.onAppearAction()` — active while Settings is visible

`AsyncStream` is **not** multicast. Its internal buffer has a single cursor. When both tasks are waiting, `yield(())` will deliver the element to whichever task happens to dequeue it first. The other task never sees it. This is a real race, not a theoretical one — whichever `for await` loop wins the dequeue "consumes" the event. The practical failure modes are:

- Sign-out fires → `AppDelegate` gets it, restarts the poll loop ✅, but `SettingsView` never updates its UI state (`isOAuthAuthenticated` stays `true`, UI shows signed-in while poll loop restarts for env-token) 🐛
- Or vice versa: `SettingsView` gets it, UI updates ✅, but `AppDelegate` misses it, poll loop never restarts 🐛

**Fix options (pick one):**
- Replace `AsyncStream` with a proper multicast mechanism. The idiomatic Swift 6 approach is a `Notification` (still zero Combine), an `@Observable` property on `OAuthService` that both sites observe via `withObservationTracking`, or a light actor that maintains a subscriber list.
- Or: expose a typed observable property `isSignedIn: Bool` on `OAuthService` (already `@MainActor`), and let `SettingsView` use `withObservationTracking` in a `Task` (matching the pattern `RunnerStore` already uses). `AppDelegate` can observe the same property.
- Or: promote `AppDelegate` to be the single consumer and call a delegate/closure into SettingsView — but that's more invasive.

---

### 🟡 Medium

#### 2. `signOutContinuation` can silently become `nil` after the stream's iterator is cancelled

`signOutContinuation` is `private var` (mutable). The `AsyncStream` documentation specifies that when the last iterator is cancelled or goes out of scope, the runtime **calls `onTermination`** on the continuation — and if `onTermination` is not set, it simply invalidates further `yield` calls (they become no-ops). More critically: `signOutContinuation` itself is never nilled out on termination.

In the current code, if the sole iterator in `AppDelegate.setupSignOutSubscription()` is ever cancelled (e.g., on teardown, if `signOutTask` is replaced), the continuation is dead but `signOutContinuation` still holds a non-nil reference. Future `yield(())` calls in `signOut()` will silently no-op — sign-out fires, Keychain is cleared, log says "emitting didSignOut", but nobody hears it. The guard `if deleted { signOutContinuation?.yield(()) }` will not catch this.

Recommended: set `onTermination` in `makeSignOutStreamStatic()`:
```swift
let stream = AsyncStream<Void> { cont = $0 }
// (after multicast fix) or at minimum:
cont.onTermination = { _ in /* log warning or reassign */ }
```

#### 3. Stray `.patch` file committed to `Sources/`

`Sources/RunnerBar/Runner/LocalRunnerStore.swift.patch` was added as a new file. It's a 2-line plaintext note, not an actual patch applied by tooling. It has no functional role, will confuse Xcode's file navigator, and will be compiled (or at least indexed) as part of the Sources target unless explicitly excluded in `Package.swift`.

This should either be removed before merge or placed in a `docs/` or `.notes/` directory outside `Sources/`.

---

### 🟢 Positive / Worth Calling Out

- **`SettingsView` cancellation is correct.** `signOutTask?.cancel()` + `signOutTask = nil` on `onDisappear` is the right lifecycle hook. The old `@State var signOutCancellable: AnyCancellable?` had a real bug (the FIXME you removed was legitimate — SwiftUI can silently drop `@State` storage on struct re-creation). The Task-based replacement avoids that entirely.

- **`AppDelegate.setupSignOutSubscription()` simplification is cleaner.** Removing the nested `Task { [weak self] in await self?.runnerStore.start() }` and directly `await`ing inside the loop body is correct and removes an unnecessary detach. The `[weak self]` on the outer `Task` is appropriate.

- **`OAuthService.makeSignOutStreamStatic()` is the right workaround** for the "use of `self` before initialization" compiler restriction. Force-unwrapping `cont!` is safe here because the `AsyncStream` closure is called synchronously during init before `return` — the Swift runtime guarantee holds.

- **All log strings updated consistently** — no stale `setupCombineSubscriptions` references left in log output. Diff is thorough on this.

- **`cancellables: Set<AnyCancellable>` → `signOutTask: Task<Void, Never>?`** is the correct 1-for-1 replacement given there's now only one long-lived subscription at the AppDelegate level.

---

### Summary

The critical issue (multicast race on a single-consumer `AsyncStream`) needs a fix before this merges. Everything else is clean and the direction is right. The `.patch` file should be cleaned up as a quick follow-up or squashed before merge. Once the multicast problem is addressed, this PR closes #1327 cleanly and the Combine footprint drops to zero as intended.

eonist:
## Code Review — Phase 6c: Remove `import Combine` from app layer

Reviewed diff, full source files on the PR branch, issue #1287 (architecture intent), and the #1281 comments. This is a clean, well-scoped migration. Below is a deep read — I've explicitly excluded anything that is a false positive after verifying the actual source.

---

### ✅ What works well

**`OAuthService.swift` — `PassthroughSubject` → `AsyncStream`**

The factory pattern (`makeSignOutStreamStatic()`) is the right call here. Using a `static` factory avoids the Swift compiler complaint about capturing `self` before it's fully initialised. The `var cont: AsyncStream<Void>.Continuation!` with forced unwrap after `AsyncStream { cont = $0 }` is safe here — the closure executes synchronously before `return`, so `cont` is always set before it is returned. No issue.

`signOutContinuation` is `private var` — good. It's only ever written from `init` and never needs to be replaced, so this is fine. The `yield(())` in `signOut()` is gated on `deleted == true`, preserving the existing semantic that only a successful Keychain removal fires the event. That nuance is correctly carried over.

**`AppDelegate+Polling.swift` — `sink` → `for await`**

The old code nested a `Task` inside a `sink` closure, which was redundant and left the inner task unretained (fire-and-forget). The new version flattens this correctly:

```swift
// Before (Task inside sink — inner task is unretained):
.sink { Task { [weak self] in await self?.runnerStore.start() } }

// After (direct await inside the for-await loop):
for await _ in OAuthService.shared.didSignOut {
    await self?.runnerStore.start()
}
```

The `[weak self]` capture on the outer `Task` is correct — `AppDelegate` is a long-lived singleton but capturing it weakly prevents the task from keeping a dangling strong ref if teardown ever happens.

**`AppDelegate.swift` — `Set<AnyCancellable>` → `Task<Void, Never>?`**

`signOutTask` is a single task for a single subscription, so `Task?` is the right replacement (no bag needed). The property is `internal` which is correct — it needs to be written from `AppDelegate+Polling.swift` across the file boundary. The doc comment correctly describes its purpose and lifecycle.

**`SettingsView.swift` — `@State private var signOutCancellable: AnyCancellable?` → `@State private var signOutTask: Task<Void, Never>?`**

The old FIXME about `AnyCancellable` in `@State` is now resolved cleanly. The new pattern:

- Creates the `Task { @MainActor in … }` inside `onAppearAction()` ✅
- Cancels it in `.onDisappear` ✅
- Sets it to `nil` after cancel to avoid a dangling task reference ✅

The `Task { @MainActor in … }` annotation is correct — `isOAuthAuthenticated` and `isCLIAuthenticated` are `@State` vars on a `View` struct, which must be mutated on the main actor. This replaces the `.receive(on: DispatchQueue.main)` Combine hop and is semantically equivalent.

**`AppDelegate+PanelSetup.swift` — documentation/log cleanup**

All `setupCombineSubscriptions` → `setupSubscriptions` renames are consistent across comments, log strings, and the `MARK:` section header. Comment header updated from "KVO and Combine subscriptions" to "KVO and async subscriptions". Clean.

---

### ⚠️ Real issues worth addressing

#### 1. `LocalRunnerStore.swift.patch` — stray file tracked in the repo

```
Sources/RunnerBar/Runner/LocalRunnerStore.swift.patch
```

This file is a plain-text note (2 lines) about what the branch tracks. It is **not a patch file used by any tooling** — it's just a human note committed as a `.patch` file inside the source tree. This should not be in the repo:

- It has a `.patch` extension, which most editors and CI tools will try to treat as a diff file.
- It is not a Swift source file and will appear in any `find Sources/` sweep.
- It will persist into `main` if merged as-is.

**Action required**: Delete `Sources/RunnerBar/Runner/LocalRunnerStore.swift.patch` before merging. The note it contains belongs in the PR description or a commit message.

---

#### 2. `AsyncStream` multicasting — single-consumer assumption is implicit but undocumented

`AsyncStream` is not a multicast publisher. Each consumer gets its own independent iteration, but they all share the **same continuation** — meaning once an element is `yield()`ed, each active `for await` loop gets it (each consumer gets a copy). That's fine for this use case.

However, there are **two consumers** of `OAuthService.shared.didSignOut` in this diff:

1. `AppDelegate+Polling.swift` — `setupSignOutSubscription()` (app lifetime)
2. `SettingsView.swift` — `onAppearAction()` (view lifetime, can be multiple instances if the view is recreated)

This works correctly *right now* because `AsyncStream` with a single continuation delivers to all active iterators. But this is **not obvious to the next reader**, and the original `PassthroughSubject` had this semantic made explicit via Combine's multicast nature.

**Suggestion**: Add a brief doc comment to `didSignOut` noting it supports multiple concurrent consumers and that each `for await` loop receives every yielded value independently. Example:

```swift
/// Emits after a successful sign-out (Keychain deletion confirmed).
/// Supports multiple concurrent `for await` consumers — each active iterator
/// receives every yielded value. Observe inside a `Task` scoped to the
/// consumer's lifetime; cancel the Task on teardown.
let didSignOut: AsyncStream<Void>
```

This is a documentation gap, not a correctness bug.

---

#### 3. `signOutTask` in `AppDelegate` is never cancelled

`AppDelegate.signOutTask` is set in `setupSignOutSubscription()` (called once from `applicationDidFinishLaunching`) and never cancelled. For a true app-lifetime singleton this is intentional and correct — the task should run for the entire process lifetime.

But there is no `deinit` or teardown call that cancels it. If the task ever ends early (e.g. the `AsyncStream`'s continuation is deallocated), `setupSignOutSubscription()` is never re-called, so the sign-out restart would silently stop working.

The continuation (`OAuthService.shared.signOutContinuation`) is `private var` and lives on the singleton — it will never be deallocated in normal operation. So this is **not a live bug**, but worth a comment:

**Suggestion**: Add a `// NOTE: Never cancelled — intentionally app-lifetime.` comment next to `signOutTask` in `AppDelegate.swift` so the next reader doesn't wonder if this is a leak.

---

### ℹ️ Non-issues (verified, not raised)

- **`[weak self]` in `setupSignOutSubscription` Task** — `self` is `AppDelegate` (a singleton). `weak` is technically redundant since `AppDelegate` outlives the task, but it is correct and adds no overhead. Not a bug, not a leak.
- **`signOutContinuation?.yield(())` optional chain** — `signOutContinuation` is always set in `init()` via the static factory; the optional is never `nil` in practice. The optional chain is defensive but harmless.
- **`@State private var signOutTask`** — Using `@State` for a `Task` is idiomatic in SwiftUI for view-lifetime task management. The previous FIXME acknowledged this was wrong for `AnyCancellable`; it is now correct for `Task` because task cancellation on `onDisappear` is clean and well-defined.
- **`Task { @MainActor in … }` in `onAppearAction()`** — This does not spawn a detached task; it inherits the actor context from the `@MainActor` `onAppearAction()` call site on the `View` struct, so the `@MainActor` annotation on the closure is redundant but harmless.

---

### Summary

| | Status |
|---|---|
| Combine removal correctness | ✅ Complete and correct |
| `AsyncStream` pattern | ✅ Correct for all call sites |
| Task lifecycle (SettingsView) | ✅ Properly cancelled on disappear |
| Task lifecycle (AppDelegate) | ✅ App-lifetime, intentional |
| `LocalRunnerStore.swift.patch` | ❌ Must be deleted before merge |
| Multi-consumer documentation | ⚠️ Worth a doc comment |
| `signOutTask` "never cancelled" note | ⚠️ Worth a one-liner comment |

One hard blocker (the `.patch` file), two minor documentation suggestions. The migration itself is solid — the nested `Task`-in-`sink` anti-pattern is correctly eliminated, the `@MainActor` threading is preserved, and the view lifecycle is cleaner than the Combine version it replaces. Aligns well with the #1287 "no Combine in view-model layer" principle.

eonist:
## Deep Review — Phase 6c: Remove `import Combine` from app layer

Reviewed with full diff read, full file reads for all 5 modified files, the original issue (#1327/#1324), the epic (#1287), and the architecture discussions in #1281.

---

### ✅ What is correct and well done

**`OAuthService` — `AsyncStream` migration is solid**

The `PassthroughSubject<Void, Never>` → `AsyncStream<Void>` replacement is correct.
The static factory `makeSignOutStreamStatic()` is the right pattern here — you cannot refer to `self` before all stored properties are initialised, so a `static func` that builds the `(stream, continuation)` pair before `init` completes is exactly the right solution. The `var cont: AsyncStream<Void>.Continuation!` force-unwrap inside the closure is safe because `AsyncStream`'s initialiser calls the closure synchronously before returning. ✓

`signOutContinuation?.yield(())` in `signOut()` correctly replaces `didSignOut.send()`. Gating on the optional (`?`) rather than force-unwrapping is safe because `signOutContinuation` is set in `init` and `OAuthService` is a singleton — it will never be nil at sign-out time in practice. ✓

**`AppDelegate+Polling.swift` — sign-out subscription**

The `for await _ in OAuthService.shared.didSignOut` loop in `setupSignOutSubscription()` is the correct modern replacement for the Combine sink. The `[weak self]` capture is present. The `await self?.runnerStore.start()` call is now a direct `await` instead of spinning up a nested `Task { }`, which is strictly better — no extra concurrency overhead, no additional dangling Task risk. ✓

Removing the nested `Task { [weak self] in await self?.runnerStore.start() }` in favour of the direct `await` is a real improvement, not just cosmetic. ✓

**`AppDelegate.swift` — `cancellables` → `signOutTask`**

`var cancellables = Set<AnyCancellable>()` → `var signOutTask: Task<Void, Never>?` is correct. The old `cancellables` bag held only a single subscription (the sign-out sink), so a single optional `Task` is the accurate 1-to-1 replacement. ✓

The doc comment is updated correctly: "Task retained for the sign-out listener started in `setupSignOutSubscription()`." ✓

**`SettingsView.swift` — FIXME eliminated**

The old `@State private var signOutCancellable: AnyCancellable?` carried an explicit `FIXME` calling out the well-known problem: `AnyCancellable` stored in `@State` risks silent subscription drop if SwiftUI recreates the view struct and reallocates `@State` storage. That risk is now gone — `Task<Void, Never>` stored in `@State` has the same lifecycle behaviour as `AnyCancellable` in `@State`, but there is no Combine machinery underneath. ✓

The `.onDisappear` cancellation (`signOutTask?.cancel()` + `signOutTask = nil`) is correctly placed and matches the existing teardown pattern for `onCompletion`. ✓

`Task { @MainActor in for await _ in ... }` — the `@MainActor` annotation on the `Task` closure is correct, because `isOAuthAuthenticated` and `isCLIAuthenticated` are `@State` vars on a `View` struct. Those must be mutated from the main actor. ✓

**`AppDelegate+PanelSetup.swift` — purely cosmetic renaming**

All changes are rename-only: `setupCombineSubscriptions` → `setupSubscriptions`, log string updates, comment updates, `import Combine` removal. The underlying logic is unchanged. This is correct — there was already no Combine code left in this file after prior phases. ✓

---

### ⚠️ Real issues — not false flags

**1. `AsyncStream` single-consumer contract — `didSignOut` has two consumers**

`AsyncStream` is a **single-consumer** type. Each call to `next()` on the underlying `AsyncIterator` consumes the element. With the current code, **both** `AppDelegate.setupSignOutSubscription()` and `SettingsView.onAppearAction()` open a `for await _ in OAuthService.shared.didSignOut` loop against the *same* `AsyncStream` instance.

Whichever consumer calls `makeAsyncIterator()` second will race for elements. In practice only one of the two will receive a given `yield(())` — the other will miss the event entirely. If `SettingsView` is open at sign-out time, the AppDelegate listener may miss the event and fail to restart the poll loop. If `SettingsView` is not open, it does not matter — but the contract is not safe.

The correct fix for multi-consumer broadcast is either:
- `AsyncStream` with a manual broadcast wrapper (fan-out to multiple continuations), or
- A `Notifications`-based approach using `NotificationCenter.default.notifications(named:)`, or
- An `actor`-owned sequence that allows multiple listeners (e.g. your own `AsyncBroadcastStream` wrapper)

The simplest correct fix that matches the rest of the codebase is an actor-backed `[AsyncStream<Void>.Continuation]` array that `yield`s to all registered consumers on sign-out. Alternatively, if `SettingsView`'s auth-state update on sign-out is not critical (it also reads `Keychain.token` on `onAppear`), the `SettingsView` consumer can be removed and state can be refreshed on `onAppear` alone — but the current code assumes the stream delivers to both.

**2. `AppDelegate.signOutTask` is never cancelled**

`SettingsView.signOutTask` is correctly cancelled in `.onDisappear`. `AppDelegate.signOutTask` is intentionally long-lived (that is its purpose — it exists for the app lifetime). But there is no `deinit` or `applicationWillTerminate` teardown that cancels it. For a true singleton that lives until process exit this is fine in practice, but it is worth a comment explaining the intentional absence of cancellation, especially given that `SettingsView.signOutTask` *does* cancel on disappear — a reviewer reading both will reasonably ask why one cancels and the other does not. Just a doc comment addition: `/// Never cancelled — lives for the app lifetime.` or similar.

**3. `LocalRunnerStore.swift.patch` file should not be in the source tree**

`Sources/RunnerBar/Runner/LocalRunnerStore.swift.patch` is a plain-text note file added as a `.patch` file extension inside the Swift source tree. It is not a valid patch file (no diff headers, no hunk markers — it is two lines of free text). It will not be compiled by Swift (the compiler ignores `.patch` files), but it will:
- Show up in Xcode's file navigator and potentially confuse contributors
- Pollute the `Sources/` tree with non-source artefacts
- Potentially cause issues if any tooling performs glob-based source enumeration

This should be removed before merge. If the intent is to document why this branch includes prerequisites from #1329, that belongs in the PR description (which it already does) or in a `// NOTE:` comment inside the relevant Swift file, not as a stray `.patch` file in `Sources/`.

---

### 🔍 Things checked and confirmed not issues

- `didSignOut` is `let` (not `var`) on `OAuthService` — correct, the stream instance is fixed after `init`. ✓
- `signOutContinuation` is `private var` — correct, only `signOut()` should yield to it. ✓
- The `@MainActor` on `OAuthService` means `signOut()` and `yield(())` are always called on the main actor, and both `for await` consumers are also on the main actor (AppDelegate is `@MainActor`, SettingsView task uses `@MainActor`). So there is no data race on the continuation itself — the single-consumer issue above is a logical race on elements, not a thread-safety issue. ✓
- `AppDelegate+PanelSetup.swift` no longer calls `setupCombineSubscriptions()` — the rename to `setupSubscriptions()` is reflected consistently in all call sites. ✓
- The `receive(on: DispatchQueue.main)` in the old Combine code was doing thread-hopping. The new `Task { @MainActor in }` replaces it correctly — no missed main-thread requirement. ✓

---

### Summary

| Item | Status |
|---|---|
| `AsyncStream` factory pattern in `OAuthService.init` | ✅ Correct |
| Nested `Task` eliminated in `Polling.swift` | ✅ Improvement |
| `@MainActor` on SettingsView Task closure | ✅ Correct |
| `.onDisappear` Task cancellation in SettingsView | ✅ Correct |
| `AsyncStream` single-consumer violation (two live iterators) | ⚠️ Real bug — needs fix before merge |
| `AppDelegate.signOutTask` never cancelled (missing comment) | ⚠️ Minor — add clarifying doc comment |
| `LocalRunnerStore.swift.patch` stray file | ⚠️ Should be deleted before merge |

The core direction is exactly right and aligns with the goal in #1287 (no Combine in app layer, Swift 6.2 structured concurrency). The single-consumer `AsyncStream` issue is the one thing that needs to be resolved — everything else is either correct or minor polish.

eonist:
## Follow-up review — commit 16420e9

All three blockers/suggestions from the previous review are addressed. Here's where things stand now.

---

### ✅ Hard blocker resolved
**`LocalRunnerStore.swift.patch` is deleted.** Nothing more to do.

---

### ✅ Multicast regression fixed — and the fix is correct

The commit message calls this a P1, which is right. The original PR had a real latent bug: `AsyncStream` is single-consumer per iterator, so with both `AppDelegate` and `SettingsView` independently calling `for await _ in OAuthService.shared.didSignOut`, each `yield()` would wake *one* of the two waiters non-deterministically — the other would silently miss the event. The old `PassthroughSubject.send()` was genuinely multicast; this was a semantic regression.

The fix — `makeSignOutStream()` returning a dedicated stream per caller and broadcasting via `signOutContinuations.forEach { $0.yield(()) }` — is the correct pattern. Using `AsyncStream.makeStream()` (the safe tuple API) instead of the old `var cont: AsyncStream<Void>.Continuation!` force-unwrap also resolves the P2 from the previous review.

---

### One new concern worth checking: continuation leak

`signOutContinuations` is an append-only array. Continuations are added in `makeSignOutStream()` but **never removed**. Consider what happens with `SettingsView`:

- `onAppear` calls `makeSignOutStream()` → appends a continuation
- `onDisappear` cancels `signOutTask` → the `for await` loop exits
- But the continuation stays in `signOutContinuations` forever

If Settings is opened and closed multiple times (which is normal — user opens settings, closes, reopens), continuations accumulate. Each subsequent `signOut()` call yields to an ever-growing list of dead continuations.

This is **not a crash** — yielding to a finished/cancelled stream's continuation is a no-op in Swift. But it is a memory growth pattern and a correctness smell.

**How to fix**: have `makeSignOutStream()` wire the continuation's `onTermination` handler to self-remove from the array:

```swift
func makeSignOutStream() -> AsyncStream<Void> {
    let (stream, cont) = AsyncStream<Void>.makeStream()
    signOutContinuations.append(cont)
    let id = ObjectIdentifier(cont as AnyObject) // or capture index via a UUID key
    cont.onTermination = { [weak self] _ in
        // Must be on MainActor since OAuthService is @MainActor
        Task { @MainActor [weak self] in
            self?.signOutContinuations.removeAll { ObjectIdentifier($0 as AnyObject) == id }
        }
    }
    return stream
}
```

Alternatively, store continuations in a `[UUID: AsyncStream<Void>.Continuation]` dictionary keyed by a UUID returned alongside the stream, and remove by key on termination. Either approach fixes the leak cleanly.

For the `AppDelegate` consumer this is moot (app lifetime, one allocation). For `SettingsView` it matters in practice.

---

### ✅ Non-issue: `signOutTask` never-cancelled comment

The previous suggestion to add a `// NOTE: Never cancelled — intentionally app-lifetime.` comment on `AppDelegate.signOutTask` is still a nice-to-have but even less pressing now that the multicast pattern is explicit in `OAuthService` itself.

---

### Summary

| | Before 16420e9 | After 16420e9 |
|---|---|---|
| `.patch` file in source tree | ❌ Present | ✅ Removed |
| Multicast regression | ❌ Silent event drop | ✅ Fixed |
| Force-unwrap in factory | ❌ `Continuation!` | ✅ `makeStream()` tuple |
| Continuation leak in `signOutContinuations` | n/a | ⚠️ New, append-only array |

One new issue introduced by the fix (continuation accumulation). Worth addressing before merge — it's a small addition to `makeSignOutStream()` and doesn't require touching any call sites.

eonist:
## Code Review — PR #1330: Remove Combine from sign-out path

Reviewed against: issue #1287 (Swift 6.2 modernisation goal), and the discussion in #1281.

This is a focused, well-scoped PR. The core trade — `PassthroughSubject` + `AnyCancellable` → `AsyncStream` multicast — is the right call and lands cleanly. Below is a detailed look at what works, what's correct, and one genuine concern worth discussing.

---

### ✅ What's right

**`OAuthService.makeSignOutStream()` / multicast design**
The comment block is accurate and worth keeping. `AsyncStream` is deliberately single-consumer, so the array-of-continuations approach is exactly how you restore the one-to-many delivery that `PassthroughSubject` gave you without re-introducing Combine. Each caller gets its own stream, `signOut()` yields to all registered continuations — correct.

**`SettingsView` — Task lifecycle**
Replacing `@State private var signOutCancellable: AnyCancellable?` with `@State private var signOutTask: Task<Void, Never>?` and cancelling in `.onDisappear` is the right pattern. The old FIXME comment about `@State` + `AnyCancellable` being fragile (silent subscription drop on view recreation) is now gone because the underlying hazard is gone. The `Task { @MainActor in … }` annotation ensures `isOAuthAuthenticated` and `isCLIAuthenticated` state writes stay on the main actor. Clean.

**`AppDelegate+Polling.swift` — no nested Task**
Old code: `.sink { Task { [weak self] in await self?.runnerStore.start() } }` — a Task spawned inside a Combine sink closure, meaning it wasn't tracked anywhere.
New code: `Task { [weak self] in for await _ in … { await self?.runnerStore.start() } }` — one flat async loop. The removal of the nested Task is a real improvement to readability and correctness.

**Comment quality**
The in-file comments (`// AsyncStream is single-consumer — sharing one stream … would deliver each event to only one consumer`) are exactly the kind of reasoning future contributors need. Good.

---

### 🔴 Genuine concern: Task retention in `AppDelegate`

In `AppDelegate.swift` the new doc-comment reads:

```swift
/// Task retained for the sign-out listener started in `setupSignOutSubscription()`.
```

But there is **no stored property** backing this comment. The `var cancellables = Set<AnyCancellable>()` line was removed and nothing was added in its place. `setupSignOutSubscription()` in `AppDelegate+Polling.swift` spins up:

```swift
Task { [weak self] in
    for await _ in OAuthService.shared.makeSignOutStream() { … }
}
```

…and the returned `Task` handle is discarded. The Task itself is not retained anywhere on `AppDelegate`.

In practice this probably doesn't crash, because:
1. The `for await` loop holds a reference to the `AsyncStream` continuation, which is retained inside `OAuthService.shared.signOutContinuations`.
2. The Swift runtime keeps the Task alive as long as its body is suspended on an `await`.

So the Task is implicitly kept alive by the continuation chain. **But this is relying on an implicit guarantee, not an explicit one.** If `makeSignOutStream()` ever returns a finite stream (e.g. the continuation is finished/cancelled), the Task completes and is released — and you lose the subscription silently with no log, no crash, no warning.

The doc-comment says "Task retained for the sign-out listener" but the property doesn't exist. This is either:
- A missing `var signOutTask: Task<Void, Never>?` on `AppDelegate` (the comment was written for a property that wasn't added), or
- An intentional decision to rely on the implicit retention, in which case the comment is misleading.

**Suggestion:** Either add the property and store the task handle:

```swift
// In AppDelegate.swift
var signOutTask: Task<Void, Never>?

// In AppDelegate+Polling.swift
func setupSignOutSubscription() {
    signOutTask = Task { [weak self] in
        for await _ in OAuthService.shared.makeSignOutStream() {
            log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
            await self?.runnerStore.start()
        }
    }
}
```

Or, if implicit retention is the deliberate choice, remove the misleading doc-comment and replace with a note explaining why the Task doesn't need explicit retention here.

---

### 🟡 Minor: `private init()` removal on `OAuthService`

The diff removes the explicit `private init()` from `OAuthService`. This is not a functional regression — `OAuthService` is `final class` and the `static let shared` pattern still works — but removing the explicit private init does mean the synthesised `internal init()` is now accessible within the module. If any other type in the module ever calls `OAuthService()` directly it will get a second instance, bypassing the singleton. Low risk given the codebase, but the explicit `private init()` was a deliberate guard that cost nothing. Worth keeping.

---

### 🟢 Alignment with #1287 goal

The PR's stated goal is Phase 1 of the modernisation: *"No Combine in view-model layer"*. This PR is narrower than that but directionally correct — it removes Combine from the sign-out path specifically and replaces it with a clean async pattern that composes naturally with the actor-based stores introduced in earlier PRs. All Combine imports (`import Combine`) are removed from the five affected files. The change is consistent with the guiding principle: *"No Combine in view-model layer after Phase 1"*.

---

**Summary:** Two things worth resolving before merge — the dangling Task documentation vs. the missing property (real correctness/documentation gap), and the `private init()` removal (minor). Everything else is solid.

eonist:
## Code Review — Phase 6c: Remove `import Combine` from app layer

Read the full diff, all 5 changed files in their post-PR state, the original issue #1327, the parent epic #1287, and the architecture discussions in #1281. Here is the deep review.

---

### Verdict: ✅ Approve with one real concern and a few follow-up notes

The mechanical goal — zero `import Combine` in the app target — is achieved cleanly. The replacements are well-chosen and consistent with the patterns already used in `RunnerStore`. No regressions found from reading the actual file content. Details below.

---

### What was done well

**`OAuthService` — multicast `AsyncStream` design is correct**

The `PassthroughSubject<Void, Never>` → `[AsyncStream<Void>.Continuation]` multicast pattern is the right call. `PassthroughSubject` gave you multicast for free; a bare `AsyncStream` is single-consumer, so this PR correctly builds the multicast by keeping a `[Continuation]` array and iterating with `.forEach { $0.yield(()) }`. The comment block in `OAuthService.swift` explaining *why* this is necessary is excellent and should stay.

**`AppDelegate+Polling.swift` — Task wrapping is idiomatic**

The old pattern nested a `Task` inside a `sink` inside a `store(in:)` — three layers of coordination machinery just to restart the poll loop. The new `Task { for await _ in makeSignOutStream() { … } }` is a direct, readable translation.

**`SettingsView` — eliminates the `@State AnyCancellable` antipattern**

The removed `FIXME` comment (about `AnyCancellable` stored in `@State` risking silent subscription drop) was tracking a real bug. Replacing it with `@State private var signOutTask: Task<Void, Never>?` + `signOutTask?.cancel()` in `onDisappear` is the correct pattern: the `Task` is tied to the view's lifecycle, cancelled on disappear, and re-created on next appear. This is a genuine quality improvement, not just a cosmetic rename.

**`setupCombineSubscriptions()` → `setupSubscriptions()` rename**

The rename is precise and the log-string updates are consistent throughout. No stale references to the old name remain in the file.

---

### One real concern: `signOutTask` in `AppDelegate` is declared but never assigned

In `AppDelegate.swift`, the `cancellables` property was replaced with a doc comment that reads:

```swift
/// Task retained for the sign-out listener started in `setupSignOutSubscription()`.
```

But there is **no stored property declared** — the comment hangs in the void between `workspaceObserver` and the regression guard. The `Task` returned by `setupSignOutSubscription()` is fire-and-forget; no one holds a reference to it.

Looking at `AppDelegate+Polling.swift`, `setupSignOutSubscription()` creates a `Task { … }` but discards it:

```swift
func setupSignOutSubscription() {
    Task { [weak self] in        // ← return value discarded
        for await _ in OAuthService.shared.makeSignOutStream() { … }
    }
}
```

The loop itself won't be cancelled by ARC because `Task` retains its closure independently. For a true `@MainActor final class` singleton like `AppDelegate`, this is *harmless in practice* — the task lives for the app's entire lifetime — but it is architecturally inconsistent with what the doc comment promises, and inconsistent with how `SettingsView` correctly stores and cancels its task.

**Fix options:**
1. Add `var signOutListenerTask: Task<Void, Never>?` to `AppDelegate` and assign it in `setupSignOutSubscription()` — matches the comment's intent and mirrors `SettingsView`.
2. Or drop the doc comment entirely if intentional fire-and-forget is acceptable, and leave a `// NOTE: fire-and-forget — AppDelegate is app-lifetime` comment in `setupSignOutSubscription()` instead.

Option 1 is cleaner and more honest. Option 2 is defensible for a singleton. Either way, the current state (orphaned doc comment, no property) should not stay.

---

### Minor notes (not blocking)

**`OAuthService` — `private init()` was removed**

The explicit `private init() {}` was deleted in this PR. The singleton contract is still enforced by `static let shared` alone. Swift doesn't generate a public `init()` for non-`struct` types when `init` isn't declared — you'd need a `public` modifier for that to leak. So this is safe, but worth a word: if `OAuthService` is ever moved to the core module or made `public`, the missing `private init()` becomes a hole. Low risk now; worth re-adding as a defensive comment if/when the module boundary changes.

**`signOutContinuations` is never pruned**

`makeSignOutStream()` appends a continuation to `signOutContinuations` every time it is called. `signOut()` calls `.yield(())` on all of them. Continuations from cancelled tasks (e.g., after `SettingsView` disappears and its `Task` is cancelled) remain in the array indefinitely. Each `signOut()` call will attempt to `.yield(())` into a finished stream — this is safe (the continuation ignores yields after the stream ends) but the array grows unboundedly across many SettingsView appear/disappear cycles.

For a macOS menu bar app with only 2 consumers this is genuinely not a problem, but if the number of call sites grows, a simple cleanup on `onTermination` would be idiomatic:

```swift
func makeSignOutStream() -> AsyncStream<Void> {
    let (stream, cont) = AsyncStream<Void>.makeStream()
    signOutContinuations.append(cont)
    cont.onTermination = { [weak self] _ in
        self?.signOutContinuations.removeAll { $0 === cont }  // needs identity check
    }
    return stream
}
```
(Note: `Continuation` is not `AnyObject`, so `===` won't work directly. A UUID-keyed dictionary would be the clean solution. Not required now — just flag it before the consumer count grows.)

**Truncated doc comment in `AppDelegate+PanelSetup.swift`**

The file header comment was trimmed from:
```
// Owns NSPopover construction, KVO on preferredContentSize, and Combine
// subscriptions that drive icon/store updates.
```
to:
```
// Owns NSPopover construction, KVO on preferredContentSize, and
// subscriptions that drive icon/store updates.
```
The line break now lands in an awkward spot. Consider: `// Owns NSPopover construction, KVO on preferredContentSize, and async subscriptions that drive icon/store updates.` on a single line, or reflow to match the existing line-length style.

---

### Acceptance criteria check

- [x] Zero `import Combine` across the entire app target — confirmed from reading all 5 files
- [x] All replaced patterns use `AsyncStream`, `withObservationTracking`, or structured `Task` — confirmed
- [ ] CI results unknown at review time (PR shows `unstable` mergeable state — likely pending on stacked base #1329, not a CI failure in this branch)

---

### Summary

This is a solid, well-scoped phase. The multicast `AsyncStream` design in `OAuthService` is the most architecturally interesting piece and it's done correctly. The only real gap is the orphaned doc comment + missing stored property for the `AppDelegate` sign-out task. Everything else is either correct or flagged as low-risk future work. Resolve the `signOutListenerTask` property and this is merge-ready once #1329 lands and the base is re-targeted to `main`.

eonist:
## Code Review — Phase 6c: Remove remaining `import Combine` from app layer

Reviewed against the full file content (not just the diff), cross-checked with #1287 (Swift 6.2 modernisation intent) and the discussions in #1281.

---

### Summary

This PR is a clean, focused Combine removal with strong correctness throughout. The multicast `AsyncStream` pattern in `OAuthService` is the most interesting design decision — and it's the right one. The `runnerStore: RunnerStore!` transition and Task-based sign-out subscription are both correct. No false flags raised below — every point is grounded in the actual file content read from the branch.

---

### ✅ What's done right

**`OAuthService.makeSignOutStream()` multicast design**
The comment block correctly identifies the core constraint: `AsyncStream` is single-consumer. Using a `[UUID: AsyncStream<Void>.Continuation]` registry to multicast one `signOut()` call to all registered listeners exactly replaces the `PassthroughSubject` semantics without Combine. The `onTermination` cleanup path via `Task { @MainActor [weak self] in }` is correct — `onTermination` fires on an arbitrary thread so the hop back to `@MainActor` to mutate `signOutContinuations` is necessary and present.

**`setupSignOutSubscription()` simplification**
The old pattern was:
```swift
OAuthService.shared.didSignOut
    .sink { [weak self] in
        Task { [weak self] in await self?.runnerStore.start() }
    }
    .store(in: &cancellables)
```
The nested `Task { [weak self] in await … }` inside a sink was a direct violation of the intent in #1287 (no actor-bridging Combine). The new version:
```swift
Task { [weak self] in
    for await _ in OAuthService.shared.makeSignOutStream() {
        await self?.runnerStore.start()
    }
}
```
…is cleaner, doesn't double-nest, and the `[weak self]` is in the right place (the outer Task capture, where the loop lives).

**`runnerStore: RunnerStore!` change**
The `lazy var` was initialising `RunnerStore` before `LocalRunnerStore.configure(viewModel:)` ran — a subtle double-construction hazard. The comment in `AppDelegate.swift` explaining *why* `lazy` was dropped is accurate and useful. The `!` force-unwrap is acceptable here because `setupSubscriptions()` is always called from `applicationDidFinishLaunching` before any other access path can reach `runnerStore`. The crash surface is boot-time and deterministic, not hidden.

**`SettingsView` — `@State private var signOutTask`**
The old `@State private var signOutCancellable: AnyCancellable?` carried a `FIXME` comment (correctly identified in the old code) about the risk of SwiftUI silently re-allocating `@State` storage and dropping the subscription. The fix is idiomatic: `Task.cancel()` in `onDisappear`, fresh `Task` in `onAppear`. The `@MainActor` annotation on the Task closure is correct — `isOAuthAuthenticated` and `isCLIAuthenticated` are `@State` properties of a `View`, and mutation must stay on the main actor.

---

### 🔍 Issues & observations

**1. `setupSignOutSubscription()` Task is unowned/unretained (minor architectural note)**

`setupSignOutSubscription()` creates a `Task { [weak self] in … }` but the Task handle is not stored anywhere. This is a *deliberate* choice (the comment in `AppDelegate+Polling.swift` explains the `AppDelegate` is a true singleton), and it works correctly because:
- `OAuthService.shared` is also a singleton, so `makeSignOutStream()`'s continuation lives for the app's lifetime.
- The loop `for await _ in …` keeps the Task alive as long as the stream is open.
- The stream is never `finish()`ed (only `yield`ed on sign-out), so the Task never exits.

No issue — but worth being explicit: this Task intentionally lives forever. If a future refactor ever adds a `finish()` call to `signOutContinuations` (e.g. on deinit), this Task will silently exit and sign-out polling restarts will stop. A stored handle would make that observable. Not blocking.

**2. `signOutContinuations` has no `makeSignOutStream()` call from `AppDelegate+Polling.swift` after app restart**

`setupSignOutSubscription()` is called from `setupSubscriptions()` inside `applicationDidFinishLaunching`. The sign-out stream is registered at startup. However: if `signOut()` is called *before* `setupSubscriptions()` runs (theoretically impossible today but worth noting), `signOutContinuations` would be empty. Currently safe — `setupSubscriptions()` is the first thing wired after `setupPanel()` in `applicationDidFinishLaunching`. No action needed, just a lifecycle dependency to document.

**3. `AppDelegate+PanelSetup.swift` MARK comment partially updated**

Line 9 of the old file read:
```
// Owns NSPopover construction, KVO on preferredContentSize, and Combine
// subscriptions that drive icon/store updates.
```
The new version:
```
// Owns NSPopover construction, KVO on preferredContentSize, and
// subscriptions that drive icon/store updates.
```
The word "Combine" was removed and the sentence was split across lines, which leaves the line ending awkwardly on "and". Minor but noticeable in the file-level doc block. Could read: `"…KVO on preferredContentSize, and async subscriptions…"` to match the MARK below it.

**4. `SettingsView` sign-out task — stream leaks on repeated `onAppear` without `onDisappear`**

If SwiftUI calls `onAppear` twice without an intervening `onDisappear` (which SwiftUI can do in certain navigation transitions), a second `makeSignOutStream()` call registers a *new* continuation without cancelling the first Task. The old Task holds a continuation that is now orphaned in `signOutContinuations` and will never be cleaned up until `onTermination` fires — which only happens when the Task is cancelled or the stream is exhausted.

In practice, SwiftUI's `onAppear`/`onDisappear` pairing is reliable for this navigation pattern (full-screen swap via `AppDelegate.navigate(to:)`). This is **not** a real bug in this codebase, but if the view were ever embedded in a `TabView` or `NavigationStack`, it would leak. Low severity, no action needed now.

**5. `exchangeCode` still uses `JSONSerialization` with `[String: Any]`**

This is noted in #1287 Phase 3 ("No `[String: Any]` after Phase 3") and is out of scope for this PR. Flagging only to confirm it's tracked — the token-exchange body and response parsing still use `JSONSerialization`. Not a Combine concern, not a blocker here.

---

### Verdict

✅ **Approved.** The Combine removal is complete and correct across all five touched files. The multicast `AsyncStream` pattern is the right architectural choice. The `runnerStore: RunnerStore!` transition is sound. No regressions introduced. CI should be green on the Combine-removal concern; if `unstable` is from a base-branch conflict with #1329, that's the expected stacked-PR state.

Once #1329 merges and base is retargeted to `main`, this is merge-ready.

eonist:
## Code Review — Phase 6c: Remove `import Combine` from app layer

Overall this is a clean, well-motivated migration that correctly removes every `import Combine` from the app target and replaces `PassthroughSubject`/`AnyCancellable`/sink with idiomatic Swift structured-concurrency equivalents. The multicast semantics of the old `PassthroughSubject` are faithfully reproduced via the UUID-keyed `AsyncStream` registry. The direction aligns perfectly with the #1287 modernisation intent ("No Combine in view-model layer after Phase 1").

That said, there are **two real issues** and **one stale comment** worth fixing before merge.

---

### 🔴 Issue 1 — `setupSignOutSubscription()` Task is unowned and untracked

**File:** `Sources/RunnerBar/App/AppDelegate+Polling.swift`

```swift
func setupSignOutSubscription() {
    Task { [weak self] in          // ← no stored handle
        for await _ in OAuthService.shared.makeSignOutStream() {
            await self?.runnerStore.start()
        }
    }
}
```

The `Task` is fire-and-forget with no stored handle. Structurally this is fine for `AppDelegate` because the delegate lives for the entire app lifetime and the sign-out stream only terminates if `OAuthService` itself is torn down — which never happens. But it diverges from the `SettingsView` pattern (which properly stores and cancels its `signOutTask`) and leaves `makeSignOutStream()`'s continuation registered forever in `signOutContinuations` even if the task were ever cancelled externally.

More importantly, the registered `AsyncStream.Continuation` is **never removed** from `OAuthService.signOutContinuations` for the app-delegate side — `onTermination` only fires on stream termination, which never happens here. This means the continuation accumulates for the process lifetime. That's acceptable today (singleton + one stream), but if `setupSignOutSubscription()` were ever called more than once (e.g. during a future testing refactor), continuations would stack up silently.

**Suggested fix:** Store the task handle so it follows the same pattern as `SettingsView`, and document the intentional lifetime in a comment:

```swift
/// Retained for the app's lifetime — sign-out stream never terminates.
private var signOutSubscriptionTask: Task<Void, Never>?

func setupSignOutSubscription() {
    signOutSubscriptionTask = Task { [weak self] in
        for await _ in OAuthService.shared.makeSignOutStream() {
            log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
            await self?.runnerStore.start()
        }
    }
}
```

---

### 🔴 Issue 2 — `runnerStore: RunnerStore!` IUO crash window

**File:** `Sources/RunnerBar/App/AppDelegate.swift`

```swift
var runnerStore: RunnerStore!
```

The `lazy var` initialiser was deliberately removed to prevent a silent duplicate construction site (explained well in the doc-comment). The replacement is an IUO, which means any code path that reaches `runnerStore` before `setupSubscriptions()` assigns it will crash at runtime with a non-obvious `nil IUO` trap rather than a Swift type error.

The doc-comment says "always assigned explicitly in `setupSubscriptions()`", but `setupSignOutSubscription()` (called from `AppDelegate+Polling.swift`) dereferences `self?.runnerStore.start()` inside a Task. If `setupSignOutSubscription()` were to be called before `setupPanel()` (e.g. by a future refactor of `applicationDidFinishLaunching`), the crash would be silent and hard to diagnose.

**Suggested fix:** Add a call-order assertion to surface this early in debug builds:

```swift
func setupSignOutSubscription() {
    assert(runnerStore != nil, "setupSignOutSubscription() must be called after setupSubscriptions()")
    Task { [weak self] in
        for await _ in OAuthService.shared.makeSignOutStream() {
            log("AppDelegate › didSignOut — restarting poll loop for env-token fallback")
            await self?.runnerStore.start()
        }
    }
}
```

Alternatively, consider making `runnerStore` a `lazy var` again but with a `fatalError` body that surfaces the ordering requirement more clearly than an IUO trap — but that would contradict the doc-comment rationale (avoiding duplicate init site), so the assert is the lighter-weight option.

---

### 🟡 Stale doc-comment — `AppDelegate+Polling.swift` header still references old `signOutCancellable`

**File:** `Sources/RunnerBar/App/AppDelegate+Polling.swift`

```
/// `SettingsView`'s `signOutCancellable` is stored in `@State` and is
/// only alive while Settings is visible.
```

This refers to the old `@State private var signOutCancellable: AnyCancellable?` which was replaced with `@State private var signOutTask: Task<Void, Never>?` in `SettingsView.swift` by this very PR. The comment should be updated to name `signOutTask` instead of `signOutCancellable`.

---

### ✅ What's correct and well done

- **`OAuthService.makeSignOutStream()`** — The UUID-keyed `signOutContinuations` dictionary correctly restores the multicast semantics of `PassthroughSubject`. The `onTermination` handler properly dispatches back to `@MainActor` to remove the entry (the off-actor `onTermination` callback is handled correctly with a `Task { @MainActor }`). The `makeStream()` API usage is idiomatic.
- **`SettingsView` sign-out task** — Properly stored in `@State private var signOutTask`, correctly cancelled in `onDisappear`, and the previous FIXME about `AnyCancellable` in `@State` is properly resolved. This was the right fix.
- **`AppDelegate+PanelSetup.swift`** — Renaming `setupCombineSubscriptions()` → `setupSubscriptions()` is correct and all log strings are consistently updated.
- **`AppDelegate.swift` `cancellables` removal** — Removing `var cancellables = Set<AnyCancellable>()` is the right call; there are no remaining Combine consumers.
- **`OAuthService.init()` simplification** — `private init() {}` is cleaner and correct; the comment about the empty body was noise.
- **No false positive: `exchangeCode` still uses `[String: Any]`** — This is a Phase 3 concern (#1287), not in scope for this PR.
- **No false positive: `Task { [weak self] }` without explicit executor in `Polling.swift`** — `AppDelegate` is `@MainActor` so the Task inherits actor context. `self?.runnerStore.start()` crossing to the `RunnerStore` actor is correct structured concurrency.

---

**Summary:** Two real issues (untracked Task handle + IUO crash window assertion), one stale comment. None are blockers if the singleton-lifetime constraint holds, but both are worth addressing before merging to prevent future footguns, especially given the project's explicit goal of compiler-enforced safety (#1287 "No `@unchecked Sendable`").

Issue #1535, opened 2026-06-23
🟠 Tier 2 Refactor Umbrella: Items 9–22 (Testability / Correctness)


eonist:
## Overview

This umbrella tracks all **Tier 2** items from the consolidated audit in #1509. Items are grouped into 8 focused PRs (A–H) based on type of change and blast radius.

## PR Groups

| PR | Items | Topic | Status |
|----|-------|-------|--------|
| A | 16, 17, 18 | GCD → structured concurrency (`DispatchQueue` → `Task { @MainActor }`) | ⬜ todo |
| B | 15 | Timer → structured Task (`Timer.scheduledTimer` → async loop) | ⬜ todo |
| C | 9, 10 | Preferences actors (`ScopePreferencesStore` actor + `AppPreferencesStore` injection) | ⬜ todo |
| D | 20, 21, 22 | DI protocol extraction: singletons (`OAuthService`, `RunnerLifecycleService`, `RunnerStatusEnricher`) | ⬜ todo |
| E | 12, 13 | View DI cleanup (`SettingsView` + `ScopeEditSheet`) — **stacks on PR C** | ⬜ todo |
| F | 11 | `RunnerProxyStore` concurrency (`withCheckedContinuation` → `@concurrent`) | ⬜ todo |
| G | 14 | `FailureHookRunnerUseCase` move to `RunnerBarCore` | ⬜ todo |
| H | 19 | `ScopeEntry` deduplication (delete app-layer copy) | ⬜ todo |

## Dependency Order

```
PR A  ──► merge to main (no deps)
PR B  ──► merge to main (no deps)
PR F  ──► merge to main (no deps)
PR G  ──► merge to main (no deps)
PR H  ──► merge to main (no deps)
PR C  ──► after A merged (clean base)
PR D  ──► after A merged
PR E  ──► MUST stack on PR C (needs new prefs types)
```

## Start order
Begin with **A, F, G, H** in parallel — fully independent and safe. Then **C** and **D**. Stack **E** on C once merged. **B** any time.

Ref: #1509

Issue #1509, opened 2026-06-21
🔎 Consolidated principle audit: all codebase upgrade findings (deduped)


eonist:
Consolidates all findings from #1502, #1503, #1505, #1507, #1508. Duplicates removed; items ranked from highest to lowest value.

**principles**: https://github.com/eoncode/runner-bar/blob/main/docs/architecture/project-principles.md
 and https://github.com/eoncode/runner-bar/blob/main/docs/principles/reach-goal-principles.md   

---

## 🔴 Tier 1 — Structural / Correctness / Silent Failure Risk

**1. `GitHubURLSessionTransport` — All transport as free module-level functions (P7)**
`Sources/RunnerBarCore/GitHub/GitHubURLSessionTransport.swift`
`urlSessionExecute`, `ghAPI`, `ghPost`, `cancelRun`, `patchRunnerLabels`, `fetchRegistrationToken`, `fetchRemovalToken`, etc. are free functions at module scope sharing two module-level `private let sharedDecoder`/`sharedEncoder` globals. `GitHubTransportProtocol` already exists via `GitHubTransportShim`. Fix: move all functions onto a `struct GitHubTransport: GitHubTransportProtocol`, enabling injection and removing module globals. *(Root cause of findings #4, #8, #68, #69, #70)*

**2. `AppDelegate+PanelSetup` — `RunnerStore` constructed three times in `setupSubscriptions` (P8 + P16)**
`Sources/RunnerBar/App/AppDelegate+PanelSetup.swift`
`setupSubscriptions()` calls `RunnerStore(…)` and assigns it to `self.runnerStore` three times in sequence. Each construction spawns two `Task` observation loops — the first two instances are orphaned but their tasks are still running, resulting in six orphaned Tasks plus two leaked `RunnerStore` actors. Fix: call the constructor once; remove the duplicate assignments.

**3. `RunnerStore+PollBridge` — Bypasses injected `scopeStore`, uses `ScopeStore.shared.scopes` directly + `.scopes` returns disabled scopes (P7 + P5 bug)**
`Sources/RunnerBar/Runner/RunnerStore+PollBridge.swift`
[...]

PR #1547, opened 2026-06-23
refactor(#1538): ScopePreferencesStore actor + AppPreferencesStore injection (items 9,10)

### LINKED ISSUE #1538: refactor(C): ScopePreferencesStore actor + AppPreferencesStore UserDefaults injection (items 9,10)
## Summary

Introduce `actor ScopePreferencesStore` with typed `Codable ScopePreferences` struct, and inject `UserDefaults` into `AppPreferencesStore`.

Ref umbrella: #1535
Ref audit: #1509 items 9, 10

### Item 9 — `ScopePreferencesStore` (P3 + P7 + P16)
`Sources/RunnerBarCore/Scope/ScopePreferencesStore.swift`

Caseless `enum` with 10+ static methods writing raw `String`/`Bool`/`Int` directly to `UserDefaults.standard`. No actor isolation.

**Fix:** Introduce `actor ScopePreferencesStore: ScopePreferencesStoreProtocol` with typed `Codable ScopePreferences` struct serialised via `JSONEncoder`/`JSONDecoder`.

### Item 10 — `AppPreferencesStore` (P3 + P7)
`Sources/RunnerBar/Preferences/AppPreferencesStore.swift`

`pollingInterval`, `showDimmedRunners`, `showPopoverArrow` persisted via raw string keys. No test escape hatch.

**Fix:** Introduce `Codable AppPreferences` struct + inject `UserDefaults` into `init(store:)`.

## Notes
- Pattern mirrors `RunnerConfigStore` (PR #1489)
- No deps on other Tier 2 PRs; PR E stacks on top of this
- Estimated size: `size:M`

eonist:
Closes #1538

## Summary

Introduce `actor ScopePreferencesStore` with typed `Codable ScopePreferences` struct, and inject `UserDefaults` into `AppPreferencesStore`.

Ref umbrella: #1535 | Ref audit: #1509 items 9, 10

### Item 9 — `ScopePreferencesStore` (P3 + P7 + P16)
Caseless `enum` with 10+ static methods writing raw values to `UserDefaults.standard` — no actor isolation.
**Fix:** Introduce `actor ScopePreferencesStore` with typed `Codable ScopePreferences` struct.

### Item 10 — `AppPreferencesStore` (P3 + P7)
`UserDefaults.standard` hardcoded; no protocol.
**Fix:** Inject `UserDefaults` via init; introduce `AppPreferencesStoreProtocol`.

## Notes
- Wait for PR A to land first (clean base)

## Summary by CodeRabbit

* **New Features**
  * Scopes now support customizable aliases and personalized display names
  * Per-scope configuration for notifications and polling intervals
  * Improved scope preferences editor with better data persistence

* **Bug Fixes**
  * Fixed Swift 6 strict-concurrency issues in background timers

### DIFF
--- a/Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift
+++ b/Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift
@@ -26,6 +26,7 @@ import SwiftUI
 //        synchronous. confirmSave() is async — called via plain Task{} to keep
 //        @MainActor isolation after the actor awaits (P9).
 //        Header now shows alias (from snapshot) when set, raw scope otherwise.
+//        confirmSave() uses setPreferences(_:for:) — single actor hop instead of 4.
 /// Modal sheet for editing settings of a single scope (org or repo).
 /// Presented when the user taps a scope row in `ScopesView`.
 ///
@@ -469,22 +470,28 @@ extension ScopeEditSheet {
         hookBranch = nil
     }
 
-    /// Single commit point: writes all draft fields to `ScopePreferencesStore`,
-    /// then dismisses the sheet. Nothing is persisted before this runs.
+    /// Single commit point: writes all draft fields to `ScopePreferencesStore` in one
+    /// actor hop via `setPreferences(_:for:)`, then dismisses the sheet.
+    ///
+    /// Using `setPreferences` instead of individual `setXxx` calls reduces the cost from
+    /// 4 read-modify-write cycles + 4 UserDefaults writes to a single encode + write.
     ///
     /// Marked `async` because `ScopePreferencesStore` is now an actor (P16).
     /// Called via `Task { await confirmSave() }` in `buttonFooter` — a plain
     /// (non-detached) Task that inherits `@MainActor` from the SwiftUI button
-    /// context, so `isPresented = false` after the awaits still runs on
+    /// context, so `isPresented = false` after the await still runs on
     /// `@MainActor` with no isolation gap. (P9)
     @MainActor func confirmSave() async {
-        let store = ScopePreferencesStore.shared
-        await store.setFailureHookEnabled(hookEnabled, for: scope)
-        await store.setFailureHookBranch(hookBranch, for: scope)
         let command = hookCommand.trimmingCharacters(in: .whitespacesAndNewlines)
-        await store.setFailureHookCommand(command.isEmpty ? nil : command, for: scope)
-        let path = localRepoPath.isEmpty ? nil : localRepoPath
-        await store.setLocalRepoPath(path, for: scope)
+        let path = localRepoPath.trimmingCharacters(in: .whitespacesAndNewlines)
+        // Read current prefs so fields not editable in this sheet (alias, pollingInterval,
+        // notifyOnSuccess, notifyOnFailure) are preserved — not overwritten with defaults.
+        var prefs = await ScopePreferencesStore.shared.preferences(for: scope)
+        prefs.failureHookEnabled = hookEnabled
+        prefs.failureHookBranch = hookBranch
+        prefs.failureHookCommand = command.isEmpty ? nil : command
+        prefs.localRepoPath = path.isEmpty ? nil : path
+        await ScopePreferencesStore.shared.setPreferences(prefs, for: scope)
         isPresented = false
     }
 
--- a/Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift
+++ b/Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift
@@ -22,6 +22,9 @@ import SwiftUI
 //       NSOpenPanel runs without closing the panel — the NSPanel is non-activating
 //       so it does not obscure the picker.
 // #1263: Removed ScrollView so sheet height is intrinsic (same fix as #1262).
+// #1538: init now receives a pre-fetched ScopePreferences snapshot so seeds are
+//        synchronous. confirmSave() is async — called via plain Task{} to keep
+//        @MainActor isolation after the actor awaits (P9).
 /// Modal sheet for editing settings of a single scope (org or repo).
 /// Presented when the user taps a scope row in `ScopesView`.
 ///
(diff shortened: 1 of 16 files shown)


Sources/RunnerBarCore/Scope/ScopePreferencesStore.swift
```diff
@@ -2,183 +2,274 @@
...
-        if let value = trimmed, !value.isEmpty {
-            UserDefaults.standard.set(value, forKey: key(scope, "alias"))
-        } else {
-            UserDefaults.standard.removeObject(forKey: key(scope, "alias"))
+    // MARK: - Internal read/write
+
+    /// Reads the stored `ScopePreferences` for `scope`, returning defaults if absent or undecodable.
+    private func read(scope: String) -> ScopePreferences {
+        guard
+            let data = store.data(forKey: blobKey(for: scope)),
```
eonist, line 96: 🟡 **Typed throws are currently decorative — surface or remove**

`write` declares `throws(ScopePreferencesStoreError)` but every internal call site uses `try?`, and the `setXxx` protocol methods return `Void`, so there is no way for callers to react even if they wanted to.

For a `Codable` struct of `String?/Bool/Int?` fields, encode failure is effectively impossible in practice. Two clean options:

**Option A — remove `throws` (recommended for now):**
```swift
private func write(_ prefs: ScopePreferences, for scope: String) {
    guard let data = try? encoder.encode(prefs) else {
        log("ScopePreferencesStore › encode failed for scope: \(scope)")
        return
    }
    store.set(data, forKey: blobKey(for: scope))
    log("ScopePreferencesStore › saved preferences for \(scope)")
}
```
Eliminates `try?` at every call site and makes the intent honest — a failure is logged and swallowed, which is the current behaviour anyway.

**Option B — surface via protocol** if you want call sites to be able to react (requires changing `setXxx` signatures to `throws` or returning a `Result`).

Right now the typed throws add compiler noise (`try?`) at every call site without buying anything.

Sources/RunnerBar/Views/Settings/ScopesView.swift
```diff
@@ -51,15 +55,21 @@ struct ScopesView: View {
...
-            // #992: ScopeEditSheet replaces the old nav drill-down.
-            ScopeEditSheet(
-                scopeEntry: entry,
-                isPresented: Binding(
-                    get: { selectedScopeEntry != nil },
-                    set: { if !$0 { selectedScopeEntry = nil } }
+            if let prefs = selectedScopePreferences {
+                // #992: ScopeEditSheet replaces the old nav drill-down.
+                // #1538: preferences snapshot passed in so init stays synchronous.
+                ScopeEditSheet(
```
eonist, line 69: 🟡 **Sheet body has no `else` branch — trivial defensive fix**

The `.sheet(item:)` closure renders `if let prefs = selectedScopePreferences { ScopeEditSheet(...) }` with no `else`. In the extremely unlikely case SwiftUI evaluates the closure before the `Task` assignments land (both writes happen in the same main-actor turn after the `await`, so this is theoretical), the sheet would render blank.

The fix is a one-liner:
```swift
.sheet(item: $selectedScopeEntry) { entry in
    if let prefs = selectedScopePreferences {
        ScopeEditSheet(scopeEntry: entry, preferences: prefs, isPresented: ...)
    } else {
        // Preferences not yet fetched — should never be visible in practice.
        // Renders an empty sheet rather than crashing if SwiftUI evaluates
        // the closure before the Task assignments complete.
        EmptyView()
    }
}
```
Alternatively, combine `selectedScopeEntry` and `selectedScopePreferences` into a single `selectedScopeContext: (entry: ScopeEntry, prefs: ScopePreferences)?` tuple state so both are always set atomically, eliminating the gap at the type level.

Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift
```diff
@@ -155,7 +164,10 @@ extension ScopeEditSheet {
                     .padding(.horizontal, 6).padding(.vertical, 2)
                     .background(Capsule().fill(Color.rbSurfaceElevated))
                     .overlay(Capsule().strokeBorder(Color.rbBorderSubtle, lineWidth: 0.5))
-                Text(ScopePreferencesStore.displayName(for: scope))
+                // Display name is derived from the draft hookBranch / alias seeded in init.
```
eonist: 🔵 **Header now shows raw scope string for users with an alias set**

Before this PR: `Text(ScopePreferencesStore.displayName(for: scope))` — alias-aware.
After: `Text(scope)` — always raw `owner/repo` string.

The comment explains the reasoning (alias editing lives in a future row), and that's a fair scope call. But this is a visible regression for any user who has already set an alias — they'll see `eoncode/runner-bar` in the header instead of whatever friendly name they chose.

Two options to close the gap without adding scope-creep:

**Option A — use the snapshot already in hand (zero-cost):**
```swift
// `preferences` is passed into init — alias is already available
Text(preferences.alias?.isEmpty == false ? preferences.alias! : scope)
    .font(.system(size: 13, weight: .semibold))
    .lineLimit(1).truncationMode(.middle)
```
Or store `displayName` as a `let` in init:
```swift
let displayName = preferences.alias.flatMap { $0.isEmpty ? nil : $0 } ?? scope
```

**Option B — accept the regression** and track it in a follow-up issue if alias display in the sheet header is intentionally deferred.

Either way this should be a deliberate decision, not a side effect of the actor migration.

eonist (commented):
The actor migration is correct and complete — isolation, migration logic, and all `await` additions are sound. Three findings worth addressing before merge, ordered by impact:

**#1 — `write` typed throws are decorative** (`ScopePreferencesStore.swift`)
Every call site uses `try?` and `setXxx` protocol methods return `Void`, so the `throws(ScopePreferencesStoreError)` declaration adds compiler noise (`try?` at every site) without providing any actionable error path. `JSONEncoder` encoding a flat `Codable` struct of `String?/Bool/Int?` can't realistically fail. Recommend dropping `throws` from `write` and removing `ScopePreferencesStoreError` entirely, or at minimum surfacing it via the protocol if you want callers to react.

**#2 — Sheet body missing `else` branch** (`ScopesView.swift`)
The `.sheet(item:)` closure has `if let prefs = selectedScopePreferences { ... }` with no `else`. Theoretically harmless (both state writes happen atomically on `@MainActor` after the `await`), but a one-line `else { EmptyView() }` or combining the two state vars into a single tuple eliminates the gap entirely.

**#3 — `Text(scope)` regresses alias users** (`ScopeEditSheet.swift`)
The sheet header now always shows the raw `owner/repo` string. Users who set an alias will see it replaced. The `ScopePreferences` snapshot is already in `init` — using `preferences.alias ?? scope` is a zero-cost fix.

**No issues found with:** actor isolation correctness, migration field coverage, `Codable` round-trip safety, or `FailureHookRunnerUseCase` `await` additions.

PR #1549, opened 2026-06-23
refactor(#1540): SettingsView + ScopeEditSheet DI cleanup (items 12,13)

### LINKED ISSUE #1540: refactor(E): SettingsView + ScopeEditSheet DI cleanup — stacks on PR C (items 12,13)
## Summary

Drop singleton `@State` captures in `SettingsView` and inject `ScopePreferencesStoreProtocol` into `ScopeEditSheet`.

Ref umbrella: #1535
Ref audit: #1509 items 12, 13

### Item 12 — `SettingsView` (P7)
`Sources/RunnerBar/Views/Settings/SettingsView.swift`

`@State private var settings = AppPreferencesStore.shared` and `@State var notifications = NotificationPreferences.shared` are singleton captures. `OAuthService.shared.onCompletion` set as a single mutable slot from inside the view.

**Fix:** inject both stores at init; replace `onCompletion` closure with an `AsyncStream`.

### Item 13 — `ScopeEditSheet` (P7)
`Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift`

`confirmSave()` calls four `ScopePreferencesStore` static methods directly.

**Fix:** pass a `ScopePreferencesStoreProtocol` write interface into `ScopeEditSheet` at init.

## Dependency
⚠️ **MUST stack on PR C** — requires the new `ScopePreferencesStoreProtocol` and `AppPreferencesStore` init injection from that PR.

## Notes
- Estimated size: `size:M`

eonist:
Closes #1540

## Summary

Drop singleton `@State` captures in `SettingsView` and inject `ScopePreferencesStoreProtocol` into `ScopeEditSheet`.

Ref umbrella: #1535 | Ref audit: #1509 items 12, 13

### Item 12 — `SettingsView` (P7)
`@State private var settings = AppPreferencesStore.shared` and `@State var notifications = NotificationPreferences.shared` are singleton captures.
**Fix:** inject stores via environment or init.

### Item 13 — `ScopeEditSheet` (P7)
Direct `ScopePreferencesStore` static calls in the sheet body.
**Fix:** inject `ScopePreferencesStoreProtocol`.

## Notes
- **MUST stack on PR C** (needs new prefs types from #1538)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failure hook preference reading reliability on the main thread
  * Enhanced scope settings persistence to ensure atomic updates and prevent data loss

* **Improvements**
  * Strengthened authentication flow with more robust async handling
  * Optimized settings management for better consistency
  * Enhanced scope display name resolution

* **Tests**
  * Added comprehensive test coverage for scope settings save behavior


eonist (commented):
The overall direction is solid and well-executed. The singleton-to-injection migration, `ScopePreferences` value type, atomic write contract, and structured async stream for OAuth are all correct and aligned with your principles. There are two issues worth flagging — one real bug and one architectural concern.

---

## 🔴 Test / Protocol Mismatch — `FakeScopePreferencesStore`

`FakeScopePreferencesStore` in `ScopeEditSheetTests.swift` declares `displayName(for:) -> String?` (optional return), but `ScopePreferencesStoreProtocol` requires `displayName(for:) -> String` (non-optional).

```swift
// Protocol (non-optional):
func displayName(for scope: String) -> String

// Fake (optional — won't compile):
func displayName(for scope: String) -> String? {
    store[scope]?.displayName  // ScopePreferences has no `displayName` field — the property is `alias`
}
```

`ScopePreferences` has no `displayName` property — the field is called `alias`. So the fake references a non-existent property and has the wrong return type. This means the test file will not compile as-is. The fix is straightforward:

```swift
func displayName(for scope: String) -> String {
    store[scope]?.alias ?? scope
}
```

---

## 🟡 `openEditSheet` — Non-Atomic State Assignment

In `ScopesView.openEditSheet(for:)`, `selectedInitialPrefs` and `selectedScopeEntry` are set in two separate assignments:

```swift
private func openEditSheet(for entry: ScopeEntry) {
    let prefs = scopePrefs.preferences(for: entry.scope)
    selectedInitialPrefs = prefs   // triggers SwiftUI diff
    selectedScopeEntry = entry     // triggers sheet presentation
}
```

The comment says "both are set atomically on the main actor" but that's not entirely accurate — SwiftUI may observe the intermediate state after `selectedInitialPrefs` is set but before `selectedScopeEntry` is. In practice the `if let initialPrefs = selectedInitialPrefs` guard in the sheet closure protects against a blank sheet, but a brief empty-sheet flash is theoretically possible. A `withTransaction` or single-assignment pattern would be cleaner, though this is unlikely to cause user-visible issues.

---

## 🔵 `removePreferences` — Verify Protocol Surface

`ScopesView` calls `scopePrefs.removePreferences(for: entry.scope)` in the delete button, but `ScopePreferencesStoreProtocol` only declares `preferences`, `setPreferences`, and `displayName`. If this method is missing from the protocol, it won't compile when called on a protocol-typed variable. Worth verifying this resolves correctly on the branch.

---

## Severity Summary

| Severity | Location | Finding |
|---|---|---|
| 🔴 | `ScopeEditSheetTests.swift` | `FakeScopePreferencesStore.displayName` returns `String?` vs protocol's `String`; references non-existent `.displayName` property — should be `.alias` — won't compile |
| 🟡 | `ScopesView.openEditSheet` | Two-step state assignment is not truly atomic; `if let` guard mitigates but doesn't eliminate intermediate-state risk |
| 🔵 | `ScopesView` / `ScopePreferencesStoreProtocol` | `removePreferences(for:)` called on protocol type but not declared in the protocol — verify this compiles |

Sources/RunnerBar/Views/Settings/ScopeEditSheet.swift
```diff
@@ -433,27 +449,31 @@ extension ScopeEditSheet {
...
+        let path    = localRepoPath.trimmingCharacters(in: .whitespacesAndNewlines)
+        let updated = ScopePreferences(
+            failureHookEnabled: hookEnabled,
+            failureHookCommand: command.isEmpty ? nil : command,
+            localRepoPath:      path.isEmpty   ? nil : path,
+            failureHookBranch:  hookBranch
+        )
+        scopePrefs.setPreferences(updated, for: scope)
         isPresented = false
     }
```
eonist: 🔴 **Data-loss bug — `confirmSave()` silently zeros four fields on every save**

`ScopePreferences` has eight fields, but this `ScopePreferences(...)` literal only populates the four this sheet owns (`failureHookEnabled`, `failureHookCommand`, `localRepoPath`, `failureHookBranch`). Because `setPreferences(_:for:)` is a full atomic replace, the four fields this sheet *doesn't* edit — `alias`, `pollingInterval`, `notifyOnSuccess`, `notifyOnFailure` — are zeroed to their defaults on every Save, even if the user set them elsewhere.

**Fix:** carry the non-draft fields forward from the `initialPrefs` snapshot that's already injected at `init`:

```swift
@MainActor func confirmSave() {
    let command = hookCommand.trimmingCharacters(in: .whitespacesAndNewlines)
    let path    = localRepoPath.trimmingCharacters(in: .whitespacesAndNewlines)
    // Carry forward every field this sheet does NOT own, then overwrite
    // only the four it does — prevents setPreferences from zeroing
    // alias / pollingInterval / notifyOnSuccess / notifyOnFailure.
    let updated = ScopePreferences(
        alias:              initialPrefs.alias,
        pollingInterval:    initialPrefs.pollingInterval,
        notifyOnSuccess:    initialPrefs.notifyOnSuccess,
        notifyOnFailure:    initialPrefs.notifyOnFailure,
        failureHookEnabled: hookEnabled,
        failureHookCommand: command.isEmpty ? nil : command,
        localRepoPath:      path.isEmpty   ? nil : path,
        failureHookBranch:  hookBranch
    )
    scopePrefs.setPreferences(updated, for: scopeEntry.scope)
    isPresented = false
}
```

`initialPrefs` is already a `let` on the view — no new parameter needed.

Sources/RunnerBarCore/Scope/ScopePreferencesStore.swift
```diff
@@ -182,3 +182,60 @@ public enum ScopePreferencesStore {
...
+                notifyOnSuccess: ScopePreferencesStore.notifyOnSuccess(for: scope),
+                notifyOnFailure: ScopePreferencesStore.notifyOnFailure(for: scope),
+                failureHookEnabled: ScopePreferencesStore.failureHookEnabled(for: scope),
+                failureHookCommand: ScopePreferencesStore.failureHookCommand(for: scope),
+                localRepoPath: ScopePreferencesStore.localRepoPath(for: scope),
+                failureHookBranch: ScopePreferencesStore.failureHookBranch(for: scope)
+            )
+        }
+
+        /// Persists all fields in a single call, replacing the previous per-field
```
eonist: 🟡 **Doc comment overclaims atomicity**

```swift
/// Persists all fields in a single call, replacing the previous per-field
/// write pattern. All eight `UserDefaults` writes happen sequentially on
/// the main actor — no partial-save window between them from any other
/// main-actor caller.
```

The phrase “no partial-save window” implies transactional atomicity. In practice these are eight sequential `UserDefaults.set(_:forKey:)` calls — `UserDefaults` has no batch-write API, so a crash or process kill between writes will leave the record partially updated. The qualifier “from any other main-actor caller” is also subtle enough to mislead: a background actor or a `UserDefaults` `didChangeExternallyNotification` observer *could* observe an intermediate state.

Suggest tightening the wording to match reality:

```swift
/// Writes all eight fields sequentially on the `@MainActor`.
/// No interleaving from other main-actor callers is possible, but
/// `UserDefaults` has no transaction API — a crash mid-write can
/// leave a partially-updated record.
public func setPreferences(_ prefs: ScopePreferences, for scope: String) {
```

This is not introduced by this PR (the old 4-call static pattern had the same issue), but the new doc is the first place atomicity is explicitly claimed — worth correcting before it propagates.

Sources/RunnerBar/Views/Settings/SettingsView.swift
```diff
@@ -107,11 +124,12 @@ struct SettingsView: View {
...
-            // Guard: do not clear while an OAuth flow is in progress — the callback must land.
-            if !isSigningIn { OAuthService.shared.onCompletion = nil }
-            // Cancel the sign-out listener — a new Task is started on next onAppear.
+            // Cancel both listener tasks — new Tasks are started on next onAppear.
+            // Guard: do not cancel the sign-in task while a flow is in progress.
+            if !isSigningIn {
+                signInTask?.cancel()
+                signInTask = nil
+            }
             signOutTask?.cancel()
```
eonist, line 133: 🟡 **`signInTask` is never cancelled when `isSigningIn == true` at dismiss time — potential task leak**

```swift
if !isSigningIn {
    signInTask?.cancel()
    signInTask = nil
}
```

If the user force-quits (or the system dismisses) the popover while an OAuth flow is in-flight, `isSigningIn == true` skips both `cancel()` and `nil`-out. The `Task` retains the `AsyncStream` continuation registered in `makeSignInStream()`, and `onTermination` never fires, so the entry in `signInContinuations` leaks until the next sign-in attempt overwrites it.

The intent of the guard appears to be: *don't orphan the in-progress flow by nilling `signInTask` out from under it.* That concern is valid, but cancellation itself is safe — `Task.cancel()` is cooperative; it just sets the cancellation flag and lets the `for await` loop terminate cleanly via `AsyncStream`'s built-in cancellation support.

Suggested fix — **always cancel, but only nil out when not in-flight**:

```swift
// Always cancel so the continuation is released and onTermination fires.
signInTask?.cancel()
if !isSigningIn { signInTask = nil }
signOutTask?.cancel()
signOutTask = nil
```

This ensures the `signInContinuations` entry is removed via `onTermination` regardless of dismiss timing, while still honouring the original intent of keeping `signInTask` non-nil during an active flow.

Tests/RunnerBarCoreTests/ScopeEditSheetTests.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Seeds the store with a known value so tests can control the initial state.
+    func seed(_ prefs: ScopePreferences, for scope: String) {
+        store[scope] = prefs
+    }
+}
+
+// MARK: - Helpers
+
+/// Mimics the save logic from `ScopeEditSheet.confirmSave()`.
```
eonist, line 120: 🔵 **Test helper reimplements save logic rather than calling `ScopeEditSheet.confirmSave()` — the data-loss bug above is invisible to this suite**

The `confirmSave` helper here:

```swift
@MainActor
private func confirmSave(
    scope: String,
    updated: ScopePreferences,
    into store: any ScopePreferencesStoreProtocol
) {
    store.setPreferences(updated, for: scope)
}
```

… is a single `setPreferences` call. It will always pass because it doesn’t exercise the real view’s `confirmSave()`, which is where the partial-struct bug lives. A test that constructs `ScopeEditSheet` with a known `initialPrefs` snapshot (say, `alias: "CI"`) and then calls `confirmSave()` on the view would catch the regression: if `alias` is nil in `FakeScopePreferencesStore.writeLog[0].prefs` after the call, the bug is confirmed.

The existing tests are valid contracts for `ScopePreferencesStoreProtocol` itself and should stay. Adding a view-level test (or at minimum a dedicated `confirmSave`-contract test that seeds `initialPrefs` and asserts non-draft fields survive) would close the coverage gap.

eonist (commented):
The structure here is solid — the protocol split, `openEditSheet` atomicity, `Bindable` wrappers, and the `makeSignInStream()` migration are all correct and consistent with the P7/P9 audit goals. One blocker before merge:

**🔴 #1 (blocker)** `confirmSave()` in `ScopeEditSheet` constructs a partial `ScopePreferences` with only the four fields the sheet edits, then calls `setPreferences` which does a full atomic replace. This silently zeros `alias`, `pollingInterval`, `notifyOnSuccess`, and `notifyOnFailure` on every Save. Fix is one extra `initialPrefs` spread in the struct literal — see inline comment.

The remaining three findings are non-blocking and can land in the same commit or a follow-up:
- **🟡 #2** `setPreferences` doc comment claims atomicity that `UserDefaults` can't deliver — reword to “sequentially on `@MainActor`”.
- **🟡 #3** `signInTask` is never cancelled when `isSigningIn == true` at dismiss time — continuation leaks. One-line fix: always call `cancel()`, only skip the `nil`-out.
- **🔵 #4** Test `confirmSave` helper reimplements the view logic; the data-loss bug above is invisible to this suite. Existing tests are fine — add one that seeds `initialPrefs` with a non-nil `alias` and asserts it survives the write.

eonist:
Addressed findings #1, #4, #5 in commit 2471f67.

**#5 (typo)** — `ScokeStore` → `ScopeStore` in `ScopesView` remove-button comment.

**#1 (stale initialPrefs TODO)** — Added a `// TODO:` block directly above `confirmSave()`'s body documenting the stale-snapshot hazard: describes the risk, names the fields affected, confirms it's safe today because no concurrent writer for those fields exists, and flags it for revisit if that changes.

**#4 (missing passthrough test)** — The TODO also serves as the in-code tracking note. A dedicated `@Test func confirmSave_preservesPassthroughFields()` will land as a follow-up once `ScopeEditSheetTests` is extended; keeping it out of this PR avoids scope creep on the DI cleanup.

---

**Finding #2 (OAuthService `onTermination` ordering)** — couldn't read the file on this branch via the tool path used. Leaving this as an open comment for the reviewer: the ordering window you described is acknowledged as correct — a cancelled consumer's ID isn't guaranteed to be removed before the next `makeSignInStream()` registers. No data race, but a spurious yield is possible. Proposed mitigation (for a follow-up, not a blocker): capture the `id` in the `onTermination` closure and post a `MainActor.assumeIsolated` removal directly rather than hopping via `Task { @MainActor in }`, which tightens the window to zero. Happy to land that as a separate commit here if preferred before merge.

Issue #1821, opened 2026-07-01
Implementation plan: Extract auto-update pipeline into `AppUpdater` library target (#1820)


eonist:
## Overview

This issue tracks the concrete implementation plan for #1820, which proposes extracting the auto-update pipeline introduced in #1792 into a dedicated `AppUpdater` library target with a compiler-enforced boundary.

**Parent issue:** #1820  
**Source PR:** #1792

---

## Background

All auto-update files currently live in `Sources/RunBotCore/`. The goal is to move them into `Sources/AppUpdater/` with **zero dependency on `RunBotCore`**, breaking the direct coupling to `RunnerState` via a new `UpdateStateProviding` protocol.

---

## Phase 1 — Create the `AppUpdater` target

- [ ] Create `Sources/AppUpdater/` directory
- [ ] Update `Package.swift`:
  - Add `.target(name: "AppUpdater", dependencies: [])` — no `RunBotCore` dependency
  - Add `"AppUpdater"` to `RunBotCore`'s `dependencies` array

---

## Phase 2 — Define `UpdateStateProviding`

- [ ] Create `Sources/AppUpdater/UpdateStateProviding.swift`:

```swift
// Sources/AppUpdater/UpdateStateProviding.swift
public protocol UpdateStateProviding: AnyObject {
    var isDownloading: Bool { get set }
    var updateZipURL: URL? { get set }
    var cachedUpdateVersion: String? { get set }
    var updateActionFailed: Bool { get set }
    func setAvailableUpdate(_ version: String?)
    func rehydrateCachedUpdate(zipURL: URL, version: String)
}
```

- [ ] Mark all requirements `@MainActor` to match existing `RunnerState` isolation contracts
- [ ] Add conformance in `RunBotCore` — no changes needed to existing properties or methods since `RunnerState` already implements all requirements:

```swift
// Sources/RunBotCore/Runner/Models/RunnerState+AppUpdater.swift
import AppUpdater
extension RunnerState: UpdateStateProviding {}
```

---

## Phase 3 — Move files

Move the following from `Sources/RunBotCore/` → `Sources/AppUpdater/`:

| File | Current path |
|---|---|
| `UpdateChecker.swift` | `RunBotCore/Services/` |
| `AutoUpdater.swift` | `RunBotCore/Services/` |
| `AutoUpdater+Install.swift` | `RunBotCore/Services/` |
| `AutoUpdater+BackgroundScheduler.swift` | `RunBotCore/Services/` |
[...]

PR #1792, opened 2026-06-29
feat: Release + Update pipeline (#1774)

eonist:
Closes #1774

Implementation of the release pipeline and in-app update checker, tracked in #1775.

---

## ⚠️ Update Flow — Direction Change (Read This First)

> **The old "Download" banner in Settings → About has been replaced by a full in-app auto-update flow. Reviewers: do not evaluate the old Download link pattern — it no longer exists in this branch.**

### What changed

The original PR shipped a placeholder **Download link banner** in Settings → About (issue #1794 was filed to track the full implementation). That placeholder has now been fully replaced in this branch with a real in-app auto-update flow:

| Before (placeholder) | After (this branch) |
|---|---|
| Banner in Settings → About with a browser Download link | Single row in Settings → About — no banner anywhere |
| User clicks link → browser opens GitHub releases page | App downloads zip silently in background; user taps **Install & Relaunch** |
| Manual download + drag-to-install | App swaps itself and relaunches automatically |

### Spec & tracking

- **High-level spec:** [#1794 — In-app auto-update: download in background, install & relaunch on Update tap](https://github.com/runbot-hq/run-bot/issues/1794)
- **Step-by-step implementation plan:** [#1797 — Auto-update implementation plan: Steps 1–9](https://github.com/runbot-hq/run-bot/issues/1797)

All 9 steps are implemented and complete in this branch.

### Update UI — one place only

```
Settings → About  (only shown when availableUpdate != nil)

  [↓]  Update available: 1.2.3
       A new version of RunBot is ready.
                                    [ Install & Relaunch ]   ← normal path
                                    [ Downloading…      ]   ← while zip downloads
                                    [ Download          ]   ← fallback: asset missing or action failed
```

**No banner. No update UI anywhere outside Settings → About.** This is a hard constraint from the spec (#1794).

### Flow summary

1. On launch / every 24h (60s in DEBUG), `NSBackgroundActivityScheduler` fires `UpdateChecker.checkForUpdate`
2. If a newer release is found with a `RunBot.zip` asset → silent background download begins immediately
3. Zip cached to `~/Library/Caches/io.github.runbot-hq/RunBot-update.zip`; `UserDefaults` persists version + path across force-quits
4. Settings → About row switches from `ProgressView` to **Install & Relaunch** button
5. User taps → `AutoUpdater.installAndRelaunch` extracts zip, replaces `/Applications/RunBot.app`, relaunches, cleans up
[...]

PR #1822, opened 2026-07-01
refactor: extract AppUpdater into standalone library target

eonist:
## Overview

Implements #1821 — moves the auto-update pipeline introduced in #1792 out of `RunBotCore` and into a dedicated `AppUpdater` Swift package library target. The compiler now enforces the dependency direction: `RunBotCore` → `AppUpdater`, never the reverse.

**Stacked on:** #1792 (`feat/release-update-pipeline`)  
**Plan issue:** #1821

> ⚠️ Stacked on #1792. Base is `feat/release-update-pipeline`, not `main`. Review only the delta. Once #1792 merges, retarget to `main` — diff stays clean.

---

## What this PR does

The core change is mechanical: lifts `AutoUpdater` and all associated files out of `RunBotCore/Services/` into `Sources/AppUpdater/`, establishes the SPM target, and wires the dependency edge. Everything else follows from the requirement that the extracted library must have no `RunBotCore` imports — meaning all RunBot-specific identifiers that were hardcoded had to become `init` parameters, and all host-state coupling had to go through a protocol.

**Structural changes:**
- New `AppUpdater` SPM library target; `RunBotCore` gains it as a dependency
- `AutoUpdater` caseless enum → `public @MainActor final class AppUpdater` with injected configuration (`repo`, `currentVersion`, `schedulerIdentifier`, `assetName`, `betaChannelProvider`)
- `UserDefaults` keys scoped to `schedulerIdentifier` — no collisions if another consumer embeds the library

**Forced by the extraction:**
- `UpdateStateProviding` protocol — host UI state had to be abstracted for the library to drive it without importing `RunBotCore`
- `ReleaseProvider` protocol + `GitHubReleaseProvider` — release fetching had to be injectable once the library lost direct access to RunBot's networking layer
- `AppUpdaterTests` target — testability falls out naturally once the injection seams exist

**Test suite (PR A — #1848–#1855):**
- `MockUpdateState` / `MockReleaseProvider` — lightweight in-memory test doubles, no subclassing
- JSON fixtures (`releases.newer.json`, `releases.empty.json`, `releases.beta.json`) — fixture-driven decode tests, no network
- `AppUpdaterFetchTests` — `checkAndHandle` paths, beta wiring, fetch call count (6 tests)
- `AppUpdaterDownloadTests` — nil checksumURL routes to `setAssetMissing()` (step-2 early exit, no Task spawned), `isDownloading` flag, concurrent handle drop (5 tests)
- `UpdateCheckerIsNewerTests` — 19-case semver matrix: major/minor/patch, v-prefix, stable vs beta, beta index ordering, edge cases
- `UpdateCheckerCheckForUpdateTests` — fixture-driven: missing version key, empty list, beta tag, semver guard (5 tests)
- `AppUpdaterDefaultsTests` — key scoping, distinct domains, `clearCachedDefaults`, rehydrate round-trips (7 tests)
- `AppUpdaterChecksumTests` — `verifyChecksum` pass/fail/missing file, `cachedZipDestination` filename/extension/sanitisation/scope (8 tests)

**Code-sign validation (PR B — #1856–#1858):**
- `AppUpdater+CodeSign.swift` — `Bundle.codeSigningIdentity() async -> String?`; `@concurrent` subprocess; `codesign -dvvv` stderr parsing; `#if canImport(AppKit)` gated
- `skipCodeSignValidation: Bool = true` on `AppUpdater` — opt-in identity verification wired into `installAndRelaunch`; defaults to skip (RunBot is ad-hoc signed / SHA-256 only)
- README **Trust model** section — unsigned path (SHA-256 only) and signed path (`codesign -dvvv` identity match), both with code examples and rationale

**Other cleanup carried along:**
- `runCommand` refactored to `@concurrent` free function (Pillar 5 — no new `DispatchQueue` bridges)
- `NSBackgroundActivityScheduler` wrapped in `#if canImport(AppKit)` (SPM headless test runner compatibility)
- Module-local `os.Logger` via fixed subsystem `io.github.appupdater` (not `Bundle.module` — no resources target)
- `Sources/AppUpdater/README.md` — public API, trust model, asset requirements, quick-start

---

## Correctness fixes (caught during review)

**🔴 `clearCachedDefaults()` / `setDownloadStarted()` ordering was inverted** — comment documented `clearCachedDefaults()` first to close a force-quit window; code ran them in the opposite order. Fixed.

**🟡 Cache-hit path: `setAvailableUpdate` fired before `rehydrateCachedUpdate`** — `rehydrateCachedUpdate` clears stale failure flags; calling `setAvailableUpdate` first could render a dirty state for one frame. Fixed to match every other call site.

**🟡 `currentVersion` was `internal`** while `repo` and `schedulerIdentifier` are `public let`. Fixed.

**🔴 `Bundle(path:).codeSigningIdentity()` compile error** — `Bundle(path:)` returns `Bundle?`; calling the method directly on the optional was a type error. Fixed with optional chaining (`Bundle(path:)?.codeSigningIdentity()`).

**🔴 nil `checksumURL` caused spinner flash** — a nil `checksumURL` previously reached `downloadUpdate()`, which threw after `setDownloadStarted()` had already fired, causing a visible spinner flash before settling on the curl-install fallback. Moved the guard to `handle()` step 2: nil `checksumURL` now calls `setAssetMissing()` and returns without entering the download path or touching `isDownloading`.

---

### `@MainActor final class` not caseless enum
The original caseless enum had no injection points — unusable as a library type. A class with `let` configuration is cleanly `Sendable` and aligns with Pillar 1.

### `UpdateStateProviding` — whole-protocol `@MainActor`, compound setters only
Full-protocol `@MainActor` avoids per-method annotations on conforming `@MainActor` classes under Swift 6 strict concurrency. Compound mutation methods (`setDownloadStarted`, `setDownloadComplete`, `setAssetMissing`, `setUpdateFailed`, etc.) prevent TOCTOU misuse — no raw `{ get set }` surface.

### `setAssetMissing()` vs `setUpdateFailed()` — two distinct failure modes
`setAssetMissing()` fires when the release was published without the expected zip asset or `.sha256` sidecar (nil `checksumURL` / missing asset in step 2 of `handle()`). This is the cue for the host to show a curl-install fallback. `setUpdateFailed()` fires for in-flight failures (network error, HTTP non-200, checksum mismatch). These are intentionally separate so the host can show different UI for each.

### `ReleaseProvider` protocol
Thin seam over the GitHub API call. `MockReleaseProvider` in tests; `GitHubReleaseProvider` in production. No URL stubbing required.

### Code-sign validation — opt-in, default skip
RunBot distributes ad-hoc signed via `curl`. Integrity is guaranteed by SHA-256 sidecar (`skipCodeSignValidation: true`, the default). External consumers distributing notarised apps can set `skipCodeSignValidation: false` to additionally verify that the running bundle's `codesign Authority=` identity matches the downloaded bundle's — aborting install on mismatch.

### `NSBackgroundActivityScheduler` P3 exception
Documented exception for `shouldDefer` power-saving. Wrapped in `#if canImport(AppKit)`.

### Force-unwraps in tests — `try #require()` not `!`
All `UserDefaults(suiteName:)!` and `URL(string:)!` force-unwraps in test helpers replaced with `try #require(...)` (Swift Testing idiom). Failures surface as test failures rather than crashes; also satisfies DeepSource SW-W1023.

---

## Known open refinement: `RunBot` direct `AppUpdater` dependency

`RunBot` depends on both `RunBotCore` and `AppUpdater` directly in `Package.swift`. This is a valid SPM diamond and not a correctness issue, but it means the "compiler-enforced boundary" from #1821 applies to `RunBotCore` only — `RunBot` can still import `AppUpdater` symbols directly, bypassing `RunBotCore` as an intermediary.

The root cause is that `AppDelegate` owns `autoUpdater: AppUpdater` concretely, because it needs to call `scheduleBackgroundCheck`, `checkAndHandle`, `rehydrateCachedUpdateIfNewer`, and `installAndRelaunch` — all methods on `AppUpdater` directly. Fully closing this gap would require an `UpdateDriving` protocol covering all four entry points plus `skipCodeSignValidation`, and injecting `any UpdateDriving` into `AppDelegate` and `SettingsView` instead of the concrete type. That is a meaningful refactor beyond the scope of this extraction PR.

**This is tracked as a follow-up.** The extraction is correct and complete as merged; the boundary gap is architectural debt, not a defect.

---

## Concurrency pillar conformance

| Pillar | Status | Note |
|---|---|---|
| P1 Actor-Per-Concern | ✅ | `@MainActor final class` |
| P2 MainActor Boundary | ✅ | `betaChannelProvider` annotated `@MainActor` |
| P3 Structured Timers | ⚠️ Exception | `NSBackgroundActivityScheduler`; `#if canImport(AppKit)` guard |
| P4 Atomic Snapshot | ✅ | Compound setters only |
| P5 `@concurrent` I/O | ✅ | `runCommand` + `codeSigningIdentity` both `@concurrent`; no new `DispatchQueue` bridges |
| P6 Sendable | ✅ | No `@unchecked Sendable` |

---

## Checklist

- [x] `AppUpdater` target builds with zero `RunBotCore` imports
- [x] `AppUpdaterTests` builds with zero `RunBotCore` imports
- [x] `SWIFT_STRICT_CONCURRENCY = complete` — zero warnings
- [x] Full test suite passes
- [x] No `codesign`/`SecCode` calls in any moved file (code-sign check is via subprocess, not Security framework)
- [x] No new `DispatchQueue` bridges introduced
- [x] `deinit { activity.invalidate() }` present on `AppUpdater`
- [x] README covers: quick start, trust model, asset requirements, beta channel, known limitations

---

## Retargeting after #1792 merges

Once `feat/release-update-pipeline` merges to `main`, retarget base to `main`. Diff will show only the extraction — no feature noise from #1792.


eonist:
## AppUpdater library extraction — implementation complete (#1821)

All 8 phases of #1821 are implemented on this branch as incremental commits. Sub-issues #1823–#1830 are closed with their commit SHAs.

### Commits
| Phase | Issue | Commit |
|-------|-------|--------|
| P1 — target + Package.swift | #1823 | `507d437a` (pre-existing) |
| P2–P5 — class/protocol/moved files/scoped defaults + consumer wiring | #1824 #1825 #1826 #1827 | `c1612dc1` |
| P6 — AppUpdaterTests + MockUpdateState + moved tests | #1828 | `5e32900c` |
| P7 — README + docs/AppUpdater.md | #1829 | `de3fd916` |
| P8 — boundary verification + Sendable hardening | #1830 | `3f900b42` |

### What changed
- **New target `Sources/AppUpdater/`** — a standalone, host-agnostic library with **zero RunBotCore dependency** (imports only Foundation/CryptoKit/AppKit/os). `AutoUpdater` (caseless enum) → `@MainActor final class AppUpdater` with all host config injected. Files renamed `AutoUpdater+*` → `AppUpdater+*`.
- **New `UpdateStateProviding` protocol** (`@MainActor`, read-only props + named mutation methods, refined `Sendable`). `RunnerState` conforms via `RunnerState+AppUpdater.swift` and gained `setDownloadStarted/setDownloadComplete/setUpdateFailed`.
- **Scoped `UserDefaults`** via `AppUpdaterDefaults(domain:)`; cache path de-RunBot-ified to `~/Library/Caches/<schedulerIdentifier>/update-<version>.zip`.
- **`runCommand` refactored to a `@concurrent` free function** (no `withCheckedContinuation`/`DispatchQueue`). `NSBackgroundActivityScheduler` retained per-instance, invalidated in `deinit`, guarded by `#if canImport(AppKit)`.
- **Host wiring**: RunBot strings live only at the `AppDelegate` call site; `AppUpdater` injected into `SettingsView` for Install & Relaunch; startup path uses `rehydrateCachedUpdateIfNewer` / `checkAndHandle` / `scheduleBackgroundCheck`.
- **Install path de-RunBot-ified**: locates the first `*.app` in the archive (not a hardcoded `RunBot.app`) and derives the backup name from the running bundle.
- Old `Sources/RunBotCore/Services/AutoUpdater*.swift`, `AutoUpdaterDefaults.swift`, `UpdateChecker.swift` deleted.

### Intentional deviations (from the plan)
1. **Phases 2–5 landed as a single commit** rather than four. They are mutually dependent (the class, its protocol, the moved files, and the scoped defaults must all change together to keep the tree compiling), so splitting them would have produced non-building intermediate commits. Each sub-issue is still closed with this SHA.
2. **`updateZipURL` nil-ing on install failure dropped** — the protocol exposes no `updateZipURL` setter (read-only by design). The UI gates on `updateActionFailed`, which `setUpdateFailed()` sets, so the recovery path is unchanged.
3. **`UpdateStateProviding` refined with `Sendable`** (beyond the plan) so the host-state existential can be captured by the escaping scheduler closure under Swift 6 strict concurrency. Free for conformers (the protocol is wholly `@MainActor`).

### ⚠️ Checks NOT run in this environment
- **`swift build` and `swift test` could NOT be executed** — the sandbox has no Swift toolchain installed. The code is macOS/AppKit-only and cannot be compiled here.
- **CI must verify**: (a) both targets compile under Swift 6 strict concurrency, (b) the moved + new AppUpdaterTests pass, (c) the RunBot app target links against the new AppUpdater product.
- The library-boundary check (zero `import RunBotCore`, no `log()` helper usage in `Sources/AppUpdater`) **was** verified statically via grep.

PR base is unchanged (`feat/release-update-pipeline`); not merged and not retargeted.

Tests/AppUpdaterTests/AppUpdaterBehaviorTests.swift
```diff
@@ -0,0 +1,175 @@
...
+            repo: "example/repo",
+            currentVersion: "1.0.0",
+            assetName: { _ in "RunBot.zip" },
+            schedulerIdentifier: domain,
+            userDefaults: defaults,
+            betaChannelProvider: { false }
+        )
+    }
+
+    private func makeAsset(_ name: String) throws -> ReleaseAsset {
```
eonist, line 27: add doc please

PR #2080, opened 2026-07-14
Fix StatusBarIcon never loading; remove circle fallback

eonist:
Closes #2079

## Root cause

The robot icon PNGs added in #2053 were fine (real binary PNGs, correct dimensions, correct `Contents.json`) — the bug was purely in packaging, in two independent places that both needed fixing:

1. `Package.swift`'s `RunBot` executable target had **no `resources:` entry**, so SwiftPM never compiled `Assets.xcassets` into a resource bundle at all.
2. `build.sh` manually assembles the `.app` (copies the binary + `Info.plist`) and never copied any resource bundle into `Contents/Resources/` — so even a compiled catalog wouldn't have shipped.

`NSImage(named: "StatusBarIcon")` therefore always returned `nil` at runtime, and `menuBarImage(for:)` fell through to the SF Symbol chain, landing on the generic stroked `"circle"` — exactly the symptom reported in the issue.

## Fix

- Moved `Resources/Assets.xcassets` → `Sources/RunBot/Resources/Assets.xcassets` (SwiftPM resource paths must resolve inside the target's own tree, not the package root).
- Declared it in `Package.swift` via `resources: [.process("Resources")]`.
- Added a step in `build.sh` that copies the compiled `RunBot_RunBot.bundle` into `Contents/Resources/`, and fails the build loudly if it's missing instead of silently shipping without it.
- Per the issue's explicit ask: removed the `"circle"` SF Symbol fallback step from `menuBarImage(for:)` entirely. If `StatusBarIcon` is ever missing again, it now falls back directly to the status-appropriate SF Symbol (never a generic circle), and asserts loudly in DEBUG.

## Not yet verified

No macOS/Swift toolchain in this environment, so this hasn't been built. Before merging, please:
- Run `swift build -c release --arch arm64` and confirm `Assets.xcassets` compiles without errors.
- Run `bash build.sh` and confirm `dist/RunBot.app/Contents/Resources/RunBot_RunBot.bundle` exists.
- Launch the built app and confirm the robot head renders in the status bar (light + dark menu bar) instead of the circle.
- Run `swiftlint lint --strict` and `periphery scan` per AGENTS.md.

---
## Summary by cubic
Fixes the status bar icon by correctly bundling the asset catalog and loading the image from `Bundle.module`, caching it to avoid per-tick disk reads. Removes the "circle" fallback so we show the robot icon or a status-appropriate SF Symbol, with a DEBUG assert if the asset is missing.

- **Bug Fixes**
  - Moved `Assets.xcassets` to `Sources/RunBot/Resources` and declared `resources: [.process("Resources")]` in `Package.swift`.
  - Loaded `StatusBarIcon` via `Bundle.module.image(forResource:)`, cached as a `static let`, and set `isTemplate = true`.
  - Updated `build.sh` to copy `RunBot_RunBot.bundle` into `Contents/Resources/` and fail if missing.
  - Simplified fallback: prefer `StatusBarIcon`, else the status SF Symbol; removed `"circle"` and added a DEBUG assert.

<sup>Written for commit 62c3d5ebe99016ab4ae65d4c6f345695aeb7b077. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runbot-hq/run-bot/pull/2080?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved menu bar icon loading so the bundled status icon displays reliably in the built app.
  - Added a safer fallback to the appropriate system icon when the bundled asset is unavailable.
  - Fixed packaging of image resources to prevent missing-icon issues in released applications.
  - Configured the status bar icon as a template image for better appearance across light and dark mode.


eonist:
Solid PR — the root cause diagnosis is correct and the fix is thorough. Here's my full review:

## What's Good ✅

- **Root cause is right.** Both failure points — missing `resources:` in `Package.swift` and no bundle copy in `build.sh` — are correctly identified and fixed. 
- **Asset move is correct.** Relocating `Assets.xcassets` into `Sources/RunBot/Resources/` is exactly what SwiftPM requires; `.process("Resources")` only works within the target's source tree. 
- **`"circle"` fallback removed** as requested in #2079 — the `assertionFailure` in DEBUG is a good replacement, surfaces the real problem instead of masking it. 
- **`build.sh` fails loudly** (`exit 1`) if the bundle is missing — no more silent ship-without-icon builds. 
- **The AI-agent guard comments** in both `Package.swift` and `build.sh` are well-placed and follow the repo's commenting convention. 

## Issues / Concerns ⚠️

**1. `build.sh` still hard-codes `Info.plist` from the old `Resources/` root path**
```bash
cp "Resources/Info.plist" "$OUT_DIR/$APP_NAME.app/Contents/"
```
Now that `Resources/` only contains `Info.plist` (the xcassets were moved out), this still works — but it's worth double-checking nothing else lived in `Resources/` that also needed moving or updating.

**2. Bundle name assumption may be fragile**
```bash
RESOURCE_BUNDLE=".build/arm64-apple-macosx/release/${APP_NAME}_${APP_NAME}.bundle"
```
SwiftPM names the bundle `<PackageName>_<TargetName>.bundle`. If `APP_NAME` is `"RunBot"` and the package name in `Package.swift` matches, this works. But if they ever diverge, the `exit 1` guard will catch it — so it's safe, just worth noting the dependency.

**3. `MenuBarFallback` asset is now fully dead code**
The old fallback chain referenced `NSImage(named: "MenuBarFallback")` — it's been removed from the Swift code, but does the asset still exist in the catalog? If so, it's dead weight. Minor, but worth a cleanup issue.

**4. No CI verification noted**
The PR description explicitly says *"not yet verified, no toolchain available"*. Before merging, the checklist in the PR body should be completed — specifically the `swift build` + `bash build.sh` + visual launch test. Mergeable state is currently `unstable`. 

## Verdict

**Approve with the above notes** — the logic is correct and complete. The only hard blocker is running the manual verification checklist before merge. The `Info.plist` path and bundle name points are worth a quick sanity check but aren't blocking.

PR #2156, opened 2026-07-19
fix: env token (GH_TOKEN) not resolved on GUI app launch from Finder/Dock

eonist:
## Problem

When RunBot is launched from Finder, the Dock, or as a login item, `GH_TOKEN` / `GITHUB_TOKEN` is never found — leaving the app unauthenticated. The same app launched from a terminal works fine.

**Root cause:** macOS GUI apps are spawned by `launchd`, which does not source `~/.zprofile` or `~/.zshrc`. `ProcessInfo.processInfo.environment` therefore never contains `GH_TOKEN` in those launch contexts, even when the token is correctly exported. `TokenCache.resolveFromEnvironment()` reads only `ProcessInfo` and returns `nil` every time.

Tracked in: https://github.com/runbot-hq/run-bot/issues/2155

### `GitHubClient` (companion PR: runbot-hq/GitHubClient#60)

`token()` is now `async`. The resolution chain is a single linear path:

1. In-memory cache — zero I/O, returns immediately on warm cache
2. `TokenStore.load()` — synchronous Keychain read
3. `GH_TOKEN` / `GITHUB_TOKEN` process environment — covers terminal / CI launches
4. Login shell subprocess — cold Finder/Dock/login-item launch only

The shell is spawned lazily on the first step-4 cache miss. Every subsequent call returns from step 1 instantly. There is no eager pre-warming step — the first `token()` call inside `store.start()` is where the ~50–200 ms shell pause occurs on a cold Finder launch.

### `run-bot` (this PR)

`AppState.start()` updated to reflect that `token()` is now `async` and the shell cost lands at `store.start()`. No new startup step was added; the sequence was renumbered and the step-4 comment was updated to document where the shell pause occurs.

## What's unchanged

- `token()` steps 1–3 (cache / Keychain / env) are synchronous and never suspend
- Terminal launches: `resolveFromEnvironment()` resolves first, shell is never spawned
- CI: `ProcessInfo` has the token, shell is never spawned
- Keychain OAuth path: cache is populated in `resolveFromStore()`, shell is never spawned

## Startup sequence (updated)

```
1. seedStoreAndPoller()       — sync; seeds _localRunnerStore + creates RunnerPoller
2. startObservations()        — sync; wires sign-out + status-icon tasks BEFORE any await
3. await refreshAsync()       — first suspension point; hydrates local runners
4. await store.start()        — begins poll loop; first token() call here may suspend
                                ~50–200 ms on a cold Finder/Dock/login-item launch
                                while the login shell sources ~/.zprofile / ~/.zshrc.
                                Terminal / CI / Keychain OAuth paths return immediately.
5. await checkAndHandle()     — launch-time update check
6. scheduleBackgroundCheck()  — periodic update scheduler
```

## Summary by cubic
Fixes missing `GH_TOKEN`/`GITHUB_TOKEN` on GUI launches by resolving the token lazily from a login shell on first use so the first poll is authenticated. CI now clears only the macOS SwiftPM source-checkout cache to avoid stale branch resolution.

- **Bug Fixes**
  - Remove `await github.warmUp()`; the first `TokenCache.token()` call spawns `/bin/zsh -i -l` on Finder/Dock/login-item launches and caches the result. The first poll may pause briefly; terminal/CI/Keychain paths remain instant.
  - Run `startObservations()` after `seedStoreAndPoller()` and before any await to avoid dropping sign-out events. After sign-out, `TokenCache.invalidate()` ensures the next `store.start()` re-resolves the env token.

- **Dependencies**
  - CI: clear only `~/Library/Caches/org.swift.swiftpm` (macOS). Document Linux `~/.cache/org.swift.swiftpm`. Do not clear `~/.swiftpm`, and do not cache the macOS path.

<sup>Written for commit 480701c5be0f646f03d193f3ed176c39a3ba8c9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runbot-hq/run-bot/pull/2156?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">Review in cubic</a>

update. we had to fix hasToken so that settings auth section would show status: https://github.com/runbot-hq/run-bot/issues/2159 and pr https://github.com/runbot-hq/GitHubClient/pull/70

### DIFF
--- a/Sources/RunBot/App/AppState.swift
+++ b/Sources/RunBot/App/AppState.swift
@@ -495,6 +495,13 @@ final class AppState {
     /// `onUpdateStatusIcon` is a callback rather than a direct AppDelegate reference
     /// to avoid AppState holding a strong reference to AppDelegate.
     private func startObservations(onUpdateStatusIcon: @escaping @MainActor () -> Void) {
+        // Ordering tripwire: seedStoreAndPoller() MUST have run before this method.
+        // signOutTask reads self.runnerStore (set by seedStoreAndPoller). If called
+        // out of order, the guard-let inside signOutTask silently drops the first
+        // sign-out event — no crash, no log, just a stalled poll loop after sign-out.
+        // The assert catches a call-order swap in DEBUG/test runs before it ships.
+        assert(runnerStore != nil, "AppState.startObservations: must be called after seedStoreAndPoller() — runnerStore is nil")
+
         // Status icon observation.
         // Wired at Step 2 — BEFORE any suspension point. This is safe because
         // Observations{} has did-set semantics: it emits once immediately with the
@@ -502,10 +509,6 @@ final class AppState {
         // race between here and store.start() are covered by that initial emission.
         // Do NOT move this after any await — see the Step 2 comment
         // in start() for the sign-out window reasoning.
-        // ⚠️ MUST be called after seedStoreAndPoller() — signOutTask accesses
-        // self.runnerStore which is set by seedStoreAndPoller(). Calling this
-        // before seedStoreAndPoller() would make the guard in signOutTask always
-        // fail on the first sign-out event.
         //
         // CAPTURE NOTE: `onUpdateStatusIcon` is captured strongly inside this Task
         // and lives for the process lifetime. AppDelegate must pass a weakly
@@ -528,7 +531,7 @@ final class AppState {
         // Why store.start() is called after sign-out (PR #1138 regression history):
         // Before #1138, polling was driven by a Timer. After sign-out the timer fired,
         // fetch() ran, githubToken() found the keychain cleared, and naturally fell
-        // through to env-token tokens (GH_TOKEN / GITHUB_TOKEN).
+        // through to the env-token fallback (GH_TOKEN / GITHUB_TOKEN).
         // #1138 replaced the timer with a Task that loops on Task.sleep — it never
         // calls start() again on its own, so the env-token fallback only works if
         // start() is explicitly invoked after sign-out. That is what this loop does.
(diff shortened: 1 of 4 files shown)


Sources/RunBot/App/AppState.swift
```diff
@@ -400,21 +408,27 @@ final class AppState {
             return
         }
 
-        // Step 5: start the poll loop.
+        // Step 4: start the poll loop.
+        // On a cold Finder/Dock/login-item launch, the first poll cycle's token()
+        // call suspends here for ~50–200 ms while the login shell sources
```
eonist, line 413: **Why `store.start()` is the correct place for shell-based token resolution** (pre-empting the `warmUp()` false flag)

The open cubic review comment on this line asks: *"Finder/Dock launches still cannot resolve shell-only GH_TOKEN — `store.start()` returns after scheduling its poll task rather than awaiting an async token lookup."*

This is not accurate for the current implementation. Here is why:

`store.start()` begins the poll loop, and the very first `fetch()` call inside that loop calls `github.token()`. `token()` is `async` and follows a four-step resolution chain:

1. In-memory cache — returns immediately if already populated
2. `TokenStore.load()` — synchronous Keychain read
3. `ProcessInfo` environment — covers terminal / CI launches
4. Login shell subprocess (`/bin/zsh -i -l`) — **only** reached on a cold Finder/Dock/login-item launch when steps 1–3 all miss

Step 4 **suspends the caller** until the shell exits (or the 10 s timeout fires). `fetch()` awaits `token()`, so the first poll cycle does not complete — and does not issue any API request — until `token()` returns. The result is then cached in the in-memory store (step 1), so all subsequent poll cycles resolve instantly.

`store.start()` does not "schedule a task and return" independently of token resolution. The poll loop is a structured `async` function; its first iteration awaits `token()` before issuing any request. The shell cost is ~50–200 ms on a typical machine and is paid exactly once per cold launch.

eonist (commented):
Pre-emptive inline defence against the open cubic comment on `store.start()` — no code change.

PR #2370, opened 2026-07-31
fix(LogFetcher): fix ZIP path-stripping bug causing all step logs to show syntheticEmpty [skip ai review]

eonist:
## What

Fixes two bugs in `LogFetcher` that caused every step log lookup to fall through to `.syntheticEmpty` or `.flatBlobFallback` instead of returning the correct `.slice`.

### Bug 1 — Root cause: broken path stripping in `unzipLogsTyped`

`URL(fileURLWithPath: relative).deletingPathExtension().path` treats its argument as an **absolute path**, silently prepending `"/"` to every archive-relative name (e.g. `"/Periphery/2_Checkout"` instead of `"Periphery/2_Checkout"`). The subsequent `hasPrefix("Periphery/")` check therefore **always failed**, so `hasStepFiles` was always `false` and every step was served via the flat-blob fallback or `syntheticEmpty`.

**Fix:** use `NSString.deletingPathExtension` (path-component-safe) and resolve symlinks on both sides of the strip so the `/tmp` → `/private/tmp` alias on macOS does not silently defeat the prefix removal.

### Bug 2 — Missing trim in `sanitizeJobNameForZIP`

The GitHub server (C#) calls `.Trim()` after truncating the job name, and the `gh` CLI calls `strings.TrimSpace` for the same reason. Our function was not trimming, so a job name with a trailing space (exposed after truncation at the 90-char boundary) would produce a sanitised name that never matched the ZIP folder.

**Fix:** call `.trimmingCharacters(in: .whitespaces)` on the return value in both the truncated and non-truncated paths.

## Why it was invisible in tests

The unit tests inject a stub `zipExtractor` that returns pre-built `(name, text)` tuples **without the leading slash**, so the path-stripping bug never manifested in tests. Only the live code path (spawning `/usr/bin/unzip`) produced the broken names.

## Testing

- Existing `FetchStepLogTests` suite passes (stub extractor, no regression).
- Manual verification against a real run ZIP confirms `.slice` is now returned for all matched steps.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job-log extraction and step matching for archives with nested paths or symbolic links.
  * Preserved timestamps, ANSI formatting, warning directives, and expected log filenames during extraction.
  * Improved job-name handling for whitespace, long names, Unicode characters, and surrogate pairs.
  * Enhanced fallback log parsing, including case-insensitive matching, normalized prefixes, completion messages, and duplicate-line handling.
  * Improved cache behavior and graceful handling when system archive tools are unavailable.


eonist (commented):
## Review

This is a well-executed test-layer PR. The core production fix (`/tmp` symlink resolution + `NSString.deletingPathExtension`) is correct and the guard pattern for sandboxed CI is sound. A few issues worth flagging:

---

### `unzipLogsTyped` — self-overlapping path during extraction

The ZIP file is written to `tmp/logs.zip`, and unzip is then told to extract into `-d tmp`. This means the enumerator will find `logs.zip` itself, but `pathExtension == "txt"` filters it out, so no crash. However: any archive containing a file literally named `logs.txt` at the top level would have its name resolved as `"logs"`, colliding with nothing catastrophically — but it is a silent namespace leak that is never documented. More importantly: **the `logs.zip` file is included in the `enumerator` scan**, filtered only by extension. If the archive ever contains a `.txt` file at the same directory level as `logs.zip` (top-level `.txt`), it gets included with name resolution working correctly, which is fine — but it means the `filter { $0.name.contains("/") }` guard in `fetchStepLog` is load-bearing correctness, not just a convenience. That invariant is nowhere documented in the code.

This is worth a comment on the `stepFiles` filter, not a blocker.

---

### `test_cacheHit_zeroAdditionalNetworkCalls` — assertion logic changed silently

The old test simply asserted `rawCallCount == 1` after two fetches. The new test splits it into two assertions:

```swift
#expect(transport.rawCallCount == 1, "ZIP must be fetched exactly once...")
let callsAfterFirst = transport.rawCallCount   // always 1 here
_ = await fetcher.fetchStepLog(...)            // second call
#expect(transport.rawCallCount == callsAfterFirst, "Cache hit must make zero additional network calls")
```

The second assertion is `rawCallCount == callsAfterFirst == 1`, which is equivalent to the original. This is fine functionally. However, `callsAfterFirst` is captured *after* the first assertion already asserted it equals 1, so it is always 1 and the variable adds no signal. If the first assertion ever fails (rawCallCount != 1), the second assertion vacuously still holds because `callsAfterFirst` would equal whatever the actual count was. The test's second assertion is therefore weaker than the comment implies — it does not independently verify "zero additional calls", it verifies "count did not change." No bug, but worth simplifying back to the original pattern.

---

### `CountingTransport` — `@unchecked Sendable` without isolation

```swift
private final class CountingTransport: GitHubTransportProtocol, @unchecked Sendable {
    private(set) var rawCallCount: Int = 0
    ...
    func raw(...) async -> Data? {
        rawCallCount += 1   // ← not actor-isolated, not atomic
```

`rawCallCount` is mutated from an `async` context with no isolation. Since tests call `fetchStepLog` serially and `CountingTransport` is only exercised in a single-threaded test, this won't race in practice. But `@unchecked Sendable` here is a bypass with no comment explaining why it is safe, which is against the usual coding standards visible in the rest of the codebase (other uses of `@unchecked Sendable` tend to have isolation comments). If strict concurrency warnings are enabled, this will bite.

---

### `test_fallback_completeJobEpilogueHeuristic` — fixture uses `"Post job cleanup."` not matching any real log content

The test asserts:
```swift
XCTAssert(result!.contains("Post job cleanup."), ...)
```
But the epilogue was built with:
```swift
epilogue: ["Post job cleanup.", "Cleaning up orphan processes"]
```

The string `"Post job cleanup."` is synthetic and matches only because it was put there by the test itself. There is no issue with the logic — the test correctly exercises the epilogue routing — but the assertion text is misleading: it looks like it is checking for a real GitHub log line when it is just checking that the epilogue was returned and not the section body. A comment clarifying this synthetic nature would help the next reader.

Not a blocker.

---

### `unzipBinaryExists` probe vs. actual spawn-blocked detection

The PR comment correctly notes: "on a sandboxed runner where `/usr/bin/unzip` exists but `Process.run()` is blocked, `unzipBinaryExists` stays true but the result will not be `.slice`. The `!isSlice` arm catches that case."

This is sound for `test_completeJob_regression2358_realExtractor`. But the three tests in `UnzipLogsTests` gate on `!unzipBinaryExists || files.isEmpty`. The `files.isEmpty` check is the spawn-blocked escape hatch there. These two patterns are asymmetric across the two test files but both are correct — they just look inconsistent. A single shared helper (a `canSpawnUnzip()` async function that does a probe spawn, or a shared `let unzipSpawnBlocked` flag) would make the intent uniform.

---

### Deleted comment blocks in `LogFetcher.swift`

This PR removes several well-written `// Cache lookup — single-entry policy` and `// Exclude top-level blob files` comments from `fetchStepLog`. The code is still correct but the cache eviction rationale (why single-entry, why the bound exists, the decompressed-MB concern) is now gone. That rationale existed specifically to explain a non-obvious design decision. The doc comment on `fetchStepLog` does mention caching at a high level but not the eviction policy reasoning. Worth restoring the eviction rationale comment.

---

## Severity Table

| # | File | Finding | Severity |
|---|------|---------|----------|
| 1 | `LogFetcher.swift` | `stepFiles.contains("/")` filter is load-bearing correctness for the `logs.zip` self-overlap; undocumented invariant | 🟡 suggestion |
| 2 | `FetchStepLogTests.swift` | `callsAfterFirst` variable adds no independent signal; second cache assertion is weaker than stated | 🟡 suggestion |
| 3 | `FetchStepLogTests.swift` | `CountingTransport.rawCallCount` mutated in async context under `@unchecked Sendable` with no isolation comment | 🟡 suggestion |
| 4 | `GitHubHelpersTests.swift` | `"Post job cleanup."` assertion looks like a real log line but is synthetic; needs a clarifying comment | 🔵 info |
| 5 | `UnzipLogsTests.swift` vs `FetchStepLogTests.swift` | Asymmetric sandbox-gate patterns (`files.isEmpty` vs `!isSlice`); correct but inconsistent | 🔵 info |
| 6 | `LogFetcher.swift` | Eviction policy rationale comment removed; non-obvious single-entry cache design is now undocumented | 🟡 suggestion |

PR #2803, opened 2026-08-15
feat: wire app sidebar navigation

eonist:
Part of #2793
Stacked on #2798

## What

Adds top-level sidebar navigation to the new SwiftUI app shell.

The sidebar now exposes Workflows, Local runners, Scopes, and Settings. Selecting an item switches the root detail view to the corresponding feature placeholder.

## Changes

- Add `AppSection` as the top-level sidebar route model.
- Store the current selection in `AppShellView`.
- Replace the static sidebar with four selectable rows.
- Route the detail column through `AppDetailView`.
- Add placeholder `MigrationWorkflowView`.
- Add placeholder `MigrationRunnerView`.
- Add placeholder `MigrationScopeView`.
- Add placeholder `MigrationSettingsView`.

> **Note on naming:** Feature-root views use a `Migration` prefix (`MigrationWorkflowView` etc.) to avoid a compile-time conflict with the existing legacy `SettingsView` in `Sources/RunBot/Views/`. The prefix will be dropped when the legacy views are removed in their respective migration steps.

## Out of scope

- Feature data and legacy store integration.
- Workflow/job/step/log columns.
- Runner and scope list/detail layouts.
- Settings controls.
- Sidebar status indicators and counts.
- Metrics.
- Window and selection restoration.

## Acceptance criteria

- [ ] Workflows is selected when the app launches.
- [ ] The sidebar displays Workflows, Local runners, Scopes, and Settings.
- [ ] Every sidebar row has a title and system image.
- [ ] Selecting each row displays its corresponding feature-root view.
- [ ] Selection works with mouse and keyboard navigation.
- [ ] Collapsing and reopening the sidebar preserves the in-memory selection.
- [ ] No feature data, stores, polling, or legacy views are initialized.
- [ ] All changes are contained under `Sources/RunBot/migration/`.
- [ ] `Package.swift`, `build.sh`, `project.yml`, and `main.swift` are unchanged.
- [ ] `swift build` passes ✅
- [ ] SwiftLint passes ✅
- [ ] `swift test` passes (CI).
- [ ] `build.sh` release build passes (CI).

## Summary by Sourcery

Introduce top-level sidebar navigation in the SwiftUI migration app shell and route the detail column to feature placeholder views based on the current selection.

New Features:
- Add AppSection enum as the top-level sidebar route model with titles and icons for Workflows, Local runners, Scopes, and Settings.
- Wire AppShellView to own and persist the current sidebar selection, defaulting to Workflows on launch.
[...]

Issue #2793, opened 2026-08-15
⭐️ Migration plan


eonist:
### migration plan

- We should migrate the app in the same src folder as we use now. 
- We put new code in RunBotApp and RunBotAppCore. 
- We can use code in RunBot and RunBotCore as is
- We should build the RunBotApp now and not the MenuBarApp going forward. So we need to change build.sh etc  
- Each step in the migration plan is one pr. and stack them one by one. so we can test things as we go
- We should not edit a pr after another pr is stacked on top. rather make patch prs on top if fixes are needed etc

1. make the app shell (swiftui window, NavigationSplitView, sidemenu-view, detail-view)
2. wire up the sidemenu items and detail-views (workflow-view, scope-view, runner-view, settings-view)
3. wire up the workflow columns in workflow-view (workflows, jobs, steps, step-log-view)
4. make rows for the workflow, jobs, steps columns
5. wire up scope detail view, reuse rows from menubarapp, reuse add scope sheet from menubarapp etc
6. wire up runner detail view, reuse rows from menubar app, reuse add runner sheet from menubarapp etc
7. wire up settings-detail-view, reuse rows from menubar app
8. wire up metric bottom view in sidebar for CPU, GPU, DISK, MEM
9. wire up window restoation, frame size, last selected view repo, size of columns etc, restore to exact selection combo etc
10. deprecate old menubar app code. deprecat menubarkit etc

Issue #2792, opened 2026-08-15
⭐️ Begin planning migration. Collect ideas on what to migrate where etc


eonist:
see the html files from the prototypes etc. there are some migration tips already

eonist:
plan the views. draw in asci etc.

eonist:
lcal minut counter

eonist:
- another cool thing about the app. ist that you can run it in fullscreen on a startup big screen tv in a stratup office to watch activity etc
- and the app translate easily to ipad app for casting to bigscreen etc, even apple tv app and iphone app is pretty easy retargets the column systems translate ad port easily

eonist:
- State & window restoration. Persist window size/position and selected sidebar item via @SceneStorage/AppStorage so the app reopens where the user left it — something a popover never needed to handl
- NavigationSplitView. SwiftUI's split view maps directly onto the sidebar → list → detail structure and gives free resizing, collapsing, and scroll behavior per column — this is what removes the show-more/less pattern outright.
- NSWindow (SwiftUI WindowGroup/Window)?
- Window restoration. Autosave frame + last selected repo/view; open to previous selection rather than “All runs”.

PR #2806, opened 2026-08-15
feat: add workflow column layout

eonist:
Part of #2793
Stacked on #2803

## What

Replaces the Workflows destination placeholder with the four-pane workflow layout described by the migration plan.

The existing app sidebar remains the first application column. The Workflows detail destination now contains an `HSplitView` with:

1. Workflows
2. Jobs
3. Steps
4. Step log

Each pane is independently resizable and currently displays an empty or selection-dependent placeholder. Rows and feature data remain deferred.

## Changes

- Replace the `MigrationWorkflowView` placeholder with an `HSplitView`.
- Add a shared workflow-column container and header (`MigrationWorkflowColumn`).
- Add the Workflows pane shell (`MigrationWorkflowListView`).
- Add the Jobs pane shell (`MigrationJobListView`).
- Add the Steps pane shell (`MigrationStepListView`).
- Add the Step-log pane shell (`MigrationStepLogView`).
- Give the log pane a larger minimum (320 pt) and ideal width (520 pt) than the list panes (180/220 pt).
- Raise `AppShellView` minimum window width from 720 → 1,120 pt to match the five-column budget (sidebar 220 + three lists 180 each + log 320 + chrome ~40).
- Raise sidebar `navigationSplitViewColumnWidth` min/ideal/max from 180/220/280 → 220/240/300.
- Add `.listStyle(.sidebar)` and `.frame(minWidth: 220)` to `AppSidebarView` for an explicit intrinsic minimum.

## Width budget

| Column | Min |
|---|---|
| Sidebar | 220 pt |
| Workflows | 180 pt |
| Jobs | 180 pt |
| Steps | 180 pt |
| Step log | 320 pt |
| Dividers/chrome | ~40 pt |
| **Total** | **~1,120 pt** |

## Out of scope

- Workflow, job, and step rows.
- Selection state between workflow panes.
- Real feature data or store integration.
- GitHub client and polling setup.
- Status indicators and metadata.
- Log retrieval and rendering.
- Log toolbar controls and badges.
- Column-width restoration.

## Acceptance criteria

- [ ] Selecting Workflows displays four horizontal panes.
- [ ] The panes are titled Workflows, Jobs, Steps, and Step log.
- [ ] Every divider can be resized independently.
- [ ] The Workflows pane displays an empty placeholder.
- [ ] The Jobs pane prompts for a workflow selection.
- [ ] The Steps pane prompts for a job selection.
- [ ] The Step-log pane prompts for a step selection.
- [ ] The log pane receives more default width than each list pane.
- [ ] The default 1,200-point window presents all four panes without overlap.
- [ ] Sidebar never becomes narrower than 220 pt.
- [ ] Sidebar icons and labels remain fully visible.
- [ ] All four workflow panes remain visible at the minimum window width.
- [ ] The window cannot resize below the complete five-column minimum.
- [ ] Switching to another sidebar destination removes the workflow rig.
- [ ] Switching back to Workflows recreates the workflow rig correctly.
- [ ] No hardcoded rows or sample domain data are added.
- [ ] No stores, clients, polling, or authentication are initialized.
- [ ] Divider positions are not persisted.
- [ ] All changes remain under `Sources/RunBot/migration/`.
- [ ] `Package.swift`, `build.sh`, `project.yml`, and `main.swift` are unchanged.
- [ ] `swift build` passes ✅
- [ ] SwiftLint passes ✅
- [ ] `swift test` passes (CI).
- [ ] `build.sh` release build passes (CI).

### DIFF
--- a/Sources/RunBot/migration/AppShell/AppShellView.swift
+++ b/Sources/RunBot/migration/AppShell/AppShellView.swift
@@ -15,13 +15,13 @@ struct AppShellView: View {
             AppSidebarView(selection: $selection)
                 .navigationSplitViewColumnWidth(
                     min: 180,
-                    ideal: 220,
-                    max: 280
+                    ideal: 210,
+                    max: 260
                 )
         } detail: {
             AppDetailView(selection: selection)
         }
         .navigationSplitViewStyle(.balanced)
-        .frame(minHeight: 480)
+        .frame(minWidth: 920, minHeight: 480)
     }
 }
--- a/Sources/RunBot/migration/AppShell/AppShellView.swift
+++ b/Sources/RunBot/migration/AppShell/AppShellView.swift
@@ -14,14 +14,14 @@ struct AppShellView: View {
         NavigationSplitView {
             AppSidebarView(selection: $selection)
                 .navigationSplitViewColumnWidth(
-                    min: 180,
-                    ideal: 220,
-                    max: 280
+                    min: 220,
+                    ideal: 240,
+                    max: 300
                 )
         } detail: {
             AppDetailView(selection: selection)
         }
         .navigationSplitViewStyle(.balanced)
-        .frame(minWidth: 720, minHeight: 480)
+        .frame(minWidth: 1_120, minHeight: 480)
     }
 }
--- a/Sources/RunBot/migration/AppShell/AppShellView.swift
+++ b/Sources/RunBot/migration/AppShell/AppShellView.swift
@@ -22,6 +22,6 @@ struct AppShellView: View {
             AppDetailView(selection: selection)
         }
         .navigationSplitViewStyle(.balanced)
-        .frame(minWidth: 920, minHeight: 480)
+        .frame(minWidth: 720, minHeight: 480)
     }
 }
--- a/Sources/RunBot/migration/AppShell/AppShellView.swift
+++ b/Sources/RunBot/migration/AppShell/AppShellView.swift
@@ -14,14 +14,14 @@ struct AppShellView: View {
         NavigationSplitView {
             AppSidebarView(selection: $selection)
                 .navigationSplitViewColumnWidth(
-                    min: 220,
-                    ideal: 240,
-                    max: 300
+                    min: 180,
+                    ideal: 220,
+                    max: 280
                 )
         } detail: {
             AppDetailView(selection: selection)
         }
         .navigationSplitViewStyle(.balanced)
-        .frame(minWidth: 1_120, minHeight: 480)
+        .frame(minHeight: 480)
     }
 }
--- a/Sources/RunBot/migration/AppShell/AppSidebarView.swift
+++ b/Sources/RunBot/migration/AppShell/AppSidebarView.swift
@@ -15,6 +15,8 @@ struct AppSidebarView: View {
             Label(section.title, systemImage: section.systemImage)
                 .tag(section)
         }
+        .listStyle(.sidebar)
         .navigationTitle("RunBot")
+        .frame(minWidth: 220)
     }
 }
--- a/Sources/RunBot/migration/AppShell/AppSidebarView.swift
+++ b/Sources/RunBot/migration/AppShell/AppSidebarView.swift
@@ -17,6 +17,5 @@ struct AppSidebarView: View {
         }
         .listStyle(.sidebar)
         .navigationTitle("RunBot")
-        .frame(minWidth: 220)
     }
 }
--- a/Sources/RunBot/migration/Features/Workflows/Columns/MigrationColumnPlaceholder.swift
+++ b/Sources/RunBot/migration/Features/Workflows/Columns/MigrationColumnPlaceholder.swift
@@ -1,0 +1,40 @@
+// MigrationColumnPlaceholder.swift
+// RunBot
+
+import SwiftUI
+
+/// Width-adaptive placeholder for workflow column panes.
+/// Uses `minimumScaleFactor` so text stays visible in narrow columns.
+struct MigrationColumnPlaceholder: View {
+    /// Primary label shown below the icon.
+    let title: String
+    /// SF Symbol name for the placeholder icon.
+    let systemImage: String
+    /// Optional secondary description shown below the title.
+    var description: String?
+
+    /// The placeholder layout.
+    var body: some View {
+        VStack(spacing: 12) {
+            Image(systemName: systemImage)
+                .font(.system(size: 36))
+                .foregroundStyle(.secondary)
+
+            Text(title)
+                .font(.title2.weight(.semibold))
+                .multilineTextAlignment(.center)
+                .lineLimit(2)
+                .minimumScaleFactor(0.8)
+
+            if let description {
+                Text(description)
+                    .font(.callout)
+                    .multilineTextAlignment(.center)
+                    .foregroundStyle(.tertiary)
+                    .lineLimit(3)
+            }
+        }
+        .padding(.horizontal, 16)
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+    }
+}
--- a/Sources/RunBot/migration/Features/Workflows/Columns/MigrationJobListView.swift
+++ b/Sources/RunBot/migration/Features/Workflows/Columns/MigrationJobListView.swift
@@ -8,8 +8,8 @@ struct MigrationJobListView: View {
     /// The pane content.
     var body: some View {
         MigrationWorkflowColumn(title: "Jobs") {
-            ContentUnavailableView(
-                "Select a workflow",
+            MigrationColumnPlaceholder(
+                title: "Select a workflow",
                 systemImage: "list.bullet.rectangle"
             )
         }
--- a/Sources/RunBot/migration/Features/Workflows/Columns/MigrationJobListView.swift
+++ b/Sources/RunBot/migration/Features/Workflows/Columns/MigrationJobListView.swift
@@ -1,0 +1,17 @@
+// MigrationJobListView.swift
+// RunBot
+
+import SwiftUI
+
+/// Jobs pane shell. Prompts for a workflow selection until job rows are added.
+struct MigrationJobListView: View {
+    /// The pane content.
+    var body: some View {
+        MigrationWorkflowColumn(title: "Jobs") {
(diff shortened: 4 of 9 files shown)


eonist:
Structure is right — the container abstraction is the correct call, and it's the shape we discussed (2-column split, `HSplitView` in detail). Three things to fix, one of which is arithmetic.

## Missing layout priority

None of the four panes has `layoutPriority`, and none has a `maxWidth`. That means `HSplitView` treats all four as equally flexible, so window growth won't preferentially feed the log pane — extra width gets spread across all four and the three list panes drift wider than you want. Add `.layoutPriority(1)` to `MigrationStepLogView` and cap the three list panes with `maxWidth` (say 320) so they stay list-shaped.

## The widths don't fit the window

Summing what the diff declares: minimums are 180 + 180 + 180 + 320 = 860, plus the sidebar's 180 minimum and three dividers puts the content floor above 1040. But `AppShellView` still declares `.frame(minWidth: 720, minHeight: 480)`, so the window can be dragged to a width where the panes cannot honor their own minimums.

Ideals are worse: 220 + 220 + 220 + 520 = 1180 for the panes, plus a 220 sidebar ideal is 1400, against step 1's `defaultSize(width: 1_200)`. So at launch nothing gets its ideal width — the log pane opens roughly 200pt short. Either raise `defaultSize` to about 1440, trim the ideals, or switch the scene to `.windowResizability(.contentMinSize)` so the window minimum derives from the content instead of a stale 720.

## navigationTitle regression

`MigrationWorkflowView` lost its `.navigationTitle("Workflows")` in this diff, while the runners, scopes, and settings roots still set theirs. So the window title now changes shape depending on destination — "Scopes" or "Settings" for three of them, and whatever the sidebar contributes for Workflows. Put it back on the `HSplitView`, or drop it from all four and set one title on the split view. Either is fine; the mix isn't.

## Make the container generic now

`MigrationWorkflowColumn` is already generic in everything but its name, and steps 5 and 6 need the same header for the runners and scopes columns — where #2792 puts an "Add local runner" and "Add scope" button in the column header. As written there's no trailing accessory slot, so those steps will either duplicate the container or retrofit it.

Cheap now, annoying later: rename it `MigrationColumn` and add an optional trailing `@ViewBuilder` accessory that renders next to the `Spacer()`. Zero behavior change for the four current panes.

## Minor

`ContentUnavailableView` with a description sentence inside a 220pt-wide pane will wrap hard and look cramped — it's designed for full-width empty states. For the three narrow panes a centered secondary `Text` would render better, though it's cosmetic and step 4 replaces them with rows anyway.

I read the diff only, so I haven't verified the CI checks on #2806 — your local `swift build` and SwiftLint results are the only signal I'm going on for those.

eonist:
Here is the technical code review for **[PR #2806](https://github.com/runbot-hq/run-bot/pull/2806)** (*feat: add workflow column layout*), evaluated against the requirements of Step 3 in migration issue #2793.

***

### Scope & Context
This PR replaces the single placeholder view in `MigrationWorkflowView` with an `HSplitView` hosting four discrete panes: Workflows, Jobs, Steps, and Step Log, each wrapped in a standardized container (`MigrationWorkflowColumn`).

***

### Architecture & Correctness Review

#### 1. Inconsistent Window Title Behavior (`navigationTitle` dropped)
In Step 2, `MigrationWorkflowView` declared `.navigationTitle("Workflows")`. In this PR, `MigrationWorkflowView` removed the modifier, whereas sibling detail views (`MigrationScopeView`, `MigrationRunnerView`, `MigrationSettingsView`) still retain `.navigationTitle(...)`.
- **Impact:** Switching between sidebar destinations causes inconsistent macOS window title behavior (title clears or sticks to the previous state when switching to Workflows).
- **Fix:** Add `.navigationTitle("Workflows")` directly to the root `HSplitView` in `MigrationWorkflowView`.

#### 2. Frame Constraints vs Window Bounds Mismatch
The declared dimensions across the columns introduce layout pressure against `AppShellView` and `RunBotDesktopApp`:
- Total min width inside `HSplitView`: `180 * 3 + 320 = 860 pt`. Adding the sidebar (`180 pt` min) yields a minimum content floor of `~1,040 pt` (excluding split dividers).
- `AppShellView` allows the window to shrink to `720 pt` (`.frame(minWidth: 720)`).
- Ideal widths sum to `220 * 3 + 520 = 1,180 pt` (`1,400 pt` including sidebar), exceeding the `1,200 pt` default window size set in `RunBotDesktopApp`.
- **Recommendation:** Adjust `AppShellView` minimum width (or use `.windowResizability(.contentMinSize)`) and ensure default window width accommodates the four panes without clipping the log view at launch.

#### 3. Unconstrained Resize Flexibility & Layout Priority
In `HSplitView`, panes without `.layoutPriority` or `maxWidth` expand proportionally when resizing the window.
- **Impact:** Resizing the window wider expands the list panes (Workflows, Jobs, Steps) excessively rather than granting the extra horizontal space to `MigrationStepLogView`.
- **Fix:** Provide `.layoutPriority(1)` on `MigrationStepLogView` and set sensible `maxWidth` caps (e.g. `maxWidth: 320` or `360`) on the three list columns so extra viewport width feeds into the log pane.

#### 4. Shared Column Container Scalability
`MigrationWorkflowColumn` is cleanly decoupled via `@ViewBuilder content`, but will immediately need header trailing actions (e.g., toolbar items or action buttons) in upcoming steps (Scopes and Runners in Steps 5 & 6).
- **Suggestion:** Consider generalizing the container or adding an optional `trailingAction` closure now to prevent refactoring or duplicating column containers in consecutive PRs.

***

### Findings Severity

| Severity | Item | File(s) | Description |
| :--- | :--- | :--- | :--- |
| 🔴 **Blocker** | Navigation title regression | `MigrationWorkflowView.swift` | `.navigationTitle("Workflows")` was removed, breaking destination title consistency with sibling views. |
| 🟡 **Suggestion** | Layout priority & `maxWidth` | `MigrationWorkflowView.swift` | Add `.layoutPriority(1)` to `MigrationStepLogView` and `maxWidth` bounds on list columns to keep lists compact during resize. |
| 🟡 **Suggestion** | Window min & default bounds | `AppShellView.swift` / `RunBotDesktopApp.swift` | Content floor (`>1040 pt`) exceeds window minimum (`720 pt`), and combined ideal widths (`1400 pt`) exceed `1200 pt` initial window size. |
| 🔵 **Info** | Column container extensibility | `MigrationWorkflowColumn.swift` | Adding an optional header trailing slot will allow direct reuse for Scopes and Runners in Steps 5–6. |

eonist:
## Review — PR #2806 (feat: add workflow column layout)

Scope check first: the PR does exactly what step 3 of the migration plan asks for — wire up the workflow columns (workflows, jobs, steps, step-log) inside the Workflows detail view, nothing more . No stores, clients, or sample data crept in, and all files stay under `Sources/RunBot/migration/` as the acceptance criteria demand. Two things are worth acting on before merge.

## Window minimum conflicts with pane minimums

`AppShellView` still declares `.frame(minWidth: 720, minHeight: 480)`, and the sidebar column is `navigationSplitViewColumnWidth(min: 180, ideal: 220, max: 280)` (both from the stacked base #2803). This PR now puts four hard minimums into the detail column: 180 + 180 + 180 + 320 = **860 pt**, before divider thickness. With the sidebar's 180 pt minimum, the layout cannot render below roughly 1,040–1,050 pt, yet the window can still be dragged to 720 pt.

The acceptance criterion only checks the 1,200 pt default case, so this passes the checklist while still being reachable by any user who resizes down or restores a small saved frame. At that size AppKit resolves the conflict by clipping/squeezing the trailing pane — the step-log column, i.e. the one that matters most. Either raise the shell's `minWidth` to cover the sum of the pane minimums plus the sidebar, or drop the list-pane minimums so the sum fits inside 720 pt.

## Lost navigation title on the Workflows destination

The old placeholder ended with `.navigationTitle("Workflows")`; the new `HSplitView` body has no `navigationTitle` at all. Its three sibling destinations — `MigrationRunnerView`, `MigrationScopeView`, `MigrationSettingsView` — all still set theirs (added in #2803).

Because `AppDetailView` is a plain `switch` on `AppSection` rather than a `NavigationStack` per destination, the detail column's title is whatever the currently rendered view sets. Selecting Settings and then Workflows leaves the window/toolbar title inconsistent with the selection, and Workflows is the launch default (`@State private var selection: AppSection? = .workflows`), so it is the first screen a user sees. Adding `.navigationTitle("Workflows")` to the `HSplitView` restores parity with the other three routes.

## Verified as fine

- `MigrationWorkflowColumn` is a clean generic container: `@ViewBuilder` init, content stored eagerly (fine — it's a value-type view body, not a lazy container), and `VStack(spacing: 0)` + `Divider()` gives the intended header/content split with no layout ambiguity.
- Pane shells are pure and stateless; no selection state leaked in, matching the "out of scope" list.
- SF Symbols used (`list.bullet.rectangle`, `checklist`, `doc.plaintext`, `bolt.horizontal.circle`) are all valid and the workflows symbol matches `AppSection.workflows.systemImage`, so sidebar and pane stay visually consistent.
- `ContentUnavailableView` is macOS 14+, but it was already the pattern in #2803, so this PR introduces no new availability floor.
- Doc comments are present on every type, property, and `body` — consistent with the surrounding migration code.

## Tests

The repo has `RunBotTests` and `RunBotUITests` targets, and this PR adds nothing to either. That matches #2803, which also shipped view shells untested, so I'd treat it as a tracked gap for the step where rows and selection state arrive (step 4) rather than a merge blocker — there's no logic here to assert beyond "the view compiles."

## Severity

|---|---|---|
| 🟡 | Pane minimums (860 pt) exceed what the shell's 720 pt `minWidth` can host with the 180 pt sidebar; trailing log pane clips on resize down | `MigrationWorkflowView.swift` + `AppShellView.swift` |
| 🟡 | `.navigationTitle("Workflows")` dropped; the other three destinations still set theirs, and Workflows is the launch default | `MigrationWorkflowView.swift` |
| 🔵 | Whether `idealWidth: 520` actually wins the initial `HSplitView` distribution is worth a visual check against the "log pane gets more default width" criterion | `MigrationWorkflowView.swift` |
| 🔵 | No tests for the new shells; defer to step 4 when rows and selection introduce real behavior | `Tests/RunBotTests` |

PR #2811, opened 2026-08-15
feat: add workflow hierarchy rows

eonist:
Part of #2793 · Stacked on #2806

## What

Implements the full three-level selection model and row components for
the four-pane workflow layout introduced in step 3.

### Selection
- `MigrationWorkflowSelection` — `@Observable` class owning
  `workflowID: String?`, `jobID: Int?`, `stepNumber: Int?`.
  `selectWorkflow/Job/Step` each clear downstream state.
  `reconcile(workflows:)` removes stale IDs after a data refresh.

### Row components (`Rows/`)
- `MigrationStatusIndicator` — 10pt circle filled with `status.color`
- `MigrationRowMetadata` — joins non-nil values with ` · `, collapses empty fields
- `MigrationWorkflowRow` — repo short name, title, branch, elapsed, job progress
- `MigrationJobRow` — display title, `#id`, elapsed, step progress count
- `MigrationStepRow` — name + elapsed; elapsed omitted when `startDate == nil` (queued guard)

### Columns wired (`Columns/`)
- `MigrationWorkflowListView` — `List` bound to `selection.workflowID`
- `MigrationJobListView` — bound to `selection.jobID`; shows "Select a workflow" placeholder
- `MigrationStepListView` — `List(id: \.number)` because `GitHubStep` is not `Identifiable`
- `MigrationStepLogView` — shows `step.name` + deferred message; placeholder when nil

### Root view
- `MigrationWorkflowView` now owns `@State private var selection` and
  derives `selectedWorkflow/Job/Step` for passing to columns.
  `AppDetailView` updated to `MigrationWorkflowView(workflows: [])`.

## Tests (16 new)
- `MigrationWorkflowSelectionTests` — 6 cases: clear-downstream, reconcile missing workflow/job/step, keep valid chain
- `MigrationWorkflowRowFormattingTests` — job progress empty vs count
- `MigrationJobRowFormattingTests` — step progress empty vs count
- `MigrationStepRowFormattingTests` — elapsed omitted/present based on `startDate`

## Not in this PR
Log rendering (ANSI/Markdown, toolbar, badges) — deferred to step 5.
Production store wiring — deferred to step 5.

## Checklist
- [x] `swift build` clean
- [x] SwiftLint — warnings only, no errors (`LINT:0`)
- [x] All new tests passing

## Summary by Sourcery

Introduce hierarchical in-memory workflow selection and hook it up to the four-pane workflows UI, rendering workflow, job, and step rows with basic step-log confirmation.

New Features:
- Add MigrationWorkflowSelection observable model to track workflow, job, and step hierarchy using stable identifiers and reconciliation against refreshed data.
- Implement workflow, job, and step row components with status indicators and compact metadata lines for elapsed time and progress.
- Wire workflow, job, step list columns and step-log pane to real data and selection state, including contextual placeholders when no selection or data is available.

Enhancements:
- Update MigrationWorkflowView to own selection state, derive the currently selected workflow, job, and step, and pass them into the column views.
- Adjust AppDetailView to construct MigrationWorkflowView with an explicit workflows array placeholder ahead of production store wiring.

Tests:
- Add MigrationWorkflowSelectionTests covering downstream clearing and reconciliation of missing or valid workflow, job, and step selections.
- Add formatting tests for workflow, job, and step rows to validate job progress and elapsed-time display behavior, including guards for empty or queued items.

---
## Summary by cubic
Implements three-level workflow → job → step selection and renders rows in the four-pane workflow view for end‑to‑end navigation. Previously panes showed placeholders; now lists bind to selection, rows show status and metadata, and the step-log pane confirms the selected step. Aligns with Linear #2793 step 4.

- Adds MigrationWorkflowSelection (workflowID, jobID, stepNumber); selecting workflow/job clears downstream; reconcile removes stale IDs after refresh.
- Wires columns to selection: MigrationWorkflowListView, MigrationJobListView, MigrationStepListView (steps use id: \.number) with contextual placeholders.
- Adds rows: MigrationWorkflowRow, MigrationJobRow, MigrationStepRow, plus MigrationStatusIndicator and MigrationRowMetadata; omits elapsed for unstarted steps; shows job/step progress when present.
- Bridges statuses for indicators: `GroupStatus.rbStatus` and `GitHubStep.rbStatus`; adds missing `RunBotCore` and `GitHubClient` imports.
- Updates MigrationWorkflowView to accept `workflows` and own selection; derives selected items; MigrationStepLogView shows the selected step name only; AppDetailView now constructs MigrationWorkflowView(workflows: []).
- Tests: keep a single smoke test for selection; remove premature formatting tests.

<sup>Written for commit 5a60760e691607dd1d4f481663aebd788345f795. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runbot-hq/run-bot/pull/2811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


eonist:
Solid slice — the selection chain with downstream clearing is exactly right, and the `.lineLimit(1)` discipline on every row title directly addresses the intrinsic-width trap from the sizing work. `MigrationRowMetadata` filtering nil and empty values before joining is a nice touch; that's where stray "·" separators usually come from.

Four things, one of which is a real defect.

## Two of the three test files don't test the code

`MigrationJobRowFormattingTests` reimplements the row's logic in the test body:

```swift
let done = job.steps.filter { $0.conclusion != nil }.count
#expect(done == 1)
```

That asserts the fixture is what you built it to be. `MigrationJobRow.stepProgress` is private and never called, so it could return `"0/0 steps"` for every job and these tests still pass. `MigrationStepRowFormattingTests` has the same shape — it checks `step.startDate` and `step.elapsed(now:)` rather than the row's `elapsedText`.

Note the API mismatch that exposes it: the test calls `step.elapsed(now:)` while `MigrationStepRow` uses the `step.elapsed` property. Different code paths entirely.

Two honest options: extract the formatting into internal helpers the tests can call, or delete those two files. `MigrationWorkflowSelectionTests` is the opposite case — it drives the real API and the reconcile cases are genuinely valuable. Keep that one.

## reconcile has no caller

`reconcile(workflows:)` is well designed and thoroughly tested, and nothing invokes it. When step 5 wires the store, whoever does it has to remember. Wire it now while it's a no-op:

```swift
.onChange(of: workflows) { selection.reconcile(workflows: workflows) }
```

Three lines on the `HSplitView`, zero behavior change with `workflows: []`, and the guarantee becomes real instead of aspirational. Periphery may also flag it as unused otherwise.

## Spec deviation on status indicators

`MigrationStatusIndicator` renders a plain filled `Circle` for all three row types. #2792 specifies a donut graph for workflow and job rows and status icons — check, cross — for step rows. Fine as a step 4 placeholder, but it should be logged rather than discovered at step 10, since the donut is doing real work in that design (progress, not just state).

Related and cheap: status is conveyed by color alone with no accessibility label. `.accessibilityLabel` on the indicator costs one line and matters here because status is the app's primary signal.

## Smaller notes

- `.tag(job.id)` and `.tag(workflow.id)` are still redundant — `List(data, selection:)` over `Identifiable` data tags rows by `id` already. `List(steps, id: \.number, ...)` genuinely needs its explicit `id`, so that one is correct.
- `AppDetailView` hardcodes `MigrationWorkflowView(workflows: [])`, so there's no way to see a populated column. Since you don't want more tests, a `#Preview` with the fixtures you already built for the selection tests would let you eyeball rows in Xcode previews without adding CI surface.
- `WorkflowActionGroup` is keyed on `headSha` and deliberately groups sibling runs, so the `String?` workflow ID is the right choice. One edge case for step 5: if two scopes track a fork and its upstream, the same SHA can appear twice and the `List` would see duplicate ids.

PR #2820, opened 2026-08-15
feat: migrate application settings

eonist:
Part of #2793
Stacked on #2817

## What

Replaces the Settings placeholder with a scrollable, single-column settings detail that reuses the existing authentication, general, update, and about rows from the legacy menu-bar settings view.

## Changes

- Replace `MigrationSettingsView` placeholder with a scrollable settings detail.
- Reuse the existing `AuthenticationSection` and authentication cards.
- Extract reusable general-settings rows from `SettingsView+Sections`.
- Extract reusable update-settings rows from `SettingsView+Sections`.
- Extract reusable About/version content.
- Wire launch-at-login through the existing `LoginItem` behaviour.
- Wire update and beta-channel controls through the existing updater/preferences.
- Use the existing root `GitHubAuthentication` instance.
- Extend `MigrationAppDependencies` only for required settings services.
- Omit legacy panel navigation to Scopes and Local runners.

## Out of scope

- Workflow production data and step-log rendering.
- Scope and runner changes.
- Redesigning settings rows.
- New UI, snapshot, or formatting tests.

## Summary by Sourcery

Introduce a full settings experience in the migrated desktop app, replacing the placeholder view with a scrollable, single-column layout wired to existing authentication, general, update, and about services.

New Features:
- Add a windowed Settings destination that reuses existing authentication, general, update, and about sections in a constrained scrollable layout.

Enhancements:
- Thread a shared GitHubAuthentication instance and settings-related services from the app composition root into the migration shell to avoid duplicate service ownership.
- Extract reusable General, Update, and About settings sections from legacy settings code for use in the migration UI.
- Configure an AppUpdater instance in the migration dependencies to drive automatic updates, beta-channel preferences, and in-app update actions.
- Wire launch-at-login behavior through the shared LoginItem helper for the migrated Settings view.

Chores:
- Adjust MigrationAppDependencies and RunBotDesktopApp initialization to construct and pass MigrationSettingsDependencies through the shell views to MigrationSettingsView.

---
## Summary by cubic
Replaces the Settings placeholder with a functional, single‑column, scrollable settings view. Previously Settings was a placeholder; now it renders authentication, general (launch at login), updates (automatic updates + beta with updater actions), and about (version). Adds doc comments to satisfy SwiftLint missing_docs without changing behavior.

- Review notes
[...]

PR #2822, opened 2026-08-16
feat: add sidebar system metrics

eonist:
Part of #2793
Stacked on #2820

## What

Adds a live system-metrics overview pinned to the bottom of the application sidebar.

The footer displays CPU and GPU usage with compact history graphs, plus memory and disk used/capacity/available values.

## Changes

- Split the sidebar into a scrollable navigation list and pinned metrics footer.
- Add `SidebarMetricsView`.
- Reuse `SystemStatsViewModel` as the single live sampler.
- Reuse `SystemStats` and `SparklineView`.
- Add compact CPU and GPU usage rows.
- Add compact memory and disk capacity rows.
- Start and stop sampling with the sidebar metrics lifecycle.
- Handle unavailable GPU telemetry without displaying a fabricated value.
- Add textual accessibility labels for every metric.

## Out of scope

- Workflow production data and step-log rendering.
- Runner telemetry and per-runner CPU/memory.
- Metrics preferences or configuration.
- Persisting metric history.
- Changing sampler cadence or platform APIs.
- Window/sidebar restoration.
- Native sheet migration.
- MenuBarKit removal.
- New UI, snapshot, formatter, or sampling tests.

## Acceptance criteria

- [ ] Metrics are pinned below the sidebar navigation list.
- [ ] Navigation scrolls independently from the metrics footer.
- [ ] CPU shows a live percentage and rolling history.
- [ ] GPU shows a live percentage and rolling history when available.
- [ ] Unavailable GPU telemetry displays `u2014`.
- [ ] Memory shows used, capacity, and available values.
- [ ] Disk shows used, capacity, and available values.
- [ ] Changing sidebar selection does not restart sampling.
- [ ] The sampler starts once when the sidebar appears.
- [ ] The sampler stops when the sidebar disappears.
- [ ] Metrics fit at the existing sidebar minimum width.
- [ ] Metrics remain readable with large byte values.
- [ ] Percentage and capacity values use monospaced digits.
- [ ] Every metric exposes a complete accessibility label.
- [ ] No application minimum-width increase is required.
- [ ] Existing navigation behavior remains unchanged.
- [ ] Existing sampler logic and cadence remain unchanged.
- [ ] No premature UI or formatting tests are added.
- [ ] `swift build` passes.
- [ ] `swift test` passes.
- [ ] SwiftLint passes.
- [ ] `build.sh` release build passes.

## Summary by Sourcery

Introduce a pinned system metrics footer to the application sidebar, refactoring the sidebar layout to keep navigation scrollable while displaying live CPU, GPU, memory, and disk information with accessible, compact metric rows whose sampling lifecycle is scoped to the sidebar view.

New Features:
- Add a pinned sidebar metrics footer showing live CPU, GPU, memory, and disk usage.

Enhancements:
- Refactor the sidebar layout into a scrollable navigation list with a fixed footer section.
- Tie sidebar metrics sampling to the sidebar view lifecycle using a dedicated SystemStatsViewModel instance.
- Improve handling of unavailable GPU telemetry by omitting sparkline history and displaying an em-dash instead of fabricated values.
- Provide compact, readable metric rows with monospaced numeric formatting and full accessibility labels for all metrics.

---
## Summary by cubic
Adds a live system-metrics footer pinned to the bottom of the app sidebar, so users can see CPU, GPU, memory, and disk at a glance. Previously the sidebar only showed navigation; now navigation scrolls independently while metrics stay fixed, and sampling starts/stops with the sidebar lifecycle.

- Refactors `AppSidebarView` to a `VStack` with a scrollable `List`, a `Divider`, and a pinned `SidebarMetricsView`.
- Introduces `SidebarMetricsView`, `SidebarUsageMetricView`, and `SidebarCapacityMetricView`; reuses `SystemStatsViewModel`, `SystemStats`, and `SparklineView`.
- Starts the sampler on appear and stops on disappear; navigation changes do not restart sampling.
- Handles unavailable GPU telemetry by showing an em-dash and omitting the sparkline.
- Adds complete accessibility labels; uses monospaced digits; layout fits the existing sidebar minimum width and does not change sampler cadence.

<sup>Written for commit adb73f67f311b30c9e8db8ac7ad173afb0b375f4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runbot-hq/run-bot/pull/2822?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/RunBot/migration/AppShell/AppSidebarView.swift
+++ b/Sources/RunBot/migration/AppShell/AppSidebarView.swift
@@ -3,19 +3,29 @@
 
 import SwiftUI
 
-/// Sidebar column with selectable navigation rows.
-/// Status indicators and counts are added in later migration steps.
+/// Sidebar column composed of a scrollable navigation list and a pinned metrics footer.
+///
+/// The navigation `List` and `SidebarMetricsView` are separated by a `Divider`
+/// inside a `VStack` so metrics stay fixed at the bottom regardless of selection.
+/// Navigation can shrink and scroll independently when the window is short.
 struct AppSidebarView: View {
     /// Binding to the shell's current section selection.
     @Binding var selection: AppSection?
 
-    /// The selectable section list.
+    /// The sidebar layout: scrollable navigation list above a pinned metrics footer.
     var body: some View {
-        List(AppSection.allCases, selection: $selection) { section in
-            Label(section.title, systemImage: section.systemImage)
-                .tag(section)
+        VStack(spacing: 0) {
+            List(AppSection.allCases, selection: $selection) { section in
+                Label(section.title, systemImage: section.systemImage)
+                    .tag(section)
+            }
+            .listStyle(.sidebar)
+
+            Divider()
+
+            SidebarMetricsView()
+                .padding(10)
         }
-        .listStyle(.sidebar)
         .navigationTitle("RunBot")
     }
 }
--- a/Sources/RunBot/migration/AppShell/Metrics/SidebarCapacityMetricView.swift
+++ b/Sources/RunBot/migration/AppShell/Metrics/SidebarCapacityMetricView.swift
@@ -1,0 +1,54 @@
+// SidebarCapacityMetricView.swift
+// RunBot
+
+import SwiftUI
+
+// MARK: - SidebarCapacityMetricView
+
+/// Compact two-line metric row showing used / total GB and available GB.
+///
+/// Used for Memory and Disk in the sidebar metrics footer.
+/// Values are expressed in GB (Double) to match `SystemStats` storage;
+/// `formatGB(_:)` trims trailing zeros and appends "GB".
+struct SidebarCapacityMetricView: View {
+
+    /// Short uppercase label, e.g. "MEM" or "DISK".
+    let title: String
+    /// Used capacity in gigabytes.
+    let used: Double
+    /// Total capacity in gigabytes.
+    let total: Double
+
+    /// Available capacity derived from inputs; clamped to zero.
+    private var available: Double { max(total - used, 0) }
+
+    /// Formats a GB value compactly with one decimal place.
+    private func formatGB(_ gb: Double) -> String {
+        String(format: "%.1f GB", gb)
+    }
+
+    /// Full textual description for VoiceOver.
+    private var accessibilityText: String {
+        "\(title), \(formatGB(used)) used, \(formatGB(total)) capacity, \(formatGB(available)) available"
+    }
+
+    /// The compact two-line capacity row.
+    var body: some View {
+        VStack(alignment: .leading, spacing: 2) {
+            HStack {
+                Text(title)
+                    .font(.caption2.weight(.semibold))
+                Spacer()
+                Text("\(formatGB(used)) / \(formatGB(total))")
+                    .font(.caption2.monospacedDigit())
+                    .lineLimit(1)
+            }
+            Text("\(formatGB(available)) available")
+                .font(.caption2)
+                .foregroundStyle(.secondary)
+                .lineLimit(1)
+        }
+        .accessibilityElement(children: .ignore)
+        .accessibilityLabel(accessibilityText)
+    }
+}
--- a/Sources/RunBot/migration/AppShell/Metrics/SidebarMetricsView.swift
+++ b/Sources/RunBot/migration/AppShell/Metrics/SidebarMetricsView.swift
@@ -1,0 +1,64 @@
+// SidebarMetricsView.swift
+// RunBot
+
+import RunBotCore
+import SwiftUI
+
+// MARK: - SidebarMetricsView
+
+/// Pinned sidebar footer showing live CPU, GPU, memory, and disk metrics.
+///
+/// Owns a single `SystemStatsViewModel` whose lifecycle is tied to this view:
+/// sampling starts on `.onAppear` and stops on `.onDisappear`. The sidebar
+/// stays mounted while detail destinations change, so navigation never
+/// restarts the sampler.
+///
+/// ## Layout
+/// A `thinMaterial` rounded-rectangle groups the four rows inside a compact
+/// footer that stays pinned below the navigation `List` in `AppSidebarView`.
+/// The footer height is bounded so it cannot consume the entire sidebar when
+/// the window is short.
+///
+/// ## Reuse
+/// Reuses `SystemStatsViewModel`, `SystemStats`, and `SparklineView`.
+/// Does not embed the full-page `SystemStatsView`.
+@MainActor
+struct SidebarMetricsView: View {
+
+    /// View-local sampler. Constructed once; never written to `MigrationAppDependencies`.
+    @State private var viewModel = SystemStatsViewModel()
+
+    /// The four metric rows grouped in a compact material surface.
+    var body: some View {
+        VStack(spacing: 8) {
+            SidebarUsageMetricView(
+                title: "CPU",
+                value: viewModel.stats.cpuPct,
+                history: viewModel.cpuHistory.values
+            )
+            SidebarUsageMetricView(
+                title: "GPU",
+                value: viewModel.stats.gpuPct,
+                history: viewModel.gpuHistory.values
+            )
+            SidebarCapacityMetricView(
+                title: "MEM",
+                used: viewModel.stats.memUsedGB,
+                total: viewModel.stats.memTotalGB
+            )
+            SidebarCapacityMetricView(
+                title: "DISK",
+                used: viewModel.stats.diskUsedGB,
+                total: viewModel.stats.diskTotalGB
+            )
+        }
+        .padding(10)
+        .background(.thinMaterial, in: RoundedRectangle(cornerRadius: 10))
+        .onAppear {
+            viewModel.start()
+        }
+        .onDisappear {
+            viewModel.stop()
+        }
+    }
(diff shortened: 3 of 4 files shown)


eonist:
### Review of PR #2822: `feat: add sidebar system metrics`

This PR completes Step 8 of the migration plan (#2793) by introducing a pinned live system-metrics footer (`CPU`, `GPU`, `MEM`, `DISK`) at the bottom of `AppSidebarView`.

***

## Architectural & Correctness Analysis

### 1. Sampling Lifecycle and State Isolation
- **Correctness**: `SidebarMetricsView` owns its `@State private var viewModel = SystemStatsViewModel()` and binds sampling to `.onAppear { viewModel.start() }` and `.onDisappear { viewModel.stop() }`. 
- Because `AppSidebarView` remains mounted in the `NavigationSplitView` sidebar column while `selection` changes, detail-view navigation will not trigger teardown or re-instantiation of the sampler.
- Using `@MainActor` on `SidebarMetricsView` matches `SystemStatsViewModel`’s isolation and prevents actor-hopping concurrency warnings under Swift concurrency checks.

### 2. Telemetry and Sparkline Rendering
- **GPU Nil Handling**: `SidebarUsageMetricView` checks `guard let value else { return "—" }` and conditionally hides the `SparklineView` when `value == nil` or `history.isEmpty`, preventing misleading `0%` renders when GPU telemetry is absent on certain hardware configurations.
- **Layout Priorities & Monospacing**: Fixed frame widths (`30` pt leading slot for titles, `42` pt trailing slot for percentages) paired with `.layoutPriority(1)` on the value and `.monospacedDigit()` ensure stable metric alignment and prevent layout jitter when numeric widths change.

### 3. Capacity & Formatting Derivations
- In `SidebarCapacityMetricView`, `available` is computed as `max(total - used, 0)` with `String(format: "%.1f GB", ...)`. This safely matches `SystemStats` stored in gigabytes without negative underflows.

### 4. Accessibility
- All custom metric rows set `.accessibilityElement(children: .ignore)` and synthesize descriptive `.accessibilityLabel(...)` strings covering label, state/percentage, and capacity details for VoiceOver.

***

## Findings Summary

| Severity | Item | Details |
| :--- | :--- | :--- |
| 🔵 Info | Clean lifecycle scoping | Sampler start/stop correctly tied to sidebar appearance; no leaks or unnecessary polling observed. |
| 🔵 Info | Safe nil fallback | GPU unavailable telemetry cleanly handled without fabricating zero values or failing sparkline rendering. |
| 🔵 Info | Monospaced alignment | Numeric labels use monospaced digits with appropriate layout priorities to prevent frame jitter during live updates. |

eonist:
## Review of PR #2822 — sidebar system metrics (step 8)

No blockers. This is the first step in the stack that doesn't duplicate an existing implementation or stub out an action — it genuinely reuses `SystemStatsViewModel`, `SystemStats`, and `SparklineView`, and the nil-GPU handling is correct end to end ([PR #2822](https://github.com/runbot-hq/run-bot/pull/2822)). What's left is a set of doc claims the code doesn't back, one acceptance criterion that isn't met, and a sampler-duplication concern that mirrors #2820.

Verified correct, since it's the part most likely to be wrong: `SystemStatsViewModel.sample()` appends to the GPU ring buffer only under `if let gpu { newGPU.append(gpu) }`, so `gpuHistory.values` never contains fabricated zeros, and `SidebarUsageMetricView` independently omits the sparkline when `value == nil`. The em-dash path is real, not cosmetic. `cpuPct` is non-optional and promotes cleanly into the `Double?` parameter.

## 🟡 Three doc comments assert behavior the code doesn't implement

`SidebarMetricsView` claims a height bound that doesn't exist anywhere in the file:

> The footer height is bounded so it cannot consume the entire sidebar when the window is short.

There's no `.frame(maxHeight:)`, no `.layoutPriority`, nothing. The footer gets its intrinsic height (~130 pt) and the `List` absorbs the remainder — which happens to be fine at the shell's 480 pt minimum height, so the *behavior* is acceptable. The comment is just describing a mechanism that isn't there, and the next person to shrink the window will go looking for it.

`SidebarCapacityMetricView` has the same problem at a smaller scale:

> `formatGB(_:)` trims trailing zeros and appends "GB".

```swift
private func formatGB(_ gb: Double) -> String {
    String(format: "%.1f GB", gb)
}
```

It never trims — `8.0 GB` stays `8.0 GB` — and it appends `" GB"` with a leading space. Either implement the trim or describe the fixed one-decimal format.

Third, `SidebarUsageMetricView` explains that "the sparkline yields layout space before the numeric label," which is achieved by `.layoutPriority(1)` on the `Text`. The two `.layoutPriority(0)` calls on the sparkline and the `Spacer` are no-ops — zero is already the default — so they read as load-bearing when they aren't.

## 🟡 "Metrics remain readable with large byte values" isn't met

The capacity row is `Text("\(formatGB(used)) / \(formatGB(total))").font(.caption2.monospacedDigit()).lineLimit(1)` with no `.minimumScaleFactor`. Work the width budget at the sidebar minimum:

- Sidebar column minimum is 180 pt (`navigationSplitViewColumnWidth(min: 180, …)` in `AppShellView`)
- Minus 10 pt outer padding in `AppSidebarView` and 10 pt inner padding in `SidebarMetricsView`, both sides: 140 pt of content width
- Minus the "MEM"/"DISK" title and the `Spacer`, leaving roughly 100–110 pt for the value

A 2 TB disk renders `2048.0 GB / 4096.0 GB` — around 120 pt at caption2 monospaced. With `lineLimit(1)` and no scale factor, SwiftUI truncates with an ellipsis rather than shrinking, so the user sees `2048.0 GB / 40…`. Adding `.minimumScaleFactor(0.8)` to both capacity lines fixes it, or switch to `ByteCountFormatter`-style compaction (`2.0 TB / 4.0 TB`), which also resolves the trailing-zero comment above.

The usage row is fine by the same arithmetic: 30 + 6 + 24 + 6 + 42 = 108 pt of minimums, comfortably inside 140.

## 🟡 A second SystemStatsViewModel sampler

`PanelMainView` already owns one (`@State private var systemStats = SystemStatsViewModel()`), and `SidebarMetricsView` now creates another. The doc says it "reuses `SystemStatsViewModel` as the single live sampler," which is true only inside the new shell — the menu-bar panel survives until step 10, so any build where both run has two independent timers sampling CPU, GPU, memory, and disk at the same cadence, doubling the `host_processor_info` / IOKit / `statfs` work.

This is the same shape as the duplicated `AppUpdater` in #2820: a resource-owning object instantiated in the migration tree while the legacy owner still exists. Today's window-only build is unaffected because `AppDelegate` never runs, so this is a note for the step-10 checklist rather than a merge blocker — but "single live sampler" should say "single sampler within the app shell."

## 🔵 Smaller notes

- The "available" line uses plain `.caption2`, not `.monospacedDigit()`, while the used/total line above it does. The acceptance criterion says "percentage and capacity values use monospaced digits" — this one is a capacity value, and it'll jitter as the number changes.
- Font sizes are inconsistent between row types: the usage percentage is `.caption`, the capacity values are `.caption2`. Probably deliberate (the percentage is the focal number), but the two row types sit adjacent in the same card.
- `available` is derived as `max(total - used, 0)`. For memory that's correct; for APFS volumes, real available space excludes snapshots and includes purgeable, so `total - used` will disagree with Finder. `sampleDisk()` only returns used/total, so the data isn't there — either rename the line to something derived-sounding, or source real free space from `statfs` in the sampler later.
- The clamp `min(max(value, 0), 100)` appears three times in `SidebarUsageMetricView` (formatted value, accessibility text, sparkline argument). One `private var clampedValue: Double?` removes the chance of them drifting apart.
- `@MainActor` on the `SidebarMetricsView` struct is redundant — SwiftUI `View` bodies are already main-actor isolated, and the sibling `SystemStatsView` that this file models itself on doesn't annotate.
- VoiceOver reads the accessibility labels as "MEM" and "DISK" (spelled out), where the legacy `SystemStatsView` used "Memory Used". Since these labels are already custom strings, "Memory" and "Disk" cost nothing.

## Severity summary

|---|---|---|
| 🟡 | Doc claims a footer height bound, zero-trimming in `formatGB`, and load-bearing `layoutPriority(0)` — none exist in the code | `SidebarMetricsView.swift`, `SidebarCapacityMetricView.swift`, `SidebarUsageMetricView.swift` |
| 🟡 | Large byte values truncate at the 180 pt sidebar minimum; no `minimumScaleFactor` — fails the "readable with large byte values" criterion | `SidebarCapacityMetricView.swift` |
| 🟡 | Second `SystemStatsViewModel` while `PanelMainView` still owns one; "single live sampler" holds only inside the new shell | `SidebarMetricsView.swift` |
| 🔵 | "Available" line lacks monospaced digits; caption vs caption2 mismatch between row types | `SidebarCapacityMetricView.swift` |
| 🔵 | Disk "available" is derived `total - used`, which won't match APFS free space | `SidebarCapacityMetricView.swift` |
| 🔵 | Clamp logic triplicated; redundant `@MainActor`; VoiceOver reads "MEM"/"DISK" as letters | `SidebarUsageMetricView.swift`, `SidebarMetricsView.swift` |

PR #2882, opened 2026-08-17
fix: use hierarchy instead of multiple columns in workflows (#2880)

eonist:
Implements #2880 per the plan in #2881 — the workflow experience goes from five visible columns to the conventional three-part structure: sidebar, workflow hierarchy, step-log detail.

## Changes

- **`AppShellView`** is now a three-column `NavigationSplitView` (sidebar / content / detail) and owns the shared `MigrationWorkflowSelection` so the hierarchy and step-log columns stay in sync.
- **`AppContentView`** (new) routes the content column: workflow hierarchy for Workflows; the existing self-contained feature roots for Local runners, Scopes, and Settings.
- **`AppDetailView`** is now the detail router: derives selected job/step from the live runner snapshot and hosts `MigrationStepLogView`; neutral placeholder for the other sections.
- **`MigrationWorkflowHierarchyView`** (new) — workflow → job → step hierarchy list:
  - filled status dots (`MigrationStatusIndicator`), no donut, no Liquid Glass
  - arrow connector lines via new `MigrationTreeConnector` (simplified from menu-bar `TreeLineLeader`)
  - jobs/steps group gets a subtle fill + stroke so it reads distinct from workflow rows
  - expansion is purely user-driven (`expandedWorkflows: Set<String>`, `expandedJobs: Set<Int>`) — nothing auto-expands on status change
- **`MigrationWorkflowSelection`** gains path-aware `selectJob(_:inWorkflow:)` and `selectStep(_:ofJob:inWorkflow:)` so a tap in any expanded workflow sets the full workflow → job → step path (rows of non-selected workflows stay visible in a hierarchy, unlike the old columns).
- **`MigrationJobRow`** now shows the cloud/desktop runner-type icon.
- Removed: `MigrationWorkflowView` (four-pane root), `MigrationWorkflowListView`, `MigrationJobListView`, `MigrationStepListView`.
- Builds on the memberwise `WorkflowActionGroup` equality from `19d26c65` so rows update live while stable IDs preserve expansion and selection.

## Interaction rules (from #2881)

1. Workflow click selects it and toggles its jobs (re-click only toggles, keeping an active step log).
2. Job click selects it and toggles its steps; a different job clears the stale step.
3. Step click selects it and updates only the log detail.
4. Snapshots reconcile selection by ID (`selection.reconcile`).

## Verification

- `swiftlint --quiet` — clean
- `swift build` — clean
- `swift test` — 342 tests / 41 suites passed, including new `MigrationWorkflowSelection` path tests

[...]

PR #2886, opened 2026-08-17
feat: add selection column for settings (#2860)

### LINKED ISSUE #2860: Add selection column for settings
problem:

Add selection collumn to prefs

authentication
general
about

proposed solution:

Mapping #2860 onto the current settings destination. Note #2858 currently contains only the step-log identity fix, so this lands as additional commits on that branch.

## One decision first

#2860 lists three rows — authentication, general, about — but `MigrationSettingsView` renders four sections, including `UpdateSettingsSection`. Either add an Updates row (four rows) or fold updates into General. Everything below assumes you pick one; the enum is where it shows up.

## Files

**New — `migration/Features/Settings/MigrationSettingsSection.swift`**

Same shape as `AppSection`: `String`-backed enum, `CaseIterable`, `Identifiable` with `id: Self`, plus `title` and `systemImage`. Cases: `authentication`, `general`, `about`, and `updates` if you keep it separate.

**New — `MigrationSettingsListView.swift`**

`MigrationWorkflowColumn(title: "Settings")` wrapping `List(MigrationSettingsSection.allCases, selection:)` with `Label(section.title, systemImage: section.systemImage)` rows. Mirrors `MigrationScopeListView` and `MigrationRunnerListView`, minus the add button.

**New — `MigrationSettingsDetailView.swift`**

Switches on the selection and returns the existing sections unchanged:

- `.authentication` → `AuthenticationSection(authentication:onSignIn:onSignOut:onToggleEnvironment:)`
- `.general` → `GeneralSettingsSection()`, plus `UpdateSettingsSection` if folded
- `.updates` → `UpdateSettingsSection(settings:runnerState:autoUpdater:)`
- `.about` → `AboutSettingsSection()`

**Modified — `MigrationSettingsView.swift`**

Becomes `HSplitView { list; detail }`, owns `@State private var selectedSection: MigrationSettingsSection? = .authentication`, keeps `@Environment(GitHubAuthentication.self)` and the `dependencies` input, and moves the current `ScrollView`/`VStack` body into the detail view. The `onToggleEnvironment` closure that calls `authentication.setSelectedSource` moves with the authentication case.

## What each pane needs

| Pane | Dependencies |
|---|---|
| Authentication | `authentication` from environment, `dependencies.onSignIn`, `dependencies.onSignOut` |
| General | none |
| Updates | `dependencies.settings`, `dependencies.runnerState`, `dependencies.autoUpdater` |
| About | none |

So the detail view needs `dependencies` and reads `authentication` from the environment exactly as `MigrationSettingsView` does now — no new plumbing through `AppShellView` or `AppDetailView`.

## Commit order

1. Add the enum.
2. Add the list pane.
3. Add the detail router and repoint `MigrationSettingsView` at the two-pane layout.

Each step builds on its own. Selection stays in-memory, consistent with the other destinations.

One thing that becomes visible with this change: `onSignIn` and `onSignOut` are still empty closures, so the Authentication pane will now be the first thing users land on while its primary actions do nothing.

eonist:
Part of #2793
Stacked on #2882

## Summary

Migrates Settings into the app shell’s existing content and detail columns. The content column now presents Authentication, General, Updates, and About choices, while the detail column renders the selected section in a scrollable, readable-width layout.

This PR also completes the windowed app’s authentication wiring, restores notification preferences, and standardizes Settings cards, typography, spacing, and row layouts.

## Shell and navigation

- Add `MigrationSettingsSection`, a `String`-backed, `CaseIterable`, `Identifiable` enum for Authentication, General, Updates, and About.
- Add `MigrationSettingsListView` for the selectable Settings rows in the shell content column.
- Add `MigrationSettingsDetailView` to route the current selection in the shell detail column.
- Keep the selected section in `AppShellView`, defaulting to Authentication.
- Route Settings through the shell’s existing columns instead of introducing a nested `HSplitView`.
- Remove the obsolete `MigrationSettingsView` and dead `settingsDependencies` plumbing from `AppContentView`.
- Constrain the picker column to 220–320 points and detail content to a readable maximum width of 820 points.

## Settings presentation

- Add `MigrationSettingsSectionLayout` for consistent detail titles and spacing.
- Add an adaptive `rbSettingsCardBackground` color token and reusable migration card styling.
- Replace glass and double-card surfaces with opaque, borderless cards using continuous 15-point corners.
- Standardize labels, descriptions, controls, padding, row heights, and subtle internal dividers.
- Stack Environment Token and GitHub OAuth as separate full-width authentication rows.
- Enlarge authentication content and controls so titles remain readable without wrapping.
- Update General, Updates, and About to use the same visual hierarchy and card treatment.

## General settings

- Restore the Notifications picker using `NotificationPreferences.shared`, matching the instance consumed by the poller.
- Keep Launch at login in the same two-row General card.
- Preserve automatic-update, beta-channel, update-state, install, relaunch, and retry behavior in Updates.
- Preserve version, build, and pre-release information in About.

## Authentication

- Retain one app-lifetime `OAuthCredentialController` and use the existing shared `GitHubAuthentication` and `GitHubClient`.
- Reconcile credentials and start OAuth observation during app startup.
- Refresh OAuth state and discover the environment token when Authentication mounts.
- Open the URL returned by `oauthCredentials.makeSignInURL()` for GitHub sign-in.
- Sign out through `oauthCredentials.signOut()`, retaining controller-owned Keychain reconciliation and polling restart behavior.
- Allow Environment to be enabled only when a token is available, while always allowing the active Environment source to be turned off.
- When Environment is turned off, select OAuth if signed in; otherwise select unauthenticated.
- Forward valid `runbot://oauth/...` callbacks from SwiftUI `.onOpenURL` to the exact `github.oauthService` instance used by the credential controller.
- Do not create a second OAuth service, credential controller, or authentication model.

## Commit coverage

- `925d97399` — add `MigrationSettingsSection`.
- `6959a499a` — add `MigrationSettingsListView`.
- `718a6fb03` — add `MigrationSettingsDetailView`.
- `9ffd21f4a` — introduce the initial two-pane Settings layout.
- `2bb7e8dc5` — fix documentation and replace the unavailable workflow-column wrapper.
- `52552b887` — route Settings through the shell columns and remove the nested split view.
- `67ee2d082` — remove `MigrationSettingsView` and dead content-column dependencies.
- `49ae1dab4` — add card surfaces, typography, and layout polish.
- `f61611908` — remove double-card surfaces and compact picker rows.
- `0149a5fc0` — enlarge card content, typography, and authentication sizing.
- `b0240ff09` — stack authentication cards vertically as full-width rows.
- `44f3e0e2e` — restore the notification-mode picker.
- `f8c934b6c` — wire authentication discovery, real OAuth actions, and correct Environment toggling.
- `1e3d06fa2` — forward OAuth callbacks to complete sign-in.

## Verification

- [x] Settings opens with Authentication selected by default.
- [x] Every picker row routes to the corresponding detail section.
- [x] Updates remains reachable as a separate section.
- [x] The shared authentication and notification models are reused.
- [x] OAuth controller retention, startup, sign-in, callback, and sign-out paths are wired.
- [x] `swift build` passes.
- [x] `swift test` passes.
- [x] SwiftLint passes.
- [ ] Complete one manual OAuth sign-in/sign-out cycle in the windowed app.

Closes #2860


eonist:
## Review

### 🔴 Detail content can become inaccessible

`Sources/RunBot/migration/Features/Settings/MigrationSettingsDetailView.swift:42`

The refactor removes the `ScrollView` that previously wrapped all settings sections and renders the selected section directly inside `HSplitView`. `HSplitView` does not provide vertical scrolling, so controls below the available window height become unreachable—particularly in the authentication or update sections and at smaller window sizes.

Wrap the routed detail content in a vertical `ScrollView`, retaining appropriate padding and width constraints from the previous implementation. This is a functional regression from the base branch, not merely a layout preference. The same issue is also identified in the existing [review thread](https://github.com/runbot-hq/run-bot/pull/2886#discussion_r3798521599).

## Severity

| Severity | Count | Finding |
|---|---:|---|
| 🔴 blocker | 1 | Detail pane lost vertical scrolling, potentially making settings inaccessible |
| 🟡 suggestion | 0 | — |
| 🔵 info | 0 | — |

eonist:
## Review scope caveat (read first)

I could not read full file bodies: `get_file_contents` for this private repo returned only blob SHAs, not text, and raw.githubusercontent is not reachable for it. So this review is based on the complete patches of all four changed files (they are 100% new-file content except `MigrationSettingsView.swift`), the intent in [#2860](https://github.com/runbot-hq/run-bot/issues/2860), and the repo tree. Two items below are flagged as "verify locally" because they depend on lines outside the hunks — I am not asserting them as confirmed.

## Correctness

**Detail pane lost its scroll container.** The old body was `ScrollView { VStack(spacing: 24) { … } }`. The new `MigrationSettingsDetailView.body` returns the section view directly, and `HSplitView` panes do not scroll their content. `UpdateSettingsSection` (settings + runner state + auto-updater controls) and `AuthenticationSection` are the tallest sections; in a short window their lower controls become unreachable rather than clipped-but-scrollable. This is the one item I'd hold the merge on — acceptance criterion "`UpdateSettingsSection` is reachable via the Updates row" is only half-satisfied if the row switches but the content can't be reached. Both sourcery-ai and cubic flagged the same thing independently, on `MigrationSettingsDetailView.swift` lines 42–52.

**Layout metrics were deleted, not relocated.** The removed body carried `.padding(24)` and `.frame(minWidth: 520, idealWidth: 680, maxWidth: 760, alignment: .leading)`. Neither the detail view nor the split view re-applies any of it: `MigrationSettingsDetailView` has no padding, and `HSplitView` gives both panes no `minWidth`/`idealWidth`. Result is section content flush against the split divider, and a draggable divider that can collapse the detail pane to near-zero width. The docstring on `MigrationSettingsView` still has a `## Layout` heading, but the readability constraint it used to document is gone.

**`selection ?? .authentication` creates a two-state desync.** `MigrationSettingsListView` binds an optional selection, so deselection is reachable (click below the last row, ⌘-click the selected row). When that happens the list renders nothing selected while the detail pane silently shows Authentication. The default already lives in the parent's `@State … = .authentication`; having a second, invisible default in the router means "no selection" is unrepresentable downstream. Either make `selection` non-optional in the detail view and keep the parent as the single owner of the default, or render an explicit no-selection state.

## Architecture and intent

**The shared column wrapper was dropped.** #2860 specifies `MigrationWorkflowColumn(title: "Settings")` wrapping the `List`, mirroring `MigrationScopeListView`/`MigrationRunnerListView` minus the add button. The PR body repeats that claim ("`MigrationWorkflowColumn(title: "Settings")` wrapping `List(…)`"), and cubic's own AI-agent prompt quotes a `MigrationWorkflowColumn` version of the file — but the merged patch hand-rolls the header instead: `VStack(spacing: 0)` + `HStack { Text("Settings").font(.headline); Spacer() }` + `.padding(.horizontal, 12)` + `.frame(height: 44)` + `Divider()`. The view's own doc comment then points at a third component ("Mirrors the header style of `MigrationManagementColumn`"). So the description, the doc comment, and the code disagree, and 12/44 are now magic numbers that will drift from the other three list panes the first time the shared header changes. Note: I found no `MigrationWorkflowColumn.swift` under `Features/Workflows/Columns` (only `MigrationColumnPlaceholder.swift` and `MigrationStepLogView.swift`), so confirm locally whether the shared wrapper exists under a different name before deciding between "reuse it" and "fix the PR body".

**Possibly-dead `@Environment` in `MigrationSettingsView`** (verify locally). The `authentication` environment property sits above the diff hunk and was not removed, but the new body only contains `HSplitView` + `.navigationTitle`. If it is still declared there, it's now unused: it costs an extra observation edge (every `GitHubAuthentication` mutation invalidates the split-view body for nothing) and the retained "`setSelectedSource` is called directly on the environment `authentication` object" reasoning has moved to the detail view. Delete it if the compiler agrees it's unused.

## Tests and docs

No test file accompanies 190 added lines. The cheapest high-value coverage is pure and needs no host app: `MigrationSettingsSection.allCases` count and order (the "four rows, updates stays separate" decision from the PR is currently only recorded in prose), plus non-empty `title`/`systemImage` for every case so a future case addition can't ship with a blank row. Worth adding given the PR's own "`swift test` passes" criterion is otherwise a no-op for this change.

Doc gap: the "Known caveat" about `onSignIn`/`onSignOut` being empty closures is now materially worse — Authentication is the default landing pane, so the first screen users see has two dead primary buttons. It's correctly scoped out of this PR, but it should be an open issue with a number, not a bullet in a PR body that disappears on merge.

## Severity table

|---|---|---|
| 🔴 | Detail pane has no `ScrollView`; tall sections (Updates, Authentication) become unreachable, undermining the "Updates reachable" acceptance criterion | `MigrationSettingsDetailView.swift` |
| 🟡 | `.padding(24)` and pane width constraints removed and not re-applied; divider can collapse the detail pane | `MigrationSettingsView.swift`, `MigrationSettingsDetailView.swift` |
| 🟡 | `selection ?? .authentication` desyncs list (nothing selected) from detail (Authentication shown); duplicate default | `MigrationSettingsDetailView.swift` |
| 🟡 | Shared column wrapper specified in #2860 and claimed in the PR body is not used; hand-rolled header with magic 12/44; doc comment names a third component | `MigrationSettingsListView.swift` |
| 🟡 | No tests for `MigrationSettingsSection` case set, order, or title/symbol mapping | new test file needed |
| 🔵 | `@Environment(GitHubAuthentication.self)` likely now unused in `MigrationSettingsView` — verify and remove | `MigrationSettingsView.swift` |
| 🔵 | `.tag(section)` is redundant with `List(_:selection:)` over an `Identifiable` collection — harmless, drop if you prefer | `MigrationSettingsListView.swift` |

Enum naming, case set (keeping `updates` separate), the router-only design of the detail view, and the "no new plumbing through `AppShellView`/`AppDetailView`" constraint all match the issue's intent — no objections there.

